Issue #369, opened 2019-01-28
Remove `once` calls in test suite


adinapoli-mndc:
As an interim measure for https://github.com/oscoin/oscoin/pull/366 , we placed a number of calls to `once` in the test suite which is limiting the number of QC runs to only 1. This is detrimental for the kind of property-based testing QC shines on. The reason we did that was to overcome the fact our test suite is growing slower due to the use of a lot of cryptographic primitives (hashing primarily).

In order to fix this ticket, we should first remove the roadblock preventing us from actually speeding up the tests, which is logged under https://github.com/oscoin/oscoin/issues/368

adinapoli-mndc:
I have started investigating this as a natural successor of #368. In particular, once I re-enabled the full 100+ QC runs, the running times (with `MockCrypto` !) inflated to like 300+ seconds o.O. I did hit the code with profiling and two things stood out:

1. From my first profiling, a lot of time was spent calling [thisPos](https://github.com/oscoin/radicle/blob/ce65e49ae3023ae0caefaa83745ff02894a50458/src/Radicle/Internal/Annotation.hs#L58) from this [Arbitrary instance](https://github.com/oscoin/oscoin/blob/master/test/Oscoin/Test/Data/Rad/Arbitrary.hs#L37) and is not totally clear to me why we are doing all of that (mostly because I am a Radicle n00b):

<img width="1906" alt="screenshot 2019-03-05 at 12 03 52" src="https://user-images.githubusercontent.com/45846748/53816601-17cb7500-3f64-11e9-9ad4-3c1f61bd9959.png">

@cloudhead , do you have any insights on the above?

2. After commenting out that tagging/untagging shenanigans (leaving only `sized go` in place) lots of tests would fail, but the next profiling would reveal that a significant chunk of time was spent in `serialise` when calling `hashTxs`.

<img width="1906" alt="screenshot 2019-03-05 at 12 11 02" src="https://user-images.githubusercontent.com/45846748/53816735-624cf180-3f64-11e9-8baf-554ad36f279b.png">

We were chatting about 2. with @kim and we were thinking that perhaps (actually, quite likely) what is biting us is the generic deriving for those `Serialise` instances. Handy for sure, but it's pretty common that these automatically-derived instances comes with a performance cost.

I would propose the tentative plan of action for this ticket:

- [ ] Investigate how we can speed up the generation of `Arbitrary` Rad's `Value`s;
[...]

Issue #368, opened 2019-01-28
Abstract over the notion of hashing


adinapoli-mndc:
Every crypto-intensive application will pay a performance tax by the use of cryptographic primitives, which are not a free lunch. This isn't a problem per se in the amortise sense, but the slowdown builds up quite rapidly when lots of these primitives are called. The primary example are tests -- we [observed](https://github.com/oscoin/oscoin/pull/366) a considerate slowdown in their execution due to the intensive use of (production-quality) hashing. 

However, for the sake of testing, we don't need anything that complicated. Provided we abide to the contract of a good hashing function, we can parameterise our `Blockchain` by an abstract concept of hashing and use a mock hash (essentially a sequential counter or similar) during tests, and switch back to a proper implementation during production.

adinapoli-mndc:
Fixed by https://github.com/oscoin/oscoin/pull/380

PR #380, opened 2019-02-14
Abstract over the cryptography in use

adinapoli-mndc:
Hey lads,

I have spent today taking a look at #368, as I think it's a good idea to abstract over the concrete cryptography in use, as "real world crypto" has a computation cost we don't want to pay during tests (see also #369 , which is its direct consequence). Not only that, but when we will start talking about addresses & stuff it might be very handy to have some mock crypto in our tests, as staring at simple integers on a test failure (or whichever mock representation we pick) is way easier than to do so with the "real deal".

As my day approaches its conclusion I wanted to take stock and discuss with you some design choices I'm not fully sure about. First of all, let me introduce the main idea:

My idea was to introduce (yet another!) type parameter (where needed) called `crypto` which parameterise some of our data structures by the cryptography in use. Then, we would have 2 concrete types, defined in `Oscoin.Crypto`:

```haskell
data RealWorldCrypto -- This is empty
data MockCrypto = MockCrypto
```

Then, at the boundary of our application (i.e. the executable and the main test runner) we would call something like `withCrypto @SomeProxy blah` to bring in scope the correct instance and everything "should just work(TM)".

The first elephant we need to move out of the room is the concept of hashing (which started it all). After some attempts, I have settled on this type class:

```haskell
class HasHashing crypto where
    type family HashAlgorithm crypto = algo | algo -> crypto

    -- | A hash using the default hash algorithm. Used instead of 'Hashed' when
    -- the hash pre-image is not known or cannot be typed.
    data family Hash crypto :: *

    fromByteArray :: forall ba. ByteArray.ByteArray ba => ba -> Hash crypto
    toByteArray   :: forall ba. ByteArray.ByteArray ba => Hash crypto -> ba

    hashAlgorithm :: HashAlgorithm crypto

    -- | The zero hash. Also the minimum hash value.
    zeroHash      :: Hash crypto

    -- | The maximum hash value.
    maxHash       :: Hash crypto
```

Where you have a real world implementation:

```haskell
instance HasHashing RealWorldCrypto where
    type HashAlgorithm RealWorldCrypto = Blake2b_256

    -- | A 'Digest' obtained by some hash algorithm @(HashAlgorithm RealWorldCrypto)@.
    --
    -- Textual representations use base-58 encoding. Serialisation and
    -- deserialisation via "Crypto.Hash.Multi".
    newtype Hash RealWorldCrypto =
        Hash { fromHash :: Digest (HashAlgorithm RealWorldCrypto) }
        deriving (Eq, Ord, ByteArrayAccess)

    fromByteArray = Hash . Crypto.hash
    toByteArray = convert . fromHash

    hashAlgorithm = Blake2b_256

    zeroHash = Hash . fromJust $
        Crypto.digestFromByteString @(HashAlgorithm RealWorldCrypto) @ByteString
            (ByteArray.zero (hashDigestSize (Proxy @(HashAlgorithm RealWorldCrypto))))

    maxHash = Hash . fromJust $
        Crypto.digestFromByteString @(HashAlgorithm RealWorldCrypto) @ByteString
            (ByteArray.replicate (hashDigestSize (Proxy @(HashAlgorithm RealWorldCrypto))) maxBound)
```

As well as a dummy one:

```haskell
instance HasHashing MockCrypto where
    type HashAlgorithm MockCrypto = MockHash

    newtype Hash MockCrypto = IntMockHash { fromMockHash :: Int } deriving (Eq, Ord)

    fromByteArray = IntMockHash . sum . map fromEnum . BS.unpack . convert
    -- FIXME(adn) Crappy hack. Benchmark and see if we can simply leverage
    -- 'Hashable ByteString' for both 'fromByteArray' and 'toByteArray'.
    toByteArray   =
        let go x | x <= 255  = [toEnum x]
                 | otherwise = toEnum 255 : go (x - 255)
        in convert . BS.pack . go . fromMockHash

    hashAlgorithm = MockHash
    zeroHash = IntMockHash 0
    maxHash  = IntMockHash maxBound
```

I kinda like it, because the concrete implementation is mentioned exactly in one place (i.e. `Blake2b_256`) but at the same time the use of a `data family` there allows to "override" the particular `Hash` type constructor with something very very simple in the mock case. Furthermore, the first type family is injective so that picking a particular crypto fully specify the algorithm in use. (that's useful for example in the `HasDigitalSignature` typeclass).

I also feel is quite natural; this typeclass is a witness that our cryptography has the concept of hashing implemented. You can do the same thing (I have done it already to an extent) in the `PubKey` module, where we introduce the concept of `HasDigitalSignature`. Note how in that module we have to also embellish both the `PublicKey` and the `PrivateKey` with the `crypto` type variable as we need to pick a particular implementation to hash stuff.

The `fromByteArray` and `toByteArray` methods are _slightly_ artificial but I have added those to be able to write some instances in a generic way (for example, the `Semigroup` one). The implementation for the `MockHash` case is a sloppy hack I will need to fix.

I have then started to amend the `Block` module as - as you would imagine - also the `Block` and the `BlockHeader` needs to be extended with the `crypto` type variable (darn, that's gonna be a painful refactoring 😅 ) but after a couple of dubious choices on my part, I think that's time to discuss. In particular:

1. We have quite a few external libraries (`multihash`, which I'm gonna touch base on later) and the `avl-auth` which works against a concrete crypto (i.e., cryptonite) and manipulate `Digest`s. In particular, I had this function in `Block.hs`:

```haskell
hashTxs :: (Foldable t, Serialise tx) => t tx -> Crypto.Hash crypto
hashTxs (toList -> txs)
    | null txs = Crypto.zeroHash
    | otherwise =
          Crypto.fromHash -- This was a function from Digest -> Hash
        . AuthTree.merkleHash
        . AuthTree.fromList
        $ [(tx, mempty :: ByteString) | tx <-
            map (LBS.toStrict . Serialise.serialise) txs]
        -- Nb. Since our Merkle tree works with key-value pairs, but we're only
        -- really interested in the keys being present or absent for this use-case,
        -- we use the empty byte string as the value component.
```

that I have turned into:

```haskell
hashTxs :: (Foldable t, Serialise tx) => t tx -> Crypto.Hash crypto
hashTxs (toList -> txs)
    | null txs = Crypto.zeroHash
    | otherwise =
          Crypto.fromByteArray
        . ByteArray.convert
        . AuthTree.merkleHash
        . AuthTree.fromList
        $ [(tx, mempty :: ByteString) | tx <-
            map (LBS.toStrict . Serialise.serialise) txs]
        -- Nb. Since our Merkle tree works with key-value pairs, but we're only
        -- really interested in the keys being present or absent for this use-case,
        -- we use the empty byte string as the value component.
```

Now, this is a bit smelly as it's piggybacking on the fact `Digest` is an instance of `ByteArrayAccess` and thus I can convert it into a `ByteString` and then call `fromByteArray`, but I am not sure this is gonna work (at runtime, I mean).

2. Multihash -- at the moment our `Multi.hs` module is (quite rightly) very concrete and very much geared towards Cryptonite. To complicate things, as @kim pointed out, we should be moving away from this module in favour of the external one living in the `ipfs` repo, so I have to treat this as a blackbox I cannot modify. This, per se, is not a problem, if not for things like the following (in `Hash.hs`):

```haskell
multiB58 :: Multihashable (HashAlgorithm crypto) => Hash crypto -> BaseN.Base58
multiB58 = Multihash.encodeAtBase BaseN.Base58 . Multihash.fromDigest . fromHash
```
This function cannot be written generically even if I do the `convert . fromByteArray` hack (which would already be smelly). This is because I need to prove to the compiler that there is a 1:1 correspondence between my `HashAlgorithm` and the `MuliHash.HashAlgorithm` which is yet another 1:1 correspondence with Cryptonite's `HashAlgorithm` (oh my, name overload 😅 ). Note that due to the latter, even implementing an `instance Cryptonite.HashAlgorithm MockHash` is not gonna work, as there is no direct correspondence of my faux hash algorithm in the multihash world.

I am starting to think the way to go might be to pull out the `multihash` external dependency and in the `MultiHash` module (which would be replaced almost entirely) create another `HasMultiHash` typeclass which would pick the concrete multihash for `RealWorldCrypto` but something simple for `MockCrypto`. Do you think that's a sensible approach?

@kim , @cloudhead what do you think of the above, also by looking at the code I have written so far? Do you think this has the potential to fly?


adinapoli-mndc:
@cloudhead @kim OK, daily update: I have made progress in my humongous refactoring and I think I am starting to see the light at the end of the tunnel. There are quite a bit of `undefined` in the codebase and I am "just" at 56/88 modules (yeah, it will take a while 😅 ) but today and Friday I settled on a particular design for `Block.hs` I would like to share with you, plus bounce off some ideas and impressions on the P2P module and how this crypto stuff affects it.

### Block.hs

As we discussed last week, I have implemented (a variant of) the design of having a `Block` being sealed or unsealed directly in the types. The first attempt was to use a single type, lift the constructors at the type level and have something like:

```haskell

data Seal c s =
    Unsealed
  | SealedWith (Crypto.Hash c) s

data BlockHeader c (seal :: Seal c s) = BlockHeader {
   blockSeal :: s
   }
```

But this ended up being a fiasco for a couple of reasons: first, now the compiler would forbid the creation of `BlockHeader c 'Unsealed` because what would be a suitable value to inhabit `blockSeal`? In case of unsealed blocks, we wouldn't have any (that's why the previous design used `()`). Secondly, now lifting `SealedWith` at the type level would make the compiler explode as you would have to write something like:

``` haskell
sealBlock :: s -> Block c 'Unsealed  -> Block c '( SealedWith (Crypto.Hash c) s) 
```

Which simply didn't compile. Eventually, I settled on splitting those type constructors into two types, so now we have:

```haskell

data Unsealed = Unsealed deriving (Generic, Show)

newtype Sealed c s = SealedWith s
    deriving (Generic, Show, Serialise)

-- | Block header.
data BlockHeader crypto s = BlockHeader
    { blockPrevHash         :: (Crypto.Hash crypto)
    , blockDataHash         :: (Crypto.Hash crypto)
    , blockStateHash        :: (Crypto.Hash crypto)
    , blockTimestamp        :: Timestamp
    , blockTargetDifficulty :: Difficulty
    , blockSeal             :: s
    } deriving (Show, Generic)
```

Which is flexible enough to allow us writing things like:

```haskell
sealBlock
    :: ( Crypto.Hashable c (BlockHeader c (Sealed c s)) )
    => s
    -> Block c tx Unsealed
    -> Block c tx (Sealed c s)
```

it also means that `Block` is now opaque (we don't export its type constructor) and that `Unsealed` is not exported (the type constructor, I mean). Now our `Blockchain` is also parametric over a crypto `c` and accepts only `Sealed` blocks, which is kinda nice.

This has also ripple effects on the `Miner` function, which is now defined as:

```haskell
type Miner c s m = forall tx.
    (Depth -> m [Block c tx s])
    -> BlockHeader c Unsealed
    -> m (Maybe (BlockHeader c (Sealed c s)))
```

Where previously we had:

```haskell
type Miner s m = forall tx unsealed.
    (Depth -> m [Block tx s])
    -> BlockHeader unsealed
    -> m (Maybe (BlockHeader s))
```

It's a subtle difference but it now means that our `BlockHeader` cannot simply carry _anything_ (i.e. the forall unsealed) but a very concrete seal. This has some consequences in the `mineNakamoto` function I am trying to untangle, but I won't share these concerns now to not inflate this big-already report. OK, next..

### P2P & co

Something I did find a bit peculiar (not necessarily wrong but worth pointing out) is that now due to the fact that also our notion of `PublicKey` is parametric over `c` (take a look at `PubKey.hs`) it means that our `NodeId` also is, and this bubbles up till the `Connection.hs` module, where we have:

```haskell
data Connection c wire = Connection
    { connNodeId   :: NodeId c
    , connAddr     :: SockAddr
    , connSendWire :: wire -> IO ()
    , connRecvWire :: ConduitT () wire IO ()
    , connClose    :: IO ()
    }
```

For some reason I find slightly awkward to talk about crypto stuff here as it almost feels out of place, but after further thought I think it might actually make perfect sense: after all, in our gossip protocol we do sign/verify messages & we hash things, so it's kinda OK to have `c` explicitly mentioned here. We also gain the added bonus that now we can mock the crypto for our gossip/network tests.

@cloudhead , @kim thoughts? Do you think this is heading into the right direction?

Ps. note to myself: I think that for the future, in case of such big PRs, would be wise for me to simply split my refactoring into multiple "WIP" comments so you guys can review only the delta and not stare at a single gargantuan commit. 🤦‍♂️

kim:
> our notion of `PublicKey` is parametric over `c` [...] it means that our `NodeId` also is

This is indeed a bit odd, as - strictly speaking - it is an implementation detail of the signature algorithm whether or not a hashing function is needed to sign / verify a message. Will need to look at it again tomorrow when fresher.

adinapoli-mndc:
> This is indeed a bit odd, as - strictly speaking - it is an implementation detail of the signature algorithm whether or not a hashing function is needed to sign / verify a message.

That's a valid point, and there are some considerations to be made here. Let me use this medium as a place for sharing my thoughts out loud. At the moment yes, we _could_ consider the hashing an implementation detail and (maybe?) make the `NodeId` parametric over the `PublicKey` with something like:

```
newtype NodeId pk = NodeId pk
```

Apart the questionable abstraction here (as the `NodeId` is not opaque anymore) we would simply move the choice to pick a concrete `pk` elsewhere in the program. I am not too convinced we can make `NodeId` completely opaque as surely for the real world crypto we do want to use hashing and I don't see a way to "Bridge the gap" at the runtime level between a `NodeId` and the crypto in use without the use of some type parameter.

But more importantly, at the moment our `PubKey` implementation is using elliptic curve cryptography for both the mock & real world crypto. But this is not OK for tests, as ECSDA will be very slow and in tests where we generate big networks is going to eat most of the time. To put it differently, I think this concern calls for an explicit `c` algorithm in the `PublicKey`, because it will allow us to swap ECSDA in favour of a fast algorithm during tests.

What do you think?

cloudhead:
Nice progress. The Kind-level seal did look quite messy and I much prefer the solution you settled on.

> it also means that `Block` is now opaque (we don't export its type constructor)

This I don't quite understand - why does it have to be opaque? Surely you want to access the transactions and header?

adinapoli-mndc:
> This I don't quite understand - why does it have to be opaque? Surely you want to access the transactions and header?

Sorry, bad wording on my part: we _do_ export the getters (so `blockHeader` and `blockData`) but not the type constructor itself. The idea was to discourage people from manually constructing a `Block` and putting whichever seal they want in.

I appreciate this is just a "half" solution as you could hijack the system and use the `blockHeader` as a setter, but the `Unsealed` type constructor is not exported so the most damage you can do is to swap the seal (which is not OK either I guess, better to go via a specialised function).

An obvious improvement over that would be to turn `blockHeader` & co into `Getter'`, so that we could only _get_ stuff, never put it back. But that can be done after this big refactoring, methinks.

Am I making sense or am I still half asleep?

kim:
@adinapoli-mndc It is completely fine that `NodeId` is not opaque -- we may even replace it by a simple type synonym at some point. I'm rather questioning what we're abstracting _over_ for the key pair. For blocks, it pans out quite nicely, because what we're dealing with is indeed hashes. A key pair could use an algorithm which does not even require a hashing function.

Thus, as you note, the "right" abstraction would be to parametrise over [public|private] key wherever needed.

Unrelatedly, I notice that you baked in `Serialise` into the `HasDigitalSignature` class, where imho it should be concerned with only `ByteString` (or `ByteArrayAccess`, as that's what we have no established as the least common denominator). Signing / verifying `Serialise`able values should be top-level convenience functions.

Here's an attempt:

```haskell
{-# LANGUAGE ScopedTypeVariables    #-}
{-# LANGUAGE TypeFamilies           #-}
{-# LANGUAGE UndecidableInstances   #-}

import qualified Crypto.Hash as Cryptonite
import qualified Crypto.PubKey.ECC.ECDSA as ECDSA
import           Crypto.Random.Types (MonadRandom)
import           Data.ByteArray
import           Oscoin.Crypto (MockCrypto, RealWorldCrypto)
import qualified Oscoin.Crypto.Hash as Crypto


data PublicKey k where
    PublicKey :: Crypto.HasHashing c => k -> Crypto.Hashed c k -> PublicKey k

newtype PrivateKey k = PrivateKey k

data Signed msg = Signed
    { sigMessage   :: msg
    , sigSignature :: Signature
    }

newtype Signature = Signature ECDSA.Signature

class HasDigitalSignature k where
    data PK k :: *
    data SK k :: *

    sign   :: (ByteArrayAccess ba, MonadRandom m) => SK k -> ba -> m (Signed ba)
    verify :: ByteArrayAccess ba => PK k -> Signed ba -> Bool

data ECDSA hashing

instance ( Crypto.HasHashing h
         , Cryptonite.HashAlgorithm (Crypto.HashAlgorithm h)
         ) => HasDigitalSignature (ECDSA h)
  where
    newtype PK (ECDSA h) = PK (PublicKey  ECDSA.PublicKey)
    newtype SK (ECDSA h) = SK (PrivateKey ECDSA.PrivateKey)

    sign (SK (PrivateKey sk)) bytes =
        let (alg :: Crypto.HashAlgorithm h) = Crypto.hashAlgorithm
         in Signed bytes . Signature <$> ECDSA.sign sk alg bytes

    verify (PK (PublicKey pk _)) (Signed bytes (Signature sig)) =
        let (alg :: Crypto.HashAlgorithm h) = Crypto.hashAlgorithm
         in ECDSA.verify alg pk sig bytes
```

(nb. the `Signature`'s type would also need to match, and the `publicKeyHash` top-level function would just return a `ByteArray`)

adinapoli-mndc:
Thanks @kim , I like where this is going. Out loud thoughts:

> Unrelatedly, I notice that you baked in `Serialise` into the `HasDigitalSignature` class, where imho it should be concerned with only `ByteString` (or `ByteArrayAccess`, as that's what we have no established as the least common denominator). Signing / verifying `Serialise`able values should be top-level convenience functions.

Yessir. I did whip up that typeclass in a hurry without even thinking about it properly, and now we should be in a better place as I am trying to painstakingly check every occurrence of `Serialise` in the codebase as an innocent use of `hashSerial` tends to be quite infectious and make the constraint pop up everywhere, but that, as you noted, should mostly be an implementation detail of which top-level crypto we decide to run our system with.

Regarding your proposal, I like it a lot. Let me try plugging it into our codebase and see how it pans out 😎 .

cloudhead:
Makes sense!

src/Oscoin/Crypto/Hash/RealWorld.hs
```diff
@@ -0,0 +1,125 @@
...
+    hash = toHashed . fromHashed . hash . LBS.toStrict
+
+instance Hashable RealWorldCrypto Word8 where
+    hash = toHashed . Hash . Crypto.hash . BS.singleton
+
+instance (ByteArrayAccess a) => Hashable RealWorldCrypto (Maybe a) where
+    hash (Just x) = toHashed . Hash $ Crypto.hash x
+    hash Nothing  = toHashed $ zeroHash
+
+instance Multihashable (HashAlgorithm RealWorldCrypto) => FromJSON (Hash RealWorldCrypto) where
```
kim: The `Multihashable` constraints here can be omitted (not smaller than instance head), and `UndecidableInstances` can be removed.

src/Oscoin/Crypto/Hash/RealWorld.hs
```diff
@@ -0,0 +1,125 @@
...
+
+instance Multihashable (HashAlgorithm RealWorldCrypto) => FromHttpApiData (Hash RealWorldCrypto) where
+    parseQueryParam =
+        bimap T.pack Hash . decodeAtBase BaseN.Base58 . encodeUtf8
+
+
+instance Buildable (Hash RealWorldCrypto) where
+    build (Hash digest) =
+        let (bs :: ByteString) = convert digest
+        in BaseN.encodedTextBuilder $ BaseN.encodeAtBase BaseN.Base58 bs
```
kim: I think this now introduced subtle bugs: `parseQueryParam . toQueryParam /= id`, `fromField . toField /= id`, `parseJSON . toJSON /= id`. 

Shall we just define `Buildable (Hash RealWorldCrypto)` to give us a Base58-encoded _multihash_, and define all other stringly instances in terms of `build`?

src/Oscoin/Crypto/Hash/RealWorld.hs
adinapoli-mndc: Something that GHC/stack/ghcid performs (seemingly) arbitrarily is exactly warning about redundant constraints. It seems like that gets triggered only if you modify a module deep down the guts which causes (forces) a recompile. Isn't there a more aggressive compiler flag to do this all the time (unlikely, but still)

src/Oscoin/Crypto/Hash/RealWorld.hs
adinapoli-mndc: I am quite toasted after a day spent in the trenches, but what you propose sounds plausible. What needs to be worked out is whether or not the `Multihash` constraint ends up being "infectious" and starts to bubble up in functions which now requires _only_ a `Buildable (Hash c)` instance. That would be bad because we do not have an `instance Multihash (Hash MockCrypto)` and we cannot even define it (for the reasons we discussed previously).

I need to take a fresh look tomorrow morning.

adinapoli-mndc:
Currently refactoring test module 34/50. Getting there :feelsgood: .

src/Oscoin/Crypto/Hash/RealWorld.hs
kim: I think this doesn't trigger a warning because `UndecidableInstances`

src/Oscoin/Crypto/Hash/RealWorld.hs
kim: The `Mutihashable` is determined by `Hash RealWorldCrypto`, so methinks the `instance Buildable (Hash MockCrypto)` could just do something else, provided the `From*`-type instances are on the concrete types.

adinapoli-mndc:
@cloudhead @kim let's finish the week with some nice results 🎉 

I have finally wired everything. I ended up _not_ rolling up my custom digital signature algorithm but rather using a weaker elliptic curve / Blake2b algorithm for the `HasDigitalSignature MockCrypto` and I am happy to report now the tests passes, both for the mock and the real world crypto. What's nice of this approach is that swapping cryptos is literally two lines of code:

```haskell
type CryptoUnderTest = IsCrypto MockCrypto

main :: IO ()
main = do
    config <- Consensus.getConfig Testing
    defaultMainWithIngredients (map failFast defaultIngredients) $ testGroup "All"
        [ Oscoin.tests (Dict @CryptoUnderTest) config
        , Multihash.tests
        , Concurrent.tests
        , Conduit.tests
        , Integration.tests
        ]
```

(One line to change the type alias, one line to import the correct type 😉 ). We can now think about using singleton types to allow our test suite to be parametric over the crypto (as in passing the crypto as an CLI argument) so that we can run tests with both cryptos and have the real-world ones being run overnight as part of CI.

Now the drum roll 🥁 .. how are running times?

**RealWorldCrypto**

screenshot 2019-03-01 at 16 53 38

**MockCrypto**

screenshot 2019-03-01 at 16 53 23

From my (failing) experiments we can bring this number down to roughly 15 seconds if we roll our own dummy digital signature, but I haven't gone there yet.


Happy Friday!

kim:
hercules

cloudhead:
img

adinapoli-mndc:
Glad to see this is been greeted with enthusiasm 😁 . This is what I am planning to do today and on Tuesday to bring this PR in shape for review/merge:

- [ ] Address conflicts (kinda obvious 😅 )
- [ ] Remove that `fromByteArray` nonsense and replace instances using it with specialised, crypto-concrete instances.
- [ ] Remove dubious instances (i.e. `Semigroup Hash`) in favour of crypto-specialised ones (for example @cloudhead would like to be able to elegantly concatenate merkle hashes together, so we might have to rewrite this to work for `MerkeHash h` hashes)
- [ ] Address hlint & weeder concerns (boy, I think this will take me a while)
- [ ] (Optional, but nice) roll my own super-simple digital scheme without passing via `cryptonite`, which should yield an extra performance boost.

Anything else which I am missing?

exe/node/Main.hs
```diff
@@ -112,7 +113,8 @@ main = do
     keys        <- runReaderT readKeyPair keysPath
     nid         <- pure (mkNodeId $ fst keys)
     mem         <- Mempool.newIO
-    gen         <- Yaml.decodeFileThrow genesis :: IO (Block RadTx Nakamoto.PoW)
+    gen         <-
+        Yaml.decodeFileThrow genesis :: IO (Block RealWorldCrypto (RadTx RealWorldCrypto) (Sealed RealWorldCrypto Nakamoto.PoW))
```
cloudhead: Can we maybe break this line up differently? It's verrry long. Alternatively, maybe type applications would be shorter?

src/Oscoin/CLI/Command.hs
```diff
@@ -14,25 +14,29 @@ import           Oscoin.CLI.Revision
...
+import qualified Oscoin.Crypto.Hash as Crypto
 import qualified Oscoin.Crypto.PubKey as Crypto
 import qualified Oscoin.Data.RadicleTx as Rad
 import           Oscoin.Data.Tx (mkTx)
 import           Oscoin.P2P (NodeAddr(..), mkNodeId)
 import           Oscoin.Time (Timestamp)
 
 
+import           Codec.Serialise (Serialise)
+import qualified Crypto.Data.Auth.Tree.Internal as AuthTree
```
cloudhead: A little strange that we need to import the `Internal` module here.

exe/node/Main.hs
adinapoli-mndc: I am wondering if we could simply type alias this long type (in an ad-hoc way in this module) and use that 🤔

src/Oscoin/CLI/Command.hs
adinapoli-mndc: It was alas an oversight of mine when patching up `avl-tree`, because the `MerkleHash` typeclass indeed lives in the `Internal` module, but it should have probably have lived in something like `.Class`. Happy to fire up a separate PR for that (if you think this can wait up until after merging this guy, but I am happy either way).

exe/node/Main.hs
adinapoli-mndc: 9a927aa looks like `GenesisBlock` is fitting here.

adinapoli-mndc:
@kim @cloudhead ok, loot for today: I have ticked all the boxes of the refactoring/polishing I wanted to do, including coming up with a dead-simple dig-sig algorithm for our mock crypto. I have finally settled on the following:

```haskell
instance HasDigitalSignature MockCrypto where

    newtype PK MockCrypto =
        MockPK (PublicKey MockCrypto ByteString) deriving Eq
    newtype SK MockCrypto = MockSK (PrivateKey (ByteString, [Word8]))

    newtype Signature MockCrypto =
        MockSignature ByteString deriving (Eq, Show)

    sign (MockSK (PrivateKey (pkSkXored, sk))) bytes =
        let sig = BS.pack . BS.zipWith xor pkSkXored . BS.pack $ sk
        in pure $ Signed bytes (MockSignature sig)

    verify (MockPK (PublicKey pk _)) (Signed _ (MockSignature sig)) =
        pk == sig

    generateKeyPair = do
        pk <- getRandomBytes 16
        sk <- getRandomBytes 16
        pure ( MockPK $ PublicKey pk (Crypto.hash pk)
             , MockSK $ PrivateKey (BS.pack $ BS.zipWith xor pk sk, BS.unpack sk)
             )
```

In a nutshell we do a simple `xor` so that eventually the `Signature` includes the public key, and things works as they should.

The little elephant in the room is that I have currently 1 failing test between me and endless glory, in particular, this one:

```haskell
encodeDecodeTx :: IsCrypto c => API.RadTx c -> Property
encodeDecodeTx tx =
    (Aeson.eitherDecode . Aeson.encode) tx === Right tx .&&. verifyTx tx
```

The offending line is `verifyTx`, which is failing. This is strange as:

a) The failure is happening after 4 tests;
b) Other tests relying on signing/verification (including gossip) passes. I would speculate this is due to some particular input picked by the generator. I will investigate in the morning.

Thanks to the mock dig-sig we are down to **16 seconds** for the entire test suite, which is pretty good 😎

adinapoli-mndc:
Uhm, the tests keep failing with:

```
Process exited with code: ExitFailure 1
/workspace/.stack/setup-exe-cache/x86_64-linux/Cabal-simple_mPHDZzAJ_2.2.0.1_ghc-8.4.4 --builddir=.stack-work/dist/x86_64-linux/Cabal-2.2.0.1 build lib:oscoin exe:oscoin exe:oscoin-cli test:documentation test:test --ghc-options " -ddump-hi -ddump-to-file"
-- While building package oscoin-0.1.0.0 using:

`gcc' failed in phase `Linker'. (Exit code: 1)
collect2: error: ld returned 1 exit status
/usr/bin/ld.gold: /workspace/.stack-work/install/x86_64-linux/lts-12.17/8.4.4/lib/x86_64-linux-ghc-8.4.4/xxhash-0.0.2-CMpRdFPrlMNDc66wDcXZqK/libHSxxhash-0.0.2-CMpRdFPrlMNDc66wDcXZqK.a(xxhash.o): previous definition here
/usr/bin/ld.gold: error: /opt/ghc/8.4.4/lib/ghc-8.4.4/rts/libHSrts_thr.a(xxhash.thr_o): multiple definition of 'XXH32_digest'
/usr/bin/ld.gold: /workspace/.stack-work/install/x86_64-linux/lts-12.17/8.4.4/lib/x86_64-linux-ghc-8.4.4/xxhash-0.0.2-CMpRdFPrlMNDc66wDcXZqK/libHSxxhash-0.0.2-CMpRdFPrlMNDc66wDcXZqK.a(xxhash.o): previous definition here
/usr/bin/ld.gold: error: /opt/ghc/8.4.4/lib/ghc-8.4.4/rts/libHSrts_thr.a(xxhash.thr_o): multiple definition of 'XXH32_update'
/usr/bin/ld.gold: /workspace/.stack-work/install/x86_64-linux/lts-12.17/8.4.4/lib/x86_64-linux-ghc-8.4.4/xxhash-0.0.2-CMpRdFPrlMNDc66wDcXZqK/libHSxxhash-0.0.2-CMpRdFPrlMNDc66wDcXZqK.a(xxhash.o): previous definition here
/usr/bin/ld.gold: error: /opt/ghc/8.4.4/lib/ghc-8.4.4/rts/libHSrts_thr.a(xxhash.thr_o): multiple definition of 'XXH32'
Linking .stack-work/dist/x86_64-linux/Cabal-2.2.0.1/build/test/test ...
```

However, I cannot reproduce this linking error on my machine (But I am _not_ using the gold linker as I am on darwin). Sob!

adinapoli-mndc:
Ok, nice, swapping the `xxhash` library in favour of the `xxhash-ffi` library did the trick. Go figure.

adinapoli-mndc:
Yay, we finally have green lights 😻 

We are consistently at 16 seconds for the whole test suite, and I think most of the time is dominated by my tests on fork selection, which is a bit strange. I suspect there is more performance work to be done in that area (i.e. the orphanage for example) but I think we all agree that's for another PR 😬 😬

src/Oscoin/CLI/Command.hs
cloudhead: Merged the other PR.

cloudhead: Happy to wait until this is merged :+1:

cloudhead:
Amazing. Now to go through the code :sweat_smile:

exe/node/Main.hs
```diff
@@ -112,7 +115,8 @@ main = do
     keys        <- runReaderT readKeyPair keysPath
     nid         <- pure (mkNodeId $ fst keys)
     mem         <- Mempool.newIO
-    gen         <- Yaml.decodeFileThrow genesis :: IO (Block RadTx Nakamoto.PoW)
+    gen         <-
+        Yaml.decodeFileThrow genesis :: IO GenesisBlock
```
cloudhead: This can be one line now : )

src/Oscoin/API/HTTP.hs
```diff
@@ -20,23 +23,66 @@ import qualified Network.Wai.Middleware.Static as Wai
...
+       , FromJSON (Crypto.PK c)
+       , FromJSON (Crypto.Signature c)
+       , Ord (Crypto.Hash c)
+       , Buildable (Crypto.Hash c)
+       , ToJSON (Crypto.Hash c)
+       , ToJSON (Crypto.Signature c)
+       )
+    => Int
+    -> Node.Handle c (RadTx c) (Rad.Env c) s i
+    -> IO ()
```
cloudhead: Do we ever need to run this node with non-`RealWorldCrypto`?

src/Oscoin/API/HTTP/Client.hs
```diff
@@ -1,3 +1,4 @@
+{-# LANGUAGE UndecidableInstances #-}
```
cloudhead: Would you say this can be avoided with a bit more work - ie. is this debt, or a necessary evil?

src/Oscoin/API/HTTP/Client.hs
```diff
@@ -53,14 +57,20 @@ newtype HttpClientT m a = HttpClientT (ReaderT (Requester m) m a)
 instance MonadTrans HttpClientT where
     lift ma = HttpClientT $ lift ma
 
-instance (Monad m, MonadIO m) => MonadClient (HttpClientT m) where
+instance ( Buildable (Hash c)
+         , Serialise (BlockHash c)
+         , Serialise (PK c)
+         , Serialise (Signature c)
```
cloudhead: Can we not say in one place: "given a `Serialise c`, `PK c` is also a `Serialise`? And likewise for all the others?

cloudhead: I'm seeing this pattern repeated in a bunch of places is why I ask, and would be nice if we just needed something like `Serialise c`

src/Oscoin/API/Types.hs
```diff
@@ -68,19 +70,23 @@ data TxLookupResponse = TxLookupResponse
...
     -- the 'Block' referenced by 'txBlockHash'.
-    , txPayload       :: RadTx
+    , txPayload       :: RadTx c
     -- ^ The transaction itself.
-    } deriving (Show, Eq, Generic)
+    } deriving (Generic)
 
-instance ToJSON TxLookupResponse where
+deriving instance (HasHashing c, Show (Hash c), Show (RadTx c)) => Show (TxLookupResponse c)
+deriving instance (Eq (Hash c), Eq (RadTx c))    => Eq (TxLookupResponse c)
```
cloudhead:
```suggestion
deriving instance (Eq (Hash c), Eq (RadTx c)) => Eq (TxLookupResponse c)
```

exe/node/Main.hs
adinapoli-mndc: True! 😁

src/Oscoin/Consensus/Mining.hs
```diff
@@ -19,51 +19,52 @@ import qualified Oscoin.Node.Mempool.Class as Mempool
...
         Nothing ->
             pure Nothing
 
 -- | Mine a genesis block with the given 'Miner'.
 mineGenesis
-    :: (Monad m, Serialise s)
-    => Miner s m
-    -> Block tx r
-    -> m (Either Text (Block tx s))
+    :: ( Monad m )
```
cloudhead:
```suggestion
    :: Monad m
```

src/Oscoin/Consensus/Nakamoto.hs
```diff
@@ -81,41 +100,71 @@ validateBlock prefix@(parent:_) blk
...
+
   where
-    go :: BlockHeader PoW -> m (Maybe (BlockHeader PoW))
-    go hdr@BlockHeader { blockSeal = PoW nonce }
-        | hasPoW hdr            = pure $ Just $ hdr $> PoW nonce
-        | nonce < maxBound      = go hdr { blockSeal = PoW (nonce + 1) }
-        | otherwise             = pure Nothing
+    mine :: Difficulty -> PoW -> Maybe PoW
+    mine target (PoW nonce)
+        | blockDifficulty < target = Just (PoW nonce)
```
cloudhead: I think it makes more sense to use `hasPoW` here like before, so we have only one place which verifies the seal.

src/Oscoin/API/HTTP.hs
adinapoli-mndc: Do we (or do we plan to) exercise this part of the codebase via the testsuite? If this is a top-level runner we use in the executable only we can monomorphise this over `RealWorldCrypto`.

src/Oscoin/Crypto.hs
```diff
@@ -0,0 +1,8 @@
+module Oscoin.Crypto
+    ( RealWorldCrypto
+    , MockCrypto(..)
+    ) where
+
+data RealWorldCrypto
```
cloudhead: I think it makes sense to call this just `Crypto`, since "real-world-ness" is implied in the rest of the codebase - but we can do this in a later PR.

cloudhead: A comment explaining what it's for would also be nice.

src/Oscoin/API/HTTP/Client.hs
adinapoli-mndc: This is a necessary evil, because we need to expand the type family `Hash` in the LHS of our constraints. I was for a long time scared by `UndecidableInstances`, but it's actually a benign extension in my experience. Cfr. https://monadfix.io/#2017-12-18-undecidableinstances

src/Oscoin/Crypto.hs
```diff
@@ -0,0 +1,8 @@
...
+
+data MockCrypto = MockCrypto
```
cloudhead: Question: is there a reason this is here, instead of under the `Test.*` hierarchy?

src/Oscoin/Crypto/Blockchain/Block.hs
```diff
@@ -54,25 +68,46 @@ type Depth = Natural
...
 
+data Unsealed = Unsealed deriving (Generic, Show)
+
+-- NOTE(adn) In the midst of the refactoring for oscoin#368, at some point
+-- GHC started requesting me a 'Serialise Unsealed' constraint in various
+-- places. My intuition is that /in theory/ we should never serialise an
+-- 'Unsealed' block, so we should do another sweep through the different
+-- constraints and try to pinpoint why is that. For now, the path of least
+-- resistence is to simply provide such (dubious) instance.
+instance Serialise Unsealed
```
cloudhead: We should maybe turn this into an issue - might be a good idea to implement it with `panic` as the implementation.

src/Oscoin/Crypto/Blockchain/Block.hs
```diff
@@ -54,25 +68,46 @@ type Depth = Natural
 -- | Block score.
 type Score = Integer
 
+data Unsealed = Unsealed deriving (Generic, Show)
```
cloudhead: Haddocks would be nice for this.

src/Oscoin/API/HTTP/Client.hs
adinapoli-mndc: This was a deliberate design decision from my side, as yes, it leads to very big constraints, but they are all unrolled in one place and not hidden behind a type alias which then contains dozens of other instances. Plus, being so fine grained means we are paying only from what we need -- If you look at the `Oscoin.Test.Crypto` module you will see I did something similar by creating a kitchen sink `IsCrypto` alias.

While we could potentially do something similar, we are now making most of our functions needlessly constrained also on stuff they don't directly use. If we start breaking down each individual functionalities into smaller type alias we eventually bundle all them together, we are back into those crazy layer of indirections where you have to expand 10 different type synonyms to get the info you want.

At the end of the day I think it comes down to personal taste, and I am happy to stick with what the majority thinks 😉

src/Oscoin/Consensus/Nakamoto.hs
adinapoli-mndc: I think I couldn't use `hasPow` directly due to a type mismatch, but maybe now that the dust settled I can give this another spin.

src/Oscoin/Crypto.hs
adinapoli-mndc: The reason being we need to import it inside the `Oscoin.Crypto.Hash.Mock` and `Oscoin.Crypto.PubKey.Mock`. Unless we also move them into the test hierarchy, we need this guy here.

src/Oscoin/Crypto/PubKey.hs
```diff
@@ -1,190 +1,104 @@
...
+ Types
+------------------------------------------------------------------------------}
 
-newtype PrivateKey = PrivateKey ECDSA.PrivateKey
-    deriving (Show, Eq)
+-- | A 'PublicKey' over a certain crypto 'c' and an algorithm.
+-- /DO NOT/ use this type directly, it's exported only for the sake of
+-- implementing the 'HasDigitalSignature' instance in 'PubKey.RealWorld'.
+-- Use the abstract PK/SK instead.
+data PublicKey c k = PublicKey k (Crypto.Hashed c k)
```
cloudhead: Hmm, seems easy to miss the docstring and use this type - any ideas how to prevent that?

src/Oscoin/Crypto.hs
adinapoli-mndc: Not particularly attached to the `RealWorldCrypto`, so I am happy to rename it if `Crypto` is more fitting 😉

src/Oscoin/Crypto/Blockchain/Block.hs
adinapoli-mndc: Totally, thanks for the remainder!

src/Oscoin/Crypto/Blockchain/Block.hs
adinapoli-mndc: Also true 😁

src/Oscoin/Crypto/PubKey.hs
adinapoli-mndc: On top of my head the only way to do that would be to move _both_ the `MockCrypto` and `RealWorldCrypto` implementations into this module, which would become a bit messy, though.

Another "sweet-spot" solution might be to move `PublicKey` and `PrivateKey` inside an `Internal` module, a bit like you did with the SQLite transaction stuff, to stress the fact it shouldn't be used directly. 

However, now we _never_ reference directly a `PublicKey` in the types, so users trying to actually use this type would have GHC complain that it cannot unify `PublicKey c k ~ PK c`. The risk is more on novel code written by contributors not familiar with this abstraction.

test/Oscoin/Test/Crypto.hs
```diff
@@ -0,0 +1,100 @@
...
+                  , AuthTree.MerkleHash (Hash c)
+                  , Ord (Hash c)
+                  , ByteArrayAccess (BlockHash c)
+                  , Buildable (Hash c)
+                  , ToField (BlockHash c)
+                  , ToField (Hashed c (RadTx c))
+                  , ToField (Sealed c Text)  -- DummySeal ~ Text
+                  , FromField (Hash c)
+                  , FromField (Sealed c Text)  -- DummySeal ~ Text
+                  )
```
cloudhead: :open_mouth:

src/Oscoin/API/HTTP.hs
cloudhead: Currently we have a separate node for testing (`TestNode`) - so I would monomorphise for now

src/Oscoin/API/HTTP/Client.hs
cloudhead: I wasn't suggesting a type alias, but rather a way to tell the compiler that given a `Serialise c`, and for ex. `instance Serialise c => Serialise (Block c)`, we wouldn't need `Serialise (Block c)` everywhere. Maybe it can't be done?

src/Oscoin/Crypto.hs
cloudhead: I see I see. Let's leave it then for now :+1:

src/Oscoin/Crypto/PubKey.hs
cloudhead: Moving to internal sounds like a good idea - I'd also think that calling the internal ones PK/SK makes more sense, while PublicKey/PrivateKey are more readable for external usage. Maybe it's worth creating an issue and tagging it with [easy]?

src/Oscoin/Consensus/Nakamoto.hs
cloudhead: I think it's worth re-jiggling the implementation of `mineNakamoto` and/or `hasPoW` to make it work, if needed.

src/Oscoin/API/HTTP/Client.hs
adinapoli-mndc: Ah, sorry, I did got the impression I was answering the wrong question while typing it 😉 

Alas it can't be done, and if you scroll through the code you will see I have done this as-aggressively-as-possible for instances which could actually be written that way.

In this particular example, if we take a look at:

``` haskell
Serialise (BlockHash c), Serialise (PK c), Serialise (Signature c)
```
The blocker here is that both `BlockHash` is just an alias for `Hash`, and the latter, `PK` and `Signature` are _all_ type families, which makes simply not possible (at least not in a way that I know about!) to write such typeclass constraints the way you want 😢 .

This is one drawback of the current approach I am afraid 😞 you gain power, but you also gain fatter constraints. There is no free lunch alas!

cloudhead:
Ok, done reviewing - really good stuff. Overarching concerns are:

1. Undecidable instances is now enabled in most important modules - not sure if this is a problem or not, but it does mean it's easier to do the wrong thing and have the compiler not complain now. I can't say I have that much experience with this though, so I'll defer to you and @kim 
2. Very large type signatures in some places that hurt readability - not sure exactly if there's a good solution (I don't like constraint aliases either) - but maybe it's possible to have ghc infer these types somehow.

Otherwise :+1:

test/Oscoin/Test/Crypto.hs
adinapoli-mndc: also known as "the sin-bin, the Cthulhu on earth, the thing you don't want to look eat or it will devour your soul." 😁

src/Oscoin/Crypto/Hash/RealWorld.hs
kim: Comment still stands :)

src/Oscoin/Crypto.hs
cloudhead: I'd be in favor of just `Crypto`, because we don't have stuff like `RealWorldGossip` and `RealWorldConsensus` in the codebase, so might raise eyebrows.

src/Oscoin/Crypto/PubKey.hs
```diff
@@ -1,190 +1,104 @@
...
-unsign = sigMessage
-
--- | Verify a signed message with the public key.
-verify :: Serialise msg => PublicKey -> Signed msg -> Bool
-verify key s = verifyBytes key $ map (LBS.toStrict . CBOR.serialise) s
-
-verifyBytes :: PublicKey -> Signed ByteString -> Bool
-verifyBytes (PublicKey key _) (Signed bytes (Signature sig)) =
-    ECDSA.verify Crypto.hashAlgorithm key sig bytes
+-- | Extracts a signed message out of the 'Signed' opaque type.
```
kim:
```suggestion
-- | Extracts the message payload out of the 'Signed' opaque type.
```

src/Oscoin/Telemetry.hs
```diff
@@ -143,89 +141,90 @@ emit Handle{..} evt = withLogger $ GHC.withFrozenCallStack $ do
                           (HTTP.queryString req)
                           duration
 
-        GossipEvent    e -> Log.withNamespace "gossip" $ traceGossip e
-        HandshakeEvent e -> Log.withNamespace "p2p"    $ traceHandshake e
+        -- NOTE(adn): Unfortunately I had no choice, as part of oscoin#368
```
kim: ✅

test/Oscoin/Test/Crypto.hs
kim: Yummy!

src/Oscoin/Crypto/Hash/RealWorld.hs
adinapoli-mndc: Thanks for the reminder! You were right, the constraint was unnecessary but alas we still need `UndecidableInstances` for the `Serialise` instance in this module. Oh well 😉

exe/node/Main.hs
adinapoli-mndc: cf21232

src/Oscoin/API/HTTP.hs
adinapoli-mndc: cf21232

src/Oscoin/API/Types.hs
adinapoli-mndc: cf21232

src/Oscoin/Consensus/Mining.hs
adinapoli-mndc: cf21232

src/Oscoin/Consensus/Nakamoto.hs
adinapoli-mndc: cf21232

src/Oscoin/Crypto.hs
adinapoli-mndc: cf21232

src/Oscoin/Crypto/Blockchain/Block.hs
adinapoli-mndc: ⚠️ Will open an issue as soon as we merge (the emoji serves to me as a visual cue).

src/Oscoin/Crypto/Blockchain/Block.hs
adinapoli-mndc: cf21232

src/Oscoin/Crypto/PubKey.hs
adinapoli-mndc: cf21232

⚠️ Will open an issue for the renaming as soon as we merge (visual cue for my own benefit).

src/Data/ByteArray/Orphans.hs
```diff
@@ -0,0 +1,27 @@
...
+import           Data.Text (Text)
+import           Data.Text.Encoding (encodeUtf8)
+
+import qualified Codec.Serialise as CBOR
+
+import           Data.ByteArray (ByteArrayAccess(..))
+import qualified Data.ByteString.Lazy as LBS
+import qualified Radicle.Extended as Rad
+
+-- NOTE(adn) This is temporary as @kim reported the latest version of the
```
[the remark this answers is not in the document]
adinapoli-mndc: cf21232

src/Oscoin/Crypto/PubKey.hs
adinapoli-mndc: cf21232

adinapoli-mndc:
@cloudhead addressed your (and Kim's) comments, give this another go and let's merge it! 💪

kim:
:shipit:

kim:
Oh, another follow-up: the multi-param typeclasses without fundeps make me nervous. We should figure out if we can remove them altogether.

adinapoli-mndc:
> Oh, another follow-up: the multi-param typeclasses without fundeps make me nervous. We should figure out if we can remove them altogether

Good point. In the same spirit of what we did for the `BlockStore`, this might be the chance for getting rid of the `MonadXXX` typeclasses in favour of just record-of-functions.

PR #386, opened 2019-02-27
Gossip Telemetry

kim:
Depends on oscoin/gossip#17

### DIFF
--- a/src/Oscoin/Telemetry.hs
+++ b/src/Oscoin/Telemetry.hs
@@ -21,13 +21,21 @@ import           Formatting.Buildable as F
 import           GHC.Stack as GHC
 import           Lens.Micro
 import           Network.HTTP.Types as HTTP
+import           Network.Socket (SockAddr)
 import           Network.Wai as HTTP
+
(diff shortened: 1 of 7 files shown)


src/Oscoin/Telemetry.hs
```diff
@@ -208,6 +296,8 @@ toActions = \case
                         defaultBuckets
                         (fromIntegral duration / fromIntegral Time.seconds)
      ]
+    GossipEvent _ -> []
+    HandshakeEvent _ -> []
```
adinapoli-mndc: I suspect something we could do is at least increment some counter associated with some failed event, both at the gossip or the handshake level. Sounds plausible/feasible?

kim: This actually brings up a shortcoming of our current system: on the gossip level, the most interesting metric operationally is the node's current partial view of the network (# of active/passive peers). Deducing this from the events is very error prone, we should rather be able to query the state whenever metrics are requested. 

Will need to think about this more.

PR #394, opened 2019-03-06
Lift the requirement for pre-shared keys of gossip seeds

kim:
Based on oscoin/gossip#18, a node can now join the network by simply passing one or more `--seed=<host>:<port>`.

This is the first step towards a more magical zeroconf autodiscovery.


exe/node/Main.hs
```diff
@@ -118,16 +121,12 @@ main = do
...
-    seeds'      <- Yaml.decodeFileThrow seeds
     config      <- Consensus.getConfig environment
 
     metricsStore <- newMetricsStore $ labelsFromList [("env", toText environment)]
     forkEkgServer metricsStore ekgHost ekgPort
 
-    withStdLogger  Log.defaultConfig { Log.cfgLevel = Log.Debug -- TODO(adn) Make it configurable
-                                     , Log.cfgStyle = Log.styleFromEnvironment environment
-                                     } $ \lgr -> Log.withExceptionLogged lgr $
+    withStdLogger (Log.configForEnvironment environment) $ \lgr -> Log.withExceptionLogged lgr $
```
adinapoli-mndc: Good man, fixing my TODOs! 😁

exe/node/Main.hs
```diff
@@ -41,7 +41,7 @@ data Args = Args
     { host          :: HostName
     , gossipPort    :: PortNumber
     , apiPort       :: PortNumber
-    , seeds         :: FilePath
+    , seeds         :: [P2P.NodeAddr Maybe Crypto]
```
adinapoli-mndc: Small nit (which boils down more to personal taste): I find hard to reason about this type due to the presence of `Maybe`. Once I look at the type it's easy to understand that the first type param is a functor `f`, but it might not be totally obvious for engineers working on code which touches only the boundary of the P2P layer.

I know it's always a delicate balance between indirection & overhead, but do you think we should type alias this guy in a way that the function is concealed, if we expect in non-library code to always be the same? 🤔

kim: If you could ask your fresh names oracle for a good name for that type alias? Mine ran out of names and yields only random character sequences...

adinapoli-mndc: I was leaning towards `NodeAddress`, but that seems error prone if we also don't rename `NodeAddr` 🤔 . `NodeSeed` maybe? It could fit quite well with the current data record field named `seeds`?

kim: 6a388e3c0fe859704eb5c7ccfb90a6d0d3c876f1

src/Oscoin/CLI/Command.hs
```diff
@@ -21,16 +21,15 @@ import qualified Oscoin.Crypto.Hash as Crypto
...
 import           Oscoin.Data.Tx (mkTx)
-import           Oscoin.P2P (NodeAddr(..), mkNodeId)
 import           Oscoin.Time (Timestamp)
 
 
 import           Codec.Serialise (Serialise)
 import qualified Crypto.Data.Auth.Tree.Internal as AuthTree
 import           Crypto.Random.Types (MonadRandom)
 import           Data.ByteArray (ByteArrayAccess)
+import           Data.ByteArray.Orphans ()
```
kim: For posterity: GHC mysteriously demanded this to be added

PR #398, opened 2019-03-06
LTS 13.10

kim:
GHC forced me to upgrade radicle, and then apply some dubious adjustments.

### DIFF
--- a/src/Oscoin/Data/RadicleTx.hs
+++ b/src/Oscoin/Data/RadicleTx.hs
@@ -22,6 +22,7 @@ import           Oscoin.Data.Tx (Tx(..))
 
 import           Codec.Serialise (Serialise)
 import           Data.Aeson (ToJSON, toJSON)
+import           Data.Copointed (copoint)
 import           Data.Default (Default(..))
 import qualified Data.IntMap as IntMap
 import qualified Data.Map as Map
@@ -33,8 +34,7 @@ import qualified Text.Show as Show
 newtype Env c = Env { fromEnv :: Rad.Bindings (Rad.PrimFns Identity) }
 
 instance Show.Show (Env c) where
-    show (Env inner) =
-        T.unpack . Rad.prettyValue . Rad.toRad $ Rad.bindingsEnv inner
+    show = show . Rad.bindingsEnv . fromEnv
 
 type RadTx c = Tx c Rad.Value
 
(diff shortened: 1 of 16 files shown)


src/Oscoin/Telemetry/Logging.hs
```diff
@@ -454,8 +454,7 @@ fprocessId = F.mapf (fromIntegral @CPid @Int32) (ftag "pid" % F.int)
 -- | When given a function from 'a' to an error message 'Text', format
 -- a domain-specific error as @error_class=foo_bar error_message=\"The error\"@
 -- where the @error_class@ is generically derived from the data constructors.
-ferror :: (SOP.Generic a, SOP.HasDatatypeInfo a)
-       => (a -> Text) -> Format r (a -> r)
+ferror :: SOP.HasDatatypeInfo a => (a -> Text) -> Format r (a -> r)
```
adinapoli-mndc: Interesting, so looks like newer versions of `generics-sop` do not necessitate anymore on the `SOP.Generic` constraint? Good to know 🤔

src/Radicle/Conversion.hs
```diff
@@ -43,22 +44,27 @@ typeMismatch expected actual =
...
+        Rad.LambdaRec{}  -> "LambdaRec"
+        Rad.ProcHandle{} -> "ProcHandle"
+        Rad.Handle{}     -> "Handle"
 
-instance ToRad a => ToRad (Set a) where
+instance ToRad WithPos a => ToRad WithPos (Set a) where
     toRad = toRad . Set.toList
 
-instance (FromRad a, Ord a) => FromRad (Set a) where
+instance (FromRad WithPos a, Ord a) => FromRad WithPos (Set a) where
```
geigerzaehler: I think `WithPos` can be replaced by a type variable.

kim: Here it can, but since we effectively always need `WithPos` I though that is less confusing

src/Oscoin/Data/RadicleTx.hs
```diff
@@ -61,10 +61,10 @@ instance Query (Env c) where
 
     query path (Env bindings) = do
         ident <- Rad.mkIdent (T.intercalate "/" path)
-        val <- Map.lookup ident (Rad.fromEnv $ Rad.bindingsEnv bindings)
-        case val of
+        val   <- Map.lookup ident (Rad.fromEnv $ Rad.bindingsEnv bindings)
+        case copoint val of
```
cloudhead: :eyes:

kim:
@cloudhead 

image

src/Oscoin/Data/RadicleTx.hs
kim: http://cokmett.github.io/cokmett/

cloudhead:
Oh that's right! :open_mouth:

src/Oscoin/Data/RadicleTx.hs
cloudhead: :joy:

adinapoli-mndc:
@kim best commit messages ever 😂

kim:
Heads up: this is now using a `stack` "custom snapshot" for non-stackage dependencies. This makes the caching shenanigans a bit more robust. Stolen from @geigerzaehler's radicle build.

adinapoli-mndc:
🎉 ! Now I can use `-hr` in my profiling efforts, at last 😂

PR #399, opened 2019-03-07
Fix block score

### LINKED ISSUE #367: Fix block score calculation
The current way we're calculating score is to just take the `height`. This is incorrect. We need to sum the target difficulty of each block.

cloudhead:
* Remove notion of generic block score
* Scoring function is always passed into BlockStore
* In tests, use Nakamoto's scoring function
* Consistently use `ScoreFn` and `Score` types

Fixes #367


test/Oscoin/Test/Storage/Block/Equivalence.hs
```diff
@@ -25,6 +21,11 @@ import           Oscoin.Test.Crypto.Blockchain.Generators
 import           Oscoin.Test.Storage.Block.SQLite (DummySeal, defaultGenesis)
 import           Oscoin.Test.Util (Condensed(..), showOrphans)
 
+import           Data.ByteArray.Orphans ()
+import qualified Data.List as List
+import qualified Data.Text as T
+import           GHC.Exception
```
adinapoli-mndc: What do we need the `GHC.Exception` import for? 🤔

cloudhead: `srcLocFile` and `srcLocStartLine` apparently

cloudhead: Made them explicit :+1:

PR #400, opened 2019-03-07
Fixed #369, remove "once" calls in the testsuite

### LINKED ISSUE #369: Remove `once` calls in test suite

adinapoli-mndc:
As this is approaching the point of diminishing returns, I have to draw a line here. Performance is now decent, running the full testsuite with each property generating 100 tests takes ~40 seconds on my machine. However, I think this is still too long, but alas I suspect we have to park things for the moment.

In particular, the biggest perf gains were yielded by:

a. Aggressively simplifying the `Arbitrary Rad.Value` instance. I think this is OK as Radicle is probably going to be removed sooner rather than later;

b. I have switched to a [patched version of qc-instances](https://github.com/oscoin/qc-instances/commit/e659eb65a421a91603de4d7645d370abf19aeeef) which provides a much faster instance for `Arbitrary ByteString`.

In case you wonder, here is now the partitioning of time for the whole test suite:

screenshot 2019-03-07 at 13 14 47

As you can see the biggest bottlenecks are the `hashTxs` function and the `arbitrary` call inside `Arbitrary Rad.Value`. In order to fix the performance of the former we would need to either review our serialisation choices (and perhaps experiment a bit with protobuf) or to find a way to remove the offending call to `Codec.serialise` inside the `hashTxs`.

## Bonus

I did try to experiment with `splitmix` either to provide a fast random byte generation and avoid the artificial `MonadRandom Gen` instance and both by patching up `qc-instances` to use it, but the performance gain was either negligible or worse, so I did abandon this effort.


adinapoli-mndc:
```
Please report this as a GHC bug: http://www.haskell.org/ghc/reportabug
(GHC version 8.4.4 for x86_64_unknown_linux)
test: internal error: Unable to commit 1048576 bytes of memory
No depth specified
```

Woah, this was.. unexpected.

adinapoli-mndc:
Ok, something is off here 😲

adinapoli-mndc:
Dang:

screenshot 2019-03-07 at 14 52 19

adinapoli-mndc:
Will continue on Monday, but this is looking a bit more promising now:

screenshot 2019-03-07 at 17 54 42

We still have a bit of a "cuspid", but the large PINNED block is gone.

adinapoli-mndc:
Alright, I will finish on Monday. Test passes and we should be good but I suspect (I might be wrong ofc) that there is still one memleak, this time around the HTTP helpers / STM blockstore.

src/Oscoin/Crypto/Hash/Mock.hs
```diff
@@ -34,9 +34,10 @@ import           Text.Read (readMaybe)
 import qualified Text.Show as Show
 import           Web.HttpApiData (FromHttpApiData(..))
 
--- For our mock implementation we use a fast, non-cryptographic hash
--- See: https://cyan4973.github.io/xxHash/
-import           Data.Digest.XXHash.FFI
+-- For the mock crypto we use the FNV-1 hashing function.
+-- Unfortunately using something faster like xxhash would generate
+-- too many collisions.
+import           Data.ByteArray.Hash
```
cloudhead: Can you explicit import `FnvHash` from here? Not clear ideas where it's coming from :smile:

cloudhead:
> Will continue on Monday, but this is looking a bit more promising now:

Nice work!

src/Oscoin/Crypto/Hash/Mock.hs
adinapoli-mndc: Yup, by all means! I was trying as a last valiant attempt to get rid of the final space spike, but no dice so far, and will stop soon anyway as I want to move to #402 quite quickly. My intuition is that there is still _something_ being retained and not fully evaluated when we insert a block in the pure `BlockStore`, but so far deliberate use of NFData didn't yield the expected result. Oh well.

adinapoli-mndc: 149d81d

adinapoli-mndc:
@cloudhead @kim OK, I think this is good to go in now (module Kim's veto, of course).

I have tried to further reduce the initial allocation but I suspect this starts to look like as if that's the physiological data allocation pattern for our testsuite. After all, we generate lots of transactions that we keep in memory in the STM-baked, pure `BlockStore` and each `Block` contains a list of `RadTx` which is itself quite beefy (due to the annotation it carries around, I suspect) plus the signed message, which means other extra bytes carried around.

I have empirically tested allocations using my Mac's "Activity Monitor" and I can see now that the leak is gone as the memory for the process grows to ~1GB but then remains stable till the end of the test run.

Having said that, we should periodically re-assess and do a couple of rounds of profiling to make sure we keep time & space at bay.

adinapoli-mndc:
> Arriving at FNV-1 and [Word8] are surprising results.

Probably swapping `xxhash` for `fnv-1a` wasn't a determining factor, but as I was trying all different solutions and running times were comparable (and I was hoping collisions were lower) I took the plunge. In hindsight I think we did the right thing -- the `xxhash` implementation was doing a bunch of FFI stuff and I am always very paranoid about finalisers being run and folks doing the due diligence in creating leak-free C shims. The `[Word8]` seems to have helped at least in removing the `PINNED` entry from the heap profile.

Considering that both bytestrings OR FFI stuff are both mapped into the heap profile as PINNED memory, at the end of the day having those two changes in the code makes a little bit more easier trying to pinpoint who is retaining the memory rather than storying blankly at a large portion of pinned memory 😁 

In any case, we can re-assess in the future.

PR #401, opened 2019-03-07
Add and fix docs on consensus modules

### LINKED ISSUE #54: Document consensus-related modules
The network simulator, simple node and Nakamoto node need haddocks, as well as the consensus-related type classes.

- [ ] Document BlockHeader and the `s` parameter.

cloudhead:
Closes #54


src/Oscoin/Consensus.hs
```diff
@@ -34,8 +34,8 @@ validateBlockchain validateBlock (Blockchain (blk :| blks)) =
     validateBlock blks blk *> validateBlockchain validateBlock (fromList blks)
 
 
--- | Validates the size (in bytes) for a block, comparing the serialised size
--- with the maximum allowed value as read from the configuration file.
+-- | Validates the size (in bytes) of a block, comparing the serialized size
```
adinapoli-mndc: so are we standardising on US spelling? :trollface:

src/Oscoin/Consensus/Nakamoto.hs
```diff
@@ -171,14 +179,14 @@ difficulty
 difficulty = encodeDifficulty . os2ip . headerHash
 
 -- | Number of blocks to consider for difficulty calculation. Roughly 2 weeks
--- withh a 10 min block time.
+-- with a 10 minute block time.
```
adinapoli-mndc: 10 _minutes_ perhaps?

src/Oscoin/Consensus.hs
cloudhead: Yeah aka "International English" :thinking:

src/Oscoin/Consensus/Nakamoto.hs
cloudhead: But it's '**a** 10 minute block time'

adinapoli-mndc: oh, right, right, sorry 'bout that 😁

Issue #402, opened 2019-03-11
Finalise architectural refactoring


adinapoli-mndc:
This ticket is about slightly reshuffling components together such that the fork selection code can be pulled out of the `SQLite.Internal` module and we have a clear, layered architecture where dependency boundaries are well-defined. This is a rough sketch of what we are aiming for:

![IMG_20190219_151814](https://user-images.githubusercontent.com/45846748/54119916-d2d98f80-43f6-11e9-9511-c54be994bb27.jpg)

adinapoli-mndc:
Delivered in https://github.com/oscoin/oscoin/pull/407

PR #407, opened 2019-03-12
Refactor storage/consensus architecture

adinapoli-mndc:
Nothing too exciting for now, but shaping up nicely.


snapshot.yaml
```diff
@@ -16,6 +16,7 @@ packages:
 - stm-containers-0.2.16@sha256:e98efa8dcf0045ea8a78a04b4e2763cf2d8bc33aad0750e2f30a67f8f4e933b1
 - tasty-fail-fast-0.0.3
 - tasty-hedgehog-0.2.0.0@sha256:83a8b777fa472040979e44dba43c32441f55d5ddb9641a4d53deee4b0e09fa34
+- discrimination-0.3@sha256:6f88488e62fd4aca71497a4909371476eb18720b3c113869e769a65d1bdb69dc
```
adinapoli-mndc: Note to myself: this is a relic, needs to go away.

src/Oscoin/Protocol.hs
```diff
@@ -0,0 +1,292 @@
...
+                         selectBestChain mgr
+
+             -- NOTE(adn) This is a side effect of the fact that storing
+             -- the genesis as an orphan would insert it into the orphanage with
+             -- its parent hash, which means that we won't be able to retrieve
+             -- a potentiol candidate branching off from genesis in our 'selectBestChain'
+             -- function (cfr. QuickCheck seed 554786). To avoid that, we avoid
+             -- inserting the genesis block into the orphanage in the first place;
+             -- After all, the genesis block should never be considered at orphan.
+             else pure $ if isGenesisBlock incomingBlock
```
adinapoli-mndc: Ops, this is a bug, this should be a `not isGenesisBlock`.

adinapoli-mndc:
Ok, this is _almost_ ready, I have only 1 test between me and endless glory:

```haskell
        confirmed transaction: FAIL
          test/Oscoin/Test/API.hs:95:
                  Ok (TxLookupResponse {
                      txHash = Hashed "Ex6CsgC",
                      txBlockHash = Just "DZCfsGL",
                      txOutput = Just (Right (Annotated (WithPos (InternalPos "src/Radicle/Internal/Core.hs:336:14 annotate
                  test/Oscoin/Test/API.hs:80:28 String") (StringF "jo")))),
          have ->     txConfirmations = 3,
          want ->     txConfirmations = 6,
                      txPayload = Tx {
                          txMessage = Signed {
                              sigMessage = Annotated (WithPos (InternalPos "src/Radicle/Internal/Core.hs:336:14 annotate
                  test/Oscoin/Test/API.hs:80:28 String") (StringF "jo")),
                              sigSignature = MockSignature [
                                  143,
                                  215,
                                  25,
                                  92,
                                  134,
                                  159,
                                  35,
                                  153
                              ]
                          },
                          txPubKey = MockPK (PK (MockKey [
                              143,
                              215,
                              25,
                              92,
                              134,
                              159,
                              35,
                              153
                          ]) (Hashed "NfbENm8")),
                          txChainId = 0,
                          txNonce = 0,
                          txContext = "1111111"
                      }
                  })

1 out of 2 tests failed (0.01s)
```

It looks like the number of confirmations is wrong, and it oscillates (sometimes it's 2, sometimes 4 etc), so I suspect this is due to the introduction of `runOscoinProtocol`. Digging deeper 👀

src/Oscoin/Storage/Block/Cache.hs
```diff
@@ -0,0 +1,65 @@
...
+    BlockCache { getBlockCache :: MVar (CircularSeq (Block c tx s)) }
+
+cached
+    :: BlockCache c tx s
+    -> Int
+    -- ^ The number of requested elements
+    -> (Int -> IO [Block c tx s])
+    -- ^ A function to \"backfill\" the cache by the number of missing
+    -- elements.
+    -> IO [Block c tx s]
```
geigerzaehler: How about using `Oscoin.Time.Chrono` here, too? We should point out that we request the X most recently added elements.

src/Oscoin/Time/Chrono.hs
```diff
@@ -0,0 +1,42 @@
+{-- | Little helpers geared towards making obvious in the types
+      how blocks are ordered.
+--}
```
geigerzaehler: Could you add a more detailed description and examples how to use this?

src/Oscoin/Time/Chrono.hs
```diff
@@ -0,0 +1,42 @@
...
+    ) where
+
+import qualified Data.List as List
+import qualified Data.List.NonEmpty as NE
+import           Oscoin.Prelude hiding (reverse)
+
+
+class ChronogicallyOrdered a where
+    type ReverseOf a :: *
+    reverse :: a -> ReverseOf a
```
geigerzaehler: It looks to me that `a` must be either `NewestFirst f` or `OldestFirst f`. Who about making this class less general
~~~haskell
class ChronologicallyOrdered f where
  toOldestFirst :: NewestFirst f -> OldestFirst f
  toNewestFirst :: OldestFirst f -> NewestFirst f

src/Oscoin/Storage/Block/Abstract.hs
```diff
@@ -8,36 +8,60 @@
...
+      ByDepth Depth
+    -- ^ Get a collection of blocks from the input 'Depth' up to the tip.
+    | ByRootHash (BlockHash c)
+    -- ^ Get a collection of blocks starting from the input 'BlockHash' up to
+    -- the tip.
+
+-- | The internals for this 'BlockStore', i.e. the private API which shouldn't
+-- be visible to \"read-only consumers\".
+data BlockStoreInternals c tx s m = BlockStoreInternals
     { scoreBlock      :: Block c tx (Sealed c s) -> Score
```
geigerzaehler: This should probably be moved to the protocol

src/Oscoin/Storage/Block/Abstract.hs
```diff
@@ -8,36 +8,60 @@
...
 import           Oscoin.Crypto.Hash (Hashed)
+import           Oscoin.Time.Chrono (NewestFirst, OldestFirst)
 
--- | An handle over a block storage backend.
-data BlockStore c tx s m = BlockStore
+-- The full 'BlockStore' API is the composition of its public and private parts.
+type FullBlockStoreAPI c tx s m =
+    ( BlockStore c tx s m
+    , BlockStoreInternals c tx s m
+    )
```
geigerzaehler: I was thinking about the naming here. How about the following?
~~~haskell
type BlockStore = (BlockStoreModifier, BlockStoreReader)
~~~

src/Oscoin/Storage/Block/Pure.hs
```diff
@@ -100,6 +102,14 @@ getBlocks (fromIntegral -> d) Handle{..} =
 -- Nb. we guarantee that there is at least one chain in the store by exposing
 -- only the 'genesisHandle' smart constructor.
 
+getChainSuffix
+    :: (Ord (BlockHash c))
+    => BlockHash c
+    -> Handle c tx s
+    -> [Block c tx (Sealed c s)]
```
geigerzaehler: Could you add some documentation here? The function would probably also benefit from `Oscoin.Time.Chronob.

src/Oscoin/Time/Chrono.hs
adinapoli-mndc: Aye, was in the back of my head as a polishing step, thanks for keeping me honest! 👍

src/Oscoin/Protocol.hs
```diff
@@ -0,0 +1,287 @@
...
+-- 'Protocol' directly but they are rather an implementation detail of
+-- whichever 'BlockStore' this manager encapsulates.
+data Protocol c tx s m = Protocol
+    { protoFullBlockStore :: FullBlockStoreAPI c tx s m
+    -- ^ The 'BlockStore', an opaque interface to a component which only
+    -- responsibility is to store and retrieve blocks.
+    , protoOrphanage      :: Orphanage c tx s
+    -- ^ The in-memory data structure which keeps orphans blocks and constructs
+    -- sub-chains suitable for fork selection.
+    , protoFetchNextBlock :: m (Maybe (Block c tx (Sealed c s)))
```
geigerzaehler: In the current implementation the action always blocks until a new block is available. I guess we can remove `Maybe`?

src/Oscoin/Storage/Block/Cache.hs
adinapoli-mndc: I've deliberately tried to go light with the usage of `Chrono` to not make this PR even beefier. Do you think we should bite the bullet or perhaps we can spin up a follow-up ticket, so that the next PR is way leaner?

src/Oscoin/Time/Chrono.hs
adinapoli-mndc: Interesting suggestion -- the only pain in the next is that now `toOldestFirst` and `toNewestFirst` clash with the "deconstructors", so I will have to rename them `getOldestFirst` and `getNewestFirst` and amend the code throughout, but it can be done 😀

src/Oscoin/Storage/Block/Abstract.hs
adinapoli-mndc: Yes, I should bite the bullet and be less lazy, I agree it doesn't belong here anymore.

src/Oscoin/Storage/Block/Abstract.hs
adinapoli-mndc: That's a reasonable suggestion (which will again have me to amend virtually every call site in the codebase 😁 ). I do agree names are important though; let's see what the other folks say before I embark into this journey. The rationale behind calling things the current way was to minimise disruption; furthermore, the `Internals` thing was not meant to be used anywhere (if not in test code) and its naming was predated from [other libraries](http://hackage.haskell.org/package/network-transport-inmemory-0.5.2/docs/Network-Transport-InMemory-Internal.html#v:createTransportExposeInternals) , but I am happy to stick with what we think is best 😉

src/Oscoin/Storage/Block/Cache.hs
geigerzaehler: Yes, it’s probably a good idea to leave this for later but please document the ordering

src/Oscoin/Protocol.hs
adinapoli-mndc: Having a `Maybe` there gives us more flexibility methinks, especially in tests. It doesn't cost us much, but (for example) gives us the ability to "peek" or trying to read into/from the queue (cfr. `peekTBQueue` and `tryReadTBQueue`).

src/Oscoin/Storage/Block/Abstract.hs
adinapoli-mndc: (addendum: personal preference -- if we go with your suggestion, I think I like `BlockStoreWriter` better 😁 )

adinapoli-mndc:
Happy Friday all!

@cloudhead @kim I promise this is the last beefy PR I issue (last famous words? 🤔 :trollface: ).

I think I am at the stage where I can call this "done", at least up to the point where we can start reviewing the new, proposed design (and @geigerzaehler indeed already started with some very useful comments).

As usual, the 10.000ft overview:

* We now have a dedicated data structure called `Protocol` which sits between the `BlockStore` and the node/gossip layer. What this does is essentially chain selection plus phase-2 validation & its the sole responsible for updating the `BlockStore`, which means it's the only one which has access to the BlockStore's private API. This is also why now the `BlockStore` has been split into two parts, namely its public and private API; the former is what we share with gossip & the miner, the latter is what is used by the `Protocol` to update the state of the world. 

* The `Protocol` per se can run in any monad and can be "stepped" in a "foldlM" style, but we expose a stateful, IO-friendly `Handle` to be used when we wire everything together. This is how things look like in the `Main.hs`:

```haskell
withStdLogger (Log.configForEnvironment environment) $ \lgr -> Log.withExceptionLogged lgr $
    BlockStore.Concrete.STM.withBlockStoreExposeInternals (fromGenesis gen) Nakamoto.blockScore $ \blkStore@(publicBlockStore, _) ->
        runOscoinProtocol GracefulShutdown blockDispatchQueue Nakamoto.validateFull (Log.err lgr) blkStore config $ do
            let telemetryHandle = Telemetry.newTelemetryStore lgr metricsStore
            withNode (mkNodeConfig environment telemetryHandle noEmptyBlocks config) [...]
```

* `runOscoinProtocol` is a function which spins up the async worker which read things (i.e. blocks) from outside and runs our protocol. Such worker is alive for the entire lifetime of the inner monadic blocks `runOscoinProtocol` wraps, and this is typically the endless loop of your application. When we Ctrl-C, the async worker is killed using one of two `ShutdownPolicy`. The `Graceful` one allows in-flight messages to be all processed.

* Validation has now been split in two parts: the "light" validation (named `validateBasic` now) happens inside `applyBlock`, after which point this "pre-validated" block is dispatched using a monadic action which is now passed as input to both `applyBlock` and `mineBlock`, and it's what "wires together" the gossip & the miner with the protocol.

* We have an in-memory cache sitting between the Protocol & the SQLite database, which kicks in when we call `getBlocks` and that's invalidated when we `switchToFork`. I ended up eating my dog food and using a very simple implementation of a circular queue I have used for another project of mine, but I am not attached to it, feel free to propose alternatives if that means less code to maintain.


I will wait for Alexis & Kim to have a chance to review this, and then I will start to tackle all the review comments at once. In the meantime, I am going to polish this one and run another profiling round to see if I have added any idiotic leak, as i could see now memory usage for our tests seems to be on the highish side.

cloudhead:
Exciting, will review this as soon as I can!

adinapoli-mndc:
Thanks for the always-very-well-thought-out review, @kim . On Friday, when submitting the PR, I had similar feelings as yours; specifically that this was an improvement (moving out chain selection, 2-phase validation & public/private API split seemed like genuine good things) but in a similar fashion when I had to insert that (provisional) artificial pauses in one of the tests, I knew something had to be reviewed. I will try to reply in-line to your original review as well as to the comments you provided below. Ok, ☕️ acquired, let's do that _rolls sleeves_. 

> A have a couple of reservations, however[..]

I think the take-home of this part is that inserting the queue feels morally "wrong" (or, at least, our data access pattern not yet such to justify that). I am not particularly attached to the queue itself and I went with it mostly as it was the design we settled upon in Berlin (see [screenshot](https://github.com/oscoin/oscoin/issues/402)). From one side, I like the idea that now the gossip layer doesn't have to wait for the full "blockStore + validation + chain selection", but I can see the concerns you are raising.

> The local miner and the network are now contending on the head of the queue (recall that there are no fairness guarantees provided by STM), where a locally mined block won't be broadcast until it got processed by the protocol/storage layers. I would think that this is a bad tradeoff.

While the fairness concern is surely true, I am slightly less concerned about this one, as when we call `dispatchBlock` all we have to do is to push the new block into the queue, without waiting for the protocol to fully process it. However, after a weekend of reflection, I am convinced that the miner _shouldn't be passed this side effect function_. In other terms, I think it was a bad design decision on my hand to hook the miner to the same queue the gossip was going to use. The reason is not just contention, but more that the data access of the two components are different: the gossip layer wants to push the block and forget about it, but the miner is actually dependent on the state of the world to be updated. To give you a concrete example, this is an excerpt of a test which was failing (and that I had to fix by inserting artificial pauses 🤢 ):

```haskell
            Just blk <- liftNode $ do
                Mempool.addTxs [tx]
                blk <- Node.mineBlock
                -- FIXME(adn) Unfortunately at the moment we need these
                -- artificial pauses to give time to the asynchronous
                -- `Protocol` to update the state of the chain (i.e. call
                -- `insertBlock`) so that the next call to `getTip` in the
                -- miner pulls the new block.
                liftIO $ threadDelay (1000 * 100)
                replicateM_ 5 (Node.mineBlock >> liftIO (threadDelay (1000 * 100)))
                pure $ blk
```

As explained in the comment, when we call `mineBlock`, the first thing we do is to call `getTip` and we "build upon" the previous block, so the precondition of this function is that the block has been already inserted in the store (as an orphan or as a tip-extension) by the time we call this function again. But this is simply not true with the current design, because in `mineBlock` we simply invoke the callback function which just push the block into the queue, without giving any guarantee of such block to be fully processed. But here we have _no reason_ to do that, as the miner already runs in its own thread and surely can wait for the `Protocol` to fully process the mined block.

> Do we have to perform fork selection for every block?

My intuition is that surely we do want to do that, because each new block that arrives in the system could trigger a "switch to fork" and if we don't do that, the next call to `getTip` (in the miner) would essentially not reflect an up-to-date state of the world. (But I might be wrong and @cloudhead can debunk me easily if needed 😅 ).

> Can we not just run fork selection in its own thread in a loop?

Oh, but this is exactly what happens -- when we call `runOscoinProtocol` we do spawn as `Async`. Are you suggesting something more subtle, like having the storage of the block to happen synchronously but fork selection to happen asynchronously? If that's your idea, my personal preference would be that we _don't go there_ -- reason being I would like to have a single point where we do DB persistence + chain selection, so that we don't expose ourselves to inconsistency created by concurrent interleaves. 

> As hinted at above, queueing makes it harder to reason about the state (how many blocks are in-flight?), and contributes to memory and GC pressure.

Very true, albeit it can be measured (and tracked via EKG) and it's bounded anyway.

> Obviously, some of these can be addressed in future patches. For this one to go in, I would like to see the following be addressed[...]

I don't have problems to amend my PR to do that -- if we scrap the queue, though, are we sure we want to do that also for the gossip end of things? This seems to be going in the opposite direction of what we agreed on-site 🤔

data/config.yaml
```diff
@@ -4,6 +4,7 @@
 
 production: &defaults
     maxBlockSize: 1000000 # 1 MB
+    mutableChainSize: 2016 # blocks
```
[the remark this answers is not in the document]
adinapoli-mndc: What do you have in mind exactly? The way I see things, you do want to have two kinds of configurations: one which is protocol specific and that can travel on-chain, and some part which is off-chain. While we may argue that `maxBlockSize` could have lived on chain, I suppose `mutableChainSize` is something we might want to leave in a separate config file.

Perhaps as part of a more luxurious, more decorated shed, we might argue that we might want to bake everything into the initial genesis file. This might feel abstruse but it's how some projects do that (my experience is only limited to Cardano obviously), i.e. [genesis.json](https://raw.githubusercontent.com/input-output-hk/cardano-sl/develop/lib/mainnet-genesis.json) (⚠️, very big file). It basically contained the money for the initial stakeholders, the start time of the chain, the certificates for the nodes, the _delegation_ certs (PoS stuff) and the "protocol configuration params", which is something at the moment we don't bake into genesis (but it lives in this `config.yaml` separate file).

src/Data/Sequence/Circular.hs
```diff
@@ -0,0 +1,142 @@
...
+import qualified Data.Sequence as Seq
+
+{------------------------------------------------------------------------------
+  Types and instances
+------------------------------------------------------------------------------}
+
+type Limit = Int
+type Size  = Int
+
+data CircularSeq a = CSeq !Limit {-# UNPACK #-} !Size !(Seq.Seq a)
```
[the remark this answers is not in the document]
adinapoli-mndc: As said in the 10.000ft overview of this PR I have borrowed this code from another project and this is what I did use at the time as I didn't need something terribly memory-efficient. I think the main reason to go with a `Seq` was that I did want something with good asymptotic performance for both cons and snoc, and that was pure, and this did fit the bill 😉 

>  Also, StrictData. Perhaps we should revisit our default-extensions practices?

Nah, this is just me being sloppy (and needless to say, this code goes back some versions of GHC where we didn't have `StrictData` yet 😁 ).

src/Oscoin/Consensus/Mining.hs
```diff
@@ -34,21 +35,23 @@ mineBlock
        , AuthTree.MerkleHash (Crypto.Hash c)
        )
     => BlockStore c tx s m
+    -> (Block c tx (Sealed c s) -> m ())
```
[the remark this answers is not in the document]
adinapoli-mndc: You are right. When I was writing some glue code for the tests, at some point I did ask to myself: wait, `mineBlock` is _already_ returning the mined block, so cannot I simply use it and do whatever I want with it without even passing the continuation here?

However, after I did refactor the code to get rid of it, I stopped and unwinded my work, mostly because it occurred to me that now is responsibility of _the caller_ to be sure the block is persisted, vs responsibility of just the `mineBlock` function. I am thinking here that perhaps it's a good thing if, as part of the `mineBlock` contract, we can guarantee that we write such logic in just one place, as it will shelter us from programmer's errors (i.e. forget to do the "right thing" with the block). What do you think?

src/Oscoin/Protocol.hs
```diff
@@ -0,0 +1,287 @@
...
+import           Oscoin.Time.Chrono as Chrono
+
+import           Control.Concurrent.Async (Async)
+import qualified Control.Concurrent.Async as Async
+import           Control.Concurrent.STM
+import qualified Data.List.NonEmpty as NonEmpty
+
+-- | A stateful handle over a 'Protocol'.
+data Handle c tx s = Handle
+    { hProtocol :: MVar (Protocol c tx s IO) }
```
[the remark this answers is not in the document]
adinapoli-mndc: My intention was to have exclusive access to the blockStore when processing a block, so that my calls to `extendsTip`, `getTip` weren't influenced by other things happening concurrently. 

When taking another look now, though, I do appreciate this feels redundant, especially in the presence of the blockstore API split. If anything, the `Protocol` is supposed to be "one and only" for the entire lifecycle of the application, so it's virtually the only component which write stuff into the blockstore.

src/Oscoin/Protocol.hs
```diff
@@ -0,0 +1,287 @@
...
+    :: forall c tx s a.
+       ( Ord (Hash c)
+       , Ord s
+       , HasHashing c
+       , Buildable (Hash c)
+       )
+    => ShutdownPolicy
+    -> TBQueue (Block c tx (Sealed c s))
+    -> Validate c tx s
+    -> (forall r. LogMsg -> Format (IO ()) r -> r)
```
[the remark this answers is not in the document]
adinapoli-mndc: You caught me. 🙈 🏃 💨

src/Oscoin/Storage/Block/SQLite.hs
```diff
@@ -2,7 +2,7 @@
 
 -- | Disk-backed block storage using SQLite.
 module Oscoin.Storage.Block.SQLite
-    ( withBlockStore
+    ( withBlockStoreExposeInternals
```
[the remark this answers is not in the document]
adinapoli-mndc: It was mostly a "tribute" to [createTransportExposeInternals](http://hackage.haskell.org/package/network-transport-inmemory-0.5.2/docs/Network-Transport-InMemory-Internal.html#v:createTransportExposeInternals) from the `network-trasport` package, but I am open to a new name -- perhaps piggybacking on @geigerzaehler 's naming suggestion, this should be simply called `withBlockStore` an would return a `(BlockStoreReader, BlockStoreWriter)`.

src/Oscoin/Storage.hs
```diff
@@ -69,21 +69,25 @@ applyBlock
        , Buildable (Crypto.Hash c)
        )
     => BlockStore c tx s m
+    -> (Block c tx (Sealed c s) -> m ())
```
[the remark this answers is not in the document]
adinapoli-mndc: I suppose nothing prevents me -- however, I was trying to be on the caution side and respect the original order of instructions (i.e. I have replace the old calls to `insertBlocks` with this infamous side-effect).

src/Oscoin/Protocol.hs
```diff
@@ -0,0 +1,287 @@
...
+    -> FullBlockStoreAPI c tx s m
+    -> Consensus.Config
+    -> (Protocol c tx s m -> m a)
+    -> m a
+withProtocol fetchNextBlock o validateFull bs config action =
+    let protocol = Protocol bs o fetchNextBlock validateFull config
+    in action protocol
+
+-- | Initialises and runs the 'Protocol', via its stateful 'Handle'.
+runOscoinProtocol
```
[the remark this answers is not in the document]
adinapoli-mndc: Yes, this is basically what I was thinking to do as well, great suggestion. This could be also used directly in the miner if we want to ensure the block is fully processed before we return from that function.

data/config.yaml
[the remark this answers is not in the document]
kim: Where I'm coming from is this: in the not-so-distant future, we want to deploy this application to a bunch of servers, preferably continuously. For the latter to work out, we want to make sure that the build catches as many issues as possible which could cause a faulty deployment. An often overlooked aspect of this is that the _configuration_ needs to be correct, too. If the configuration is scattered around a bunch of semi-structured text files, magic environment variables, _and_ a set of CLI options (some of which having default values), this actually becomes a rather hairy problem.

One solution, as you suggest, is to stick *everything* into a single config file, which is parsed into a strongly typed representation + some runtime validation if necessary. An actual configuration - either checked in to source control, or generated on the fly as part of the build - can now be tested by running the parser and validation _without running the application itself_. 

The other option I've used in the past is to stick *everything* into CLI options. `optparse-applicative`, for all it's quirks, is actually quite composable, and a given `Parser` can be evalutated purely. The advantage of this is that the configuration is always explicit (modulo default values), and visible even at runtime (just `ps`). A drawback is that secret configuration values have to still be resolved from files or via network calls, but that problem doesn't really go away with the mono-config-file approach (the file has to be rewritten at runtime).

As for this particular config file, it is not so clear if and when we would actually want to supply a different configuration from the outside. Why can't this be a function `defaultConfig :: Environment -> Config` in the code?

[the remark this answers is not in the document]
adinapoli-mndc: Right, thanks for explaining the rationale!

> As for this particular config file, it is not so clear if and when we would actually want to supply a different configuration from the outside. Why can't this be a function defaultConfig :: Environment -> Config in the code?

I suppose we totally can -- if we don't care about the virtue of having an amendable config file which can be modified without recompiling and releasing a new executable, for sure we can. Happy to go with what the team decides!

kim:
Hm okay. I think this is the culprit:

> the gossip layer wants to push the block and forget about it, but the miner is actually dependent on the state of the world to be updated.

In the _current_ design (prior to this patch), the gossip layer actually does wait until the block is inserted, because that is the only way currently to determine if a parent is missing, and we piggyback on this to implement a simplistic anti-entropy scheme. Note that this blocking occurs _per-connection_, so a peer sending us a lot of blocks in quick succession will receive backpressure, while the other active peers are not affected.

The miner, on the other hand, will see some snapshot of the world each time it enters the mining loop.

If we insert a shared (between miner and networking) queue, the observed behaviour is either the same if the frequency of incoming blocks is "normal" and the same on average for each connected peer, _or_ the queue is at capacity, in which case the miner has the same priority as the peers waiting to be enqueued. If I send you many blocks, I might be able to stall your miner (note that this doesn't have to be adversarial). _One_ more active peer can also stall all others (the queue is bounded, so enqueues will block if it is full).

Also, there are no stronger consistency guarantees: the miner is scheduled at some point in time, where "logically" the world may have advanced and it _should_ wait for some blocks to be applied before proceeding.

So either way, things are concurrent, and the best thing the miner can do is to block until its own block is processed before starting with the next one.

> Are you suggesting something more subtle, like having the storage of the block to happen synchronously but fork selection to happen asynchronously?

I think we have now agreed that the miner should block on the whole thing, right? So the only thing potentially left is that when a block comes in via the network, we insert it synchronously and then enqueue it (s hash) for a fork selection worker to pick up asynchronously. The only benefit I can see would be to achieve higher throughput of _transactions_ coming from the same peer. Not sure this is a concern at all.


**tl;dr** We do have convinced ourselves that our initial idea that there needs to be a queue somewhere causes more trouble than it solves. Or did we? 😸

src/Oscoin/Consensus/Mining.hs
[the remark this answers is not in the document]
kim: I think that we should appreciate the fact that the `Node` is our "App Monad", which is basically omniscient and wires up all the different components (perhaps we could even start to consider moving more of `main` into `Node.run`).

Since the `Node` controls the mining loop, and also knows about storage resp. protocol, we'd end up with exactly one place to wire the two together.

adinapoli-mndc:
> I think we have now agreed that the miner should block on the whole thing, right?

Aye, sir.

> So the only thing potentially left is that when a block comes in via the network, we insert it synchronously and then enqueue it (s hash) for a fork selection worker to pick up asynchronously. The only benefit I can see would be to achieve higher throughput of transactions coming from the same peer. Not sure this is a concern at all.

Right, which could be itself an improvement over this PR, by which I mean we don't have to necessarily commit on doing this async stuff _just now_, but rather as an optimisation step (I will see how easy/hard is to do this now or if it's better to refrain).

> tl;dr We do have convinced ourselves that our initial idea that there needs to be a queue somewhere causes more trouble than it solves. Or did we?

Yes, I am with you on this one. To recap, this is the design I am shooting for (shout if I am saying silly things):

* In the miner, we run `block storage + chain selection` synchronously, as that ensure that the subsequent `getTip` is consistent and things works the way they should;

* For the network (i.e., `applyBlock`) the baseline is that `block storage` must happen synchronously, but fork selection could happen asynchronously, if we don't want to slow down the gossip layer. Whether or not I will end up having a separate thread in `exe/Main.hs` which does that or not is something I can figure out while working on the amended design.

src/Oscoin/Consensus/Mining.hs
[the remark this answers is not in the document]
adinapoli-mndc: That's a good observation. The _only_ thing we should check is whether or not the execution order is important. Currently we have (inside `Consensus.Mining`):

```haskell
            for maybeSealedBlock $ \blk -> do
                Mempool.delTxs (blockData blk)
                dispatchBlock blk
                StateStore.storeState st'
                for_ receipts addReceipt
                pure blk
```

If `storeState` and `addReceipt` have as precondition that the block must be stored then we might need to keep the (synchronous) callback, otherwise doing this in `Node.mineBlock` is perfectly OK. I will try to assume the order is independent, and hopefully tests should catch any regression I might have introduced.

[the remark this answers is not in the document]
kim: Ah fair point!

kim:
> For the network (i.e., applyBlock) the baseline is that block storage must happen synchronously, but fork selection could happen asynchronously

Actually, I think you're free to implement this partially, fully, or not asynchronous at all. The property that two peer connections waiting on an `applyBlock` should have an equal chance to get scheduled should hold, though.

And now that I've wrote this, I realise that I've confused myself sufficiently about fairness for this case to warrant more meditation. Consider that we maintain a broadcast tree - hence, in a stable topology, this is much less concurrent than it might seem 🤔

adinapoli-mndc:
Alright, me and @kim paired a bit on this and I like the end result. For now I have skipped the "error handling" part we discussed together (namely dispatching a block synchronously for now yields `m ()`) but that's easy to add tomorrow once we reach consensus on whether or not we should now turn the miner "inside out" to return the mined block, the `st'` and the `receipts` so that we wouldn't update the state in the miner but rather outside it, in our `NodeT`. I think this will also allow us to remove a bunch of constraints like `MonadStateStore` from the `mineBlock` type, which is a plus.

Furthermore, I think we should (but @cloudhead might disagree) delete the transactions from the mempool only _after_ the block has been inserted and _only_ if such insertion successfully happened. For that, we will indeed need to modify `dispatchBlockSync` to return an `Either`.

I am getting a weird test failure on CI which I will investigate tomorrow (which, needless to say, doesn't happen locally, hehe).

src/Oscoin/Storage/Block/SQLite.hs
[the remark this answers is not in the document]
adinapoli-mndc: 5fa7b38

src/Oscoin/Telemetry/Logging.hs
```diff
@@ -351,6 +352,10 @@ errM :: (MonadLogger r m, HasCallStack) => LogMsg -> Format (m ()) a -> a
 errM = logM Err getLoc
 {-# INLINE errM #-}
 
+dontLog :: Applicative m => LogMsg -> Format (m ()) a -> a
```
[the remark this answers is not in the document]
adinapoli-mndc: I've got rid of it 😅  5fa7b38

src/Oscoin/Protocol.hs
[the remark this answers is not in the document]
adinapoli-mndc: 5fa7b38

src/Oscoin/Protocol.hs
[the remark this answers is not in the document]
adinapoli-mndc: We discussed this together and we agreed that indeed `stepProtocol` is a critical section as it performs the delicate task of extending the chain, and as such it makes sense to run in a mutex. (5fa7b38).

data/config.yaml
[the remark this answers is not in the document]
cloudhead: CLI-only is a good option in my opinion.

One thing I'd like to bring up though is that stuff like `maxBlockSize` and `mutableChainSize` *shouldn't* be configurable so easily. I think this should be a static variable in the source code. The reason being simply that it can break consensus if changed. In my opinion of course, "configuration" should be things you can change on your node without too much risk.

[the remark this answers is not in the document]
adinapoli-mndc: > shouldn't be configurable so easily[...] I think this should be a static variable in the source code.

Or, even better, stored on chain 😁 Without entering a bikeshed round, should I simply bit the bullet and add an extra commit to this (already gigantic) PR and simply scrap the read-from-file `Config` in favour of a fully static `Config`? I don't think it's going to be invasive as the external interface wouldn't change, only the fact now in `exe/Main.hs` we don't need to do IO to acquire a config. 👍 ? 👎  ?

cloudhead:
> Furthermore, I think we should (but @cloudhead might disagree) delete the transactions from the mempool only after the block has been inserted and only if such insertion successfully happened. For that, we will indeed need to modify dispatchBlockSync to return an Either.

Hmm - it seems like this will lead to an ever growing mempool of permanently failed transactions. We have to cleanup failed txs at some point.

> Do we have to perform fork selection for every block?

One thing we *could* do is to do it lazily, when the miner is looking for a tip. Not sure what the repercussions of this are though!

adinapoli-mndc:
@cloudhead @kim something worth pointing out is that after this PR our `oscoin` client fails to mine blocks:

Screenshot 2019-03-19 at 11 43 20

I am fairly convinced this patch now revealed a latent bug we had, which we didn't capture because no orphans passed the validation. Unless you think this is due to something I have triggered, I suggest we fix this on a separate patch.

data/config.yaml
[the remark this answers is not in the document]
cloudhead: Yeah, I think it's a good idea. If we want it configurable later we can add a CLI flag, but for now it can be static.

src/Oscoin/Consensus/Config.hs
```diff
@@ -21,13 +21,18 @@ configPath = "data/config.yaml"
 
 -- | Consensus configuration parameters.
 data Config = Config
-    { maxBlockSize :: !Int
+    { maxBlockSize     :: Int
     -- ^ The maximum block size in bytes.
+    , mutableChainSize :: Int
+    -- ^ The size, in blocks, of the \"mutable part\" of the chain,
+    -- i.e. the part of the chain still subject to rollbacks. In other words
+    -- this is the size of the not-yet-fully-adopted portion of the chain.
```
cloudhead: I'd use `Depth` here for the type and name: `mutableChainDepth :: Depth` for clarity. I think the last sentence of the docstring can also be removed, as it introduces a new concept ("adopted") which is a bit confusing imo.

src/Oscoin/Node.hs
```diff
@@ -133,14 +137,29 @@ mineBlock
...
     time <- currentTick
-    hoist liftIO $ Consensus.mineBlock (hoistBlockStore lift hBlockStore) hConsensus hEval time
+    res  <- hoist liftIO $ Consensus.mineBlock (hoistBlockStoreReader lift hBlockStore)
+                                               hConsensus
+                                               hEval
+                                               time
+    case res of
+      Nothing -> pure Nothing
+      Just (blk, st', receipts) -> do
+          -- NOTO(adn) Here we should dispatch the block and wait for the
```
cloudhead:
```suggestion
          -- NOTE(adn) Here we should dispatch the block and wait for the
```
:it: :question:

src/Oscoin/Node.hs
```diff
@@ -199,13 +219,17 @@ getPathLatest
...
         pure (stateHash, v)
 
-getBlocks :: (MonadIO m) => Depth -> NodeT c tx st s i m [Block c tx (Sealed c s)]
-getBlocks d = withBlockStore (`BlockStore.getBlocks` d)
+getBlocks
+    :: MonadIO m
+    => Depth
+    -> NodeT c tx st s i m [Block c tx (Sealed c s)]
+getBlocks d =
+  Chrono.toNewestFirst <$> withBlockStore (`BlockStore.getBlocks` ByDepth d)
```
cloudhead: I haven't gotten to the definition of `Chrono.toNewestFirst` yet, but this reads funny: why is a time-related function operating on blocks? I guess I'd expect `Blockchain.toNewestFirst` if anything.

src/Oscoin/Protocol.hs
```diff
@@ -0,0 +1,276 @@
...
+import qualified Oscoin.Telemetry as Telemetry
+import           Oscoin.Time.Chrono as Chrono
+
+import           Control.Concurrent.Async (Async)
+import qualified Control.Concurrent.Async as Async
+import           Control.Concurrent.STM
+import qualified Data.List.NonEmpty as NonEmpty
+
+-- | This data structures incorporates all the different components and
+-- data structures in the system devoted to store \"things\". Certain
```
cloudhead: This doesn't seem right: it's also used to validate blocks and score them.

src/Oscoin/Protocol.hs
```diff
@@ -0,0 +1,276 @@
...
+    -- ^ A validation function to (fully) validate a block.
+    , protoScoreBlock     :: Block c tx (Sealed c s) -> Score
+    -- ^ A function to score blocks.
+    , protoConfig         :: Consensus.Config
+    }
+
+
+-- | A 'Handle' is a record of functions containing two possible
+-- \"strategies\" to dispatch block; calling `dispatchBlockSync` ensures the
+-- function won't return up until the `Protocol` hasn't been fully stepped
```
cloudhead:
```suggestion
-- function won't return until the `Protocol` has been fully stepped
```

src/Oscoin/Protocol.hs
```diff
@@ -0,0 +1,276 @@
...
+-- (i.e. block inserted in the store, chain selection fully performed etc.)
+-- `dispatchBlockAsync` as the name implies, doesn't block and return the
+-- control to the caller immediately, but offer less integrity guarantees.
```
cloudhead:
```suggestion
-- control to the caller immediately, but offers less integrity guarantees.
```

src/Oscoin/Protocol.hs
```diff
@@ -0,0 +1,276 @@
...
+    -- ^ A function to score blocks.
+    , protoConfig         :: Consensus.Config
+    }
+
+
+-- | A 'Handle' is a record of functions containing two possible
+-- \"strategies\" to dispatch block; calling `dispatchBlockSync` ensures the
+-- function won't return up until the `Protocol` hasn't been fully stepped
+-- (i.e. block inserted in the store, chain selection fully performed etc.)
+-- `dispatchBlockAsync` as the name implies, doesn't block and return the
```
cloudhead:
```suggestion
-- `dispatchBlockAsync` as the name implies, doesn't block and returns the
```

:book: :pen:

src/Oscoin/Protocol.hs
```diff
@@ -0,0 +1,276 @@
...
+    -- ^ TODO(Adn): Return an error?
+    , dispatchBlockAsync :: Block c tx (Sealed c s) -> m ()
+    }
+
+{------------------------------------------------------------------------------
+  Using the 'Protocol'
+------------------------------------------------------------------------------}
+
+-- | Given the data necessary to initialise and build the 'Protocol', runs
+-- the input action and return the final result.
```
cloudhead:
```suggestion
-- the input action and returns the final result.
```

src/Oscoin/Protocol.hs
```diff
@@ -0,0 +1,276 @@
...
+    -> (Protocol c tx s m -> m a)
+    -> m a
+withProtocol o validateFull scoreBlock bs config action =
+    let protocol = Protocol bs o validateFull scoreBlock config
+    in action protocol
+
+-- | Initialises and runs the 'Protocol', by spinning an asynchronous worker
+-- which executes the 'fetchNextBlock' function to pull new blocks from outside
+-- and steps the protocol.
+runOscoinProtocol
```
cloudhead: Any reason this function isn't called `runProtocol`, to match `withProtocol`?

src/Oscoin/Protocol.hs
```diff
@@ -0,0 +1,276 @@
...
+       , Buildable (Hash c)
+       , Monad m
+       )
+    => Protocol c tx s m
+    -> Block c tx (Sealed c s)
+    -> m (Protocol c tx s m, [NotableEvent])
+stepProtocol mgr incomingBlock = do
+    -- Step 2. Try to store the block; if this is recognised as to be an
+    -- orphan, we store it immediately (/PRECONDITION/: the block has been
+    -- pre-validate at the gossip level). If this block is scheduled for
```
cloudhead: I'd move the precondition to the haddock, or else it won't be as useful I think.

cloudhead: It's also not actually a pre-condition if I understand correctly, ie. the function will not fail if the block wasn't basic-validated, or am I wrong? If so, I'd turn it around and add `Nb. Does not perform basic validation`

src/Oscoin/Protocol.hs
```diff
@@ -0,0 +1,276 @@
...
+    -- inclusion to extend the tip of the chain, then its full validity is
+    -- enforced.
+    extendsTip <- do
+        currentTip <- BlockStore.getTip bsPublicAPI
+        pure $ blockHash currentTip == parentHash incomingBlock
+
+    (mgr', evts) <-
+       if extendsTip
+          then do
+              -- Get the last (cached) 'mutableBlocksConstant' blocks and performs
```
cloudhead: I think you mean `mutableBlocksSize` (or `Depthz if you follow my suggestion)

src/Oscoin/Storage/Block/Abstract.hs
```diff
@@ -7,37 +7,59 @@
...
 
--- | An handle over a block storage backend.
-data BlockStore c tx s m = BlockStore
-    { scoreBlock      :: Block c tx (Sealed c s) -> Score
-    -- ^ When given a 'Block', return its 'Score'.
-    , insertBlock     :: Block c tx (Sealed c s) -> m ()
+data GetBlocksQuery c =
+      ByDepth Depth
+    -- ^ Get a collection of blocks from the input 'Depth' up to the tip.
+    | ByRootHash (BlockHash c)
```
cloudhead: I think the usage of `Root` here may be confusing (since the root of the block tree would actually be the genesis) - maybe `ByParentHash` or `ByBlockHash`, depending on which one you mean?

src/Oscoin/Time/Chrono.hs
```diff
@@ -0,0 +1,66 @@
...
+-- one, where the API contract stats that elements are ordered from the newest
+-- to the oldest.
+newtype NewestFirst f a = NF { toNewestFirst :: f a }
+  deriving (Show, Eq, Functor, Foldable, Semigroup, Monoid)
+
+-- | A wrapper over a 'Foldable' functor 'f', typically a list or a non-empty
+-- one, where the API contract stats that elements are ordered from the oldest
+-- to the newest.
+newtype OldestFirst f a = OF { toOldestFirst :: f a }
+  deriving (Show, Eq, Functor, Foldable, Semigroup, Monoid)
```
cloudhead: I think the correct terms are simply `Chronological` (OldestFirst) and `ReverseChronological` (NewestFirst)

src/Oscoin/Time/Chrono.hs
```diff
@@ -0,0 +1,66 @@
...
+-- [3,2,1]
+--
+class ChronogicallyOrdered a where
+    type ReverseOf a :: *
+    reverse :: a -> ReverseOf a
+
+-- | A wrapper over a 'Foldable' functor 'f', typically a list or a non-empty
+-- one, where the API contract stats that elements are ordered from the newest
+-- to the oldest.
+newtype NewestFirst f a = NF { toNewestFirst :: f a }
```
cloudhead: I appreciate that `NF` is faster to type, but wouldn't calling the constructor the same as the type better, both for readability and consistency?

src/Oscoin/Storage/Block/Abstract.hs
cloudhead: More general comment - is there a reason why we don't just have two functions: `getBlocksByDepth` and `getBlocksByParentHash`? It's a lot simpler if there isn't a strong reason for the extra type.

src/Oscoin/Node.hs
adinapoli-mndc: @cloudhead was merely let the suspense to build up. You know, Italians are notoriously over-dramatic 😁 :trollface: .

adinapoli-mndc: ps. It's not even Italian, that would be "nota" 😂

src/Oscoin/Protocol.hs
adinapoli-mndc: 😭 🏫

src/Oscoin/Protocol.hs
adinapoli-mndc: I just wanted to convey the idea this is like "the thing" we use to run _the_ Oscoin protocol, but for consistency, I'll amend it 😉

src/Oscoin/Storage/Block/Abstract.hs
adinapoli-mndc: No strong reason if not API economy -- I thought this could belong to a set of related query on the chain, but perhaps you are right, this is premature optimisation and actually clutters a bit the API.

src/Oscoin/Time/Chrono.hs
adinapoli-mndc: I am on the fence here: from one side I agree that the original naming is not the most elegant, but at the same time is brutally efficient: it states exactly how the data is ordered, and "oldest" vs "newest" helps building the intuition immediately. 

Conversely, `Chronological` and `ReverseChronological` don't really help in this sense (but I might be just me and the tunnel vision I have developed on these names).

Do you think the sweet spot could be to name the types the way you want but to keep the getters the same (i.e. retain `toOldestFirst` and `toNewestFirst`)?

If you have very strong sentiments here, I can surely change everything, that's not a problem 😉

adinapoli-mndc:
@cloudhead if I didn't make any fatigue-dictated mistakes, the last commit should address all your concerns modulo the chrono bikeshedding. For that, I am happy to either abide to your suggestion or to fire that as a subsequent PR, in case that help expediting the merge of this one 😁

src/Oscoin/Node.hs
cloudhead: Hahahahaha

src/Oscoin/Storage/Block/Abstract.hs
cloudhead: Yeah I think it introduces "yet another type" I need to understand to use the function.

src/Oscoin/Time/Chrono.hs
cloudhead: I think it's fine as it is : ) - the part that feels icky to me is that old/new sounds like we're talking about something which ages :derelict_house: . For blocks, recency or height might have been more appropriate.  I can live with it :wink:

src/Oscoin/Storage/Block/SQLite.hs
```diff
@@ -37,42 +31,42 @@ import           Codec.Serialise (Serialise)
...
-                , Abstract.lookupTx        = lookupTx internalHandle
-                , Abstract.getOrphans      = getOrphans internalHandle
-                , Abstract.getBlocks       = getBlocks internalHandle
-                , Abstract.getTip          = getTip internalHandle
+            ( (Abstract.BlockStoreReader {
+                  Abstract.getGenesisBlock       = getGenesisBlock (hConn internalHandle)
+                , Abstract.lookupBlock           = lookupBlock internalHandle
+                , Abstract.lookupTx              = lookupTx internalHandle
+                , Abstract.getBlocksByDepth      = getBlocks internalHandle
+                , Abstract.getBlocksByParentHash = getChainSuffix (hConn internalHandle)
```
cloudhead: Just to make sure: this won't return the block corresponding to the hash, correct? Ie. if you do `getBlocksByParentHash a`, then with the chain `a <- b <- c`, it should only return `b <- c`.

src/Oscoin/Storage/Block/Pure.hs
```diff
@@ -100,6 +102,16 @@ getBlocks (fromIntegral -> d) Handle{..} =
 -- Nb. we guarantee that there is at least one chain in the store by exposing
 -- only the 'genesisHandle' smart constructor.
 
+-- | Gets a suffix of the 'BlockChain' originating from the input 'BlockHash'.
```
cloudhead:
```suggestion
-- | Gets a suffix of the 'Blockchain' originating from the input 'BlockHash'.
```
Not sure haddock is case-sensitive, but...

src/Oscoin/Protocol.hs
```diff
@@ -0,0 +1,275 @@
...
+            -- Compare the orphan best chain with the chain suffix currently adopted
+            chainSuffix <- BlockStore.getBlocksByParentHash bsPublicAPI parentHash
+
+            -- Switch-to-better-chain condition: either the chain suffix is
+            -- empty, which means we are extending directly the tip, or if we
+            -- found a better candidate.
+            let scoreFn       = protoScoreBlock mgr
+                newChainFound =
+                    null chainSuffix
+                 || O.fromChainSuffix scoreFn (NonEmpty.fromList . toNewestFirst $ chainSuffix) < bestChain
```
cloudhead: This is suspicious to me, why did we find a new chain if it is lesser than the best chain?

src/Oscoin/Protocol.hs
```diff
@@ -0,0 +1,275 @@
...
+    -> m (Protocol c tx s m)
+selectBestChain mgr = do
+    -- NOTE(adn) Grab the \"mutable\" part of the chain and
+    -- compare each root hash to see if there is any fork originating from that.
+    let depth = fromIntegral $ mutableChainDepth (protoConfig mgr)
+    mutableBlockHashes <-
+        map blockHash . Chrono.reverse <$> BlockStore.getBlocksByDepth bsPublicAPI depth
+
+    case O.selectBestChain (toOldestFirst mutableBlockHashes) orphanage of
+        Just (parentHash, bestChain) | isValid bestChain -> do
```
cloudhead: Do we have to check `isValid` here? Isn't that really expensive? I'd assume that chains returned from the orphanage are always valid.

src/Oscoin/Storage/Block/Cache.hs
```diff
@@ -0,0 +1,66 @@
+{-- | A cache for 'Blocks', implemented as a ring buffer.
+ -- It caches the \"mutable\" part of the chain, which are usually the
+ -- last @k@ blocks, where @k@ is a protocol constant.
+ --}
```
cloudhead: I love it when you can use a ring buffer.

src/Oscoin/Protocol.hs
```diff
@@ -0,0 +1,275 @@
...
+                      in pure (mgr, [evt])
+                  Right () -> do
+                      -- Step 3: store the fully-validated block.
+                      BlockStore.insertBlock bsPrivateAPI incomingBlock
+                      pure (mgr, mempty)
+
+          -- NOTE(adn) This is a side effect of the fact that storing
+          -- the genesis as an orphan would insert it into the orphanage with
+          -- its parent hash, which means that we won't be able to retrieve
+          -- a potentiol candidate branching off from genesis in our 'selectBestChain'
```
cloudhead:
```suggestion
          -- a potential candidate branching off from genesis in our 'selectBestChain'
```

src/Oscoin/Protocol.hs
```diff
@@ -0,0 +1,275 @@
...
+          -- function (cfr. QuickCheck seed 554786). To avoid that, we avoid
+          -- inserting the genesis block into the orphanage in the first place;
+          -- After all, the genesis block should never be considered at orphan.
```
cloudhead:
```suggestion
          -- After all, the genesis block should never be considered an orphan.
```

src/Oscoin/Protocol.hs
cloudhead: There's also the issue that if the best chain is invalid, then perhaps there was another chain that scored lesser, but is actually valid.

src/Oscoin/Storage/Block/Cache.hs
```diff
@@ -0,0 +1,66 @@
...
+    -- elements.
+    -> IO (NewestFirst [] (Block c tx s))
+cached (BlockCache bc) requestedElements backFill =
+    modifyMVar bc $ \cseq -> do
+        let currentSize = CSeq.getSize cseq
+        if currentSize < requestedElements
+          then do
+              newElements <- backFill (requestedElements - currentSize)
+              let cseq' = foldr (CSeq.<|) cseq newElements
+              pure (cseq', NF $ take requestedElements $ F.toList cseq')
```
cloudhead: The `NF` thing hasn't been changed - I don't know if you intended to or not?

src/Oscoin/Storage/Block/SQLite.hs
adinapoli-mndc: I've originally extracted this function from the (already existing) `getChainSuffix` of the SQLite internal module. From what I can see:

```
getChainSuffix conn hsh = do
    result <- lookupBlockTimestamp conn hsh
    case result of
        Nothing -> pure mempty
        Just t -> do
            rows :: [Only (BlockHash c) :. BlockHeader c s] <- Sql.query conn
                [sql|  SELECT hash, parenthash, datahash, statehash, timestamp, difficulty, seal
                         FROM blocks WHERE timestamp > ?
                     ORDER BY timestamp DESC
                        |] (Only t)
            map NF <$> for rows $ \(Only h :. bh) ->
                mkBlock bh <$> getBlockTxs conn h
```

Indeed the behaviour is what you hinted at, as the query is `WHERE timestamp > ?`. 

However, with a fresh mind, I do wonder if this `ByParentHash` suffix ends up being more confusing than anything -- should we relabel this function `getBlocksFromParentHash`? If we want to make this more Java-esque, we can call it `getBlocksFromParentHashExclusive` 😂

src/Oscoin/Protocol.hs
adinapoli-mndc: oh, because the `Ord` instance compares by block score (and height!) so this is essentially saying that if the chain suffix _of the main chain_ is "less" than the best chain, we switch.

src/Oscoin/Protocol.hs
adinapoli-mndc: > Do we have to check isValid here? Isn't that really expensive?

Uhm, but wasn't one of the objective of this whole refactoring split validation in two halves where the heavy part (i.e. this one) would be done only at this stage?

>  I'd assume that chains returned from the orphanage are always valid.

When we insert a block in the orphanage we check only "phase-1" validation, so it's actually possible that the overall best chain is not valid when we try to hook in into the main chain, no?

> There's also the issue that if the best chain is invalid, then perhaps there was another chain that scored lesser, but is actually valid.

Ah, fair point. Yes, this was something I didn't have to worry about in the first iteration of fork selection, but this is now missing from this patch.

src/Oscoin/Protocol.hs
adinapoli-mndc: I think it's time for me to start coding with `:set spell` always on 😂

src/Oscoin/Storage/Block/Cache.hs
adinapoli-mndc: Left it out on purpose whilst we decided on the fate of the name of the data structures on `Oscoin.Time.Chrono` to not embark into another "search & replace", but looks like now is time to do that 😉

src/Oscoin/Protocol.hs
adinapoli-mndc: I am also thinking about something else -- in particular, we are calling `validateBlockchain` without taking a look at the prefix this new chain will insert onto. In other terms, say you have found a fork, and you need to validate such fork. shouldn't we also make sure that the fork _plus_ a bunch of blocks in the main chain also validate? Otherwise it might be possible that the fork is _locally_ valid, but not "globally", if you see what I mean?

Did I lose the plot? 🤔

src/Oscoin/Storage/Block/Cache.hs
adinapoli-mndc: a42f62e

src/Oscoin/Protocol.hs
adinapoli-mndc: a42f62e

src/Oscoin/Protocol.hs
adinapoli-mndc: a42f62e

src/Oscoin/Storage/Block/Pure.hs
adinapoli-mndc: a42f62e

src/Oscoin/Protocol.hs
adinapoli-mndc: Ah, I understand your confusion now -- no no, the naming perhaps doesn't help -- `bestChain` is actually the chain _we extracted from the orphanage_, not the current, adopted one 😉

adinapoli-mndc:
> The code still does scoring and chain selection on insert. (This might even mess with invariants we expect from the dumb block store now.)

@geigerzaehler That's true -- however, we _do need_ such thing as a reference implementation, otherwise we wouldn't be able to run the `Equivalence` tests. I propose we let the `Protocol` stabilise and once we gained enough confidence in its behaviour being correct, we perform the removal 😉

src/Oscoin/Storage/Block/Abstract.hs
```diff
@@ -7,37 +7,53 @@
...
 
--- | An handle over a block storage backend.
-data BlockStore c tx s m = BlockStore
-    { scoreBlock      :: Block c tx (Sealed c s) -> Score
-    -- ^ When given a 'Block', return its 'Score'.
-    , insertBlock     :: Block c tx (Sealed c s) -> m ()
+-- | The internals for this 'BlockStore', i.e. the private API which shouldn't
+-- be visible to \"read-only consumers\".
+data BlockStoreWriter c tx s m = BlockStoreWriter
+    { insertBlock     :: Block c tx (Sealed c s) -> m ()
```
geigerzaehler: It would be great if we could document that this assumes that the parent of `block` the same is as the result of `getTip`.

src/Oscoin/Storage/Block/Abstract.hs
```diff
@@ -7,37 +7,53 @@
...
 
 import           Oscoin.Prelude
 
-import           Oscoin.Consensus.Types (Validate)
 import           Oscoin.Crypto.Blockchain (TxLookup)
 import           Oscoin.Crypto.Blockchain.Block
 import           Oscoin.Crypto.Hash (Hashed)
+import           Oscoin.Time.Chrono (NewestFirst, OldestFirst)
+
+-- The full 'BlockStore' API is the composition of its public and private parts.
```
geigerzaehler: It would be great if we could document the model and invariants behind a block store in details. I’d suggest the following
~~~plain
Morally, 'BlockStore' is a linked list starting with the oldest block (the genesis block)
and ending at the tip. New blocks can be appended at the end with 'insertBlock' and
the last @n@ blocks can be replaced with 'switchToFork'.
~~~

src/Oscoin/Storage/Block/Cache.hs
```diff
@@ -0,0 +1,66 @@
...
+
+cached
+    :: BlockCache c tx s
+    -> Int
+    -- ^ The number of requested elements
+    -> (Int -> IO (NewestFirst [] (Block c tx s)))
+    -- ^ A function to \"backfill\" the cache by the number of missing
+    -- elements.
+    -> IO (NewestFirst [] (Block c tx s))
+cached (BlockCache bc) requestedElements backFill =
```
geigerzaehler: Could you add tests for this function?

test/Oscoin/Test/Storage/Block/Cache.hs
```diff
@@ -0,0 +1,50 @@
...
+
+worksRingBufferProp :: forall c. Dict (IsCrypto c) -> Property
+worksRingBufferProp Dict = monadicIO $ do
+    cache <- liftIO (newBlockCache 10)
+    forAllM (arbitrary `suchThat` ((> 10) . length)) $ \(blks :: [Block c Text Text]) -> do
+        storedBlocks <- liftIO $ do
+            forM_ (reverse blks) (consBlock cache)
+            viewBlocks cache
+        assert $ head storedBlocks == head blks
+        assert $ storedBlocks /= blks
```
geigerzaehler: We could drop the two assertions and replace them with
```suggestion
        assert $ storedBlocks == take 10 blks
```

src/Oscoin/Storage/Block/Abstract.hs
```diff
@@ -7,37 +7,53 @@
...
-    -- ^ When given a 'Block', return its 'Score'.
-    , insertBlock     :: Block c tx (Sealed c s) -> m ()
+-- | The internals for this 'BlockStore', i.e. the private API which shouldn't
+-- be visible to \"read-only consumers\".
+data BlockStoreWriter c tx s m = BlockStoreWriter
+    { insertBlock     :: Block c tx (Sealed c s) -> m ()
     -- ^ Inserts a 'Block' into the store.
-    , getGenesisBlock :: m (Block c tx (Sealed c s))
+    , switchToFork    :: Depth -> OldestFirst NonEmpty (Block c tx (Sealed c s)) -> m ()
+    -- ^ Virtually isomorphic to 'insertBlocks . rollBackBlocks', but wrapped up
```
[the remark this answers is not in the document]
adinapoli-mndc: Good catch!

src/Oscoin/Storage/Block/Abstract.hs
adinapoli-mndc: Uhm, I am a bit on the fence with this one. The `BlockStore` per se is an opaque interface to "some persistence", and it feels strange to essentially describe it in the same terms you would do for a `Blockchain`. Furthermore, is a bit of a lie that "new blocks can be appended at the end", isn't it? In reality, we always store new things "in front", so the description here has the risk of confusing the reader even more. But even the last description is misleading, as (in reality) how we store things is purely an implementation detail of the concrete store we pick.

What about we say something along the lines of: 

> Morally, the 'BlockStore' offers an opaque interface to some kind of persistent storage for blocks, with a set of operations to insert, remove and retrieve these blocks.

test/Oscoin/Test/Storage/Block/Cache.hs
adinapoli-mndc: 3d20d7a

src/Oscoin/Storage/Block/Cache.hs
adinapoli-mndc: 3d20d7a

src/Oscoin/Storage/Block/Abstract.hs
adinapoli-mndc: 3d20d7a

src/Oscoin/Storage/Block/Abstract.hs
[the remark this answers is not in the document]
adinapoli-mndc: 3d20d7a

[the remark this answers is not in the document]
adinapoli-mndc: ps. Ended up also shuffling around the comments inside `Telemetry.Events` as I did notice they were breaking `stack haddock`, because apparently GADTs cannot have inline comments in the `-- ^ ` style.

adinapoli-mndc:
Dang, CI fails again. The good news is that I have already fixed that faulty test in #415, so I suggest we:

- [ ] Do another round of reviews;
- [ ] Get approval for this PR (missing Alexis & Thomas);
- [ ] Merge all the PRs built on top of this and pre-approved here
- [ ] Merge the Cthulhu

src/Oscoin/Storage/Block/Abstract.hs
geigerzaehler: I’d be ok with that. I just wanted to make the ordering clearer. Especially that `getTip` gets the latest block from `insertBlock` and `switchToFork` replaces the most recently inserted elements. But that is probably covered by the field documentation already.

src/Oscoin/Storage/Block/SQLite.hs
cloudhead: Ok - I think the name is fine, but I see what you mean. I think `getBlocksWithParentHash` would be definitely wrong, but `by` just means we're using a parent hash to get the blocks, but not what the relationship of the blocks and the hash is.

src/Oscoin/Protocol.hs
cloudhead: Oh! yes it's the naming. Maybe even reversing the condition would help:
`bestChain > O.fromChainSuffix ...`, or renaming `bestChain` to `bestFork`

src/Oscoin/Protocol.hs
cloudhead: I was going to suggest that :joy:

src/Oscoin/Protocol.hs
cloudhead: Ok it seems like this is a bigger conversation - can we create an issue and link to this comment? I don't want to hold up the PR.

adinapoli-mndc: Yes, sir 👍

adinapoli-mndc: https://github.com/oscoin/oscoin/issues/420

geigerzaehler:
:clap:

PR #410, opened 2019-03-13
Remove 'Oscoin.Node.Tree'

geigerzaehler:
Remove the `Oscoin.Node.Tree` module. Only the type aliases `Key` and `Path` were used from this module. We inline these for now.


geigerzaehler:
> By the time you will be finished, no code will be left in the repo :trollface:

https://github.com/kelseyhightower/nocode

cloudhead:
We will need something like this back quite soon though, as the node will have to hold on to a mutable latest state.

PR #412, opened 2019-03-14
Regen genesis after LTS bump.

adinapoli-mndc:
This PR fixes the miner, which silently broke after commit `1a45548bf8fb64f3edd2a7a0d6cdc0c438524f32` (the one which bumped the LTS). Specifically, the following pattern match _always_ yielded `Nothing`:

```haskell
mineBlock bs Consensus{cMiner} eval time = do
    txs <- (map . map) snd Mempool.getTxs
    parent <- BlockStore.getTip bs
    maybeState <- StateStore.lookupState $ blockStateHash $ blockHeader parent
    case maybeState of
        Just st -> do
            let (blockCandidate, st', receipts) = buildBlock eval time st txs (blockHash parent)
            maybeSealedBlock <- cMiner (BlockStore.getBlocks bs) blockCandidate
            for maybeSealedBlock $ \blk -> do
                Mempool.delTxs (blockData blk)
                BlockStore.insertBlock bs blk
                StateStore.storeState st'
                for_ receipts addReceipt
                pure blk
        Nothing ->
            pure Nothing
```

I have a somewhat plausible explanation: in order for the lookup to succeed, we need the `blockHash` of the current tip to be found in the state store. For the genesis block, this is computed from the initial radicle environment. However, since the LTS bump, the Radicle `Env` changed, as now it includes also documentation annotations (i.e. `Docd`). This caused a silent hash mismatch, and now the lookup would always yield nothing.

For now, as an interim solution, I have regenerated the `genesis.yaml` and added a test which test that the miner needs to emit `block mined` events. That would catch this sneaky misbehaviour in the future. Having said that, methinks we should also think hard about the repercussions of such design. I appreciate we might be replacing Radicle soon, but we should somehow make sure that our genesis block hash cannot change and be influenced by external factors, otherwise this would be bad once we hit production 😨


cloudhead:
Yeah - we need to have an extremely predictable way of hashing state, which will be a lot easier with that data being inside of oscoin, vs in radicle.

@kim I think the value changed because the hash is of the "empty pure env" of radicle, which does contain some primops and stuff.

cloudhead:
> As a follow-up, I think we should consider mining the genesis at node startup, and comparing the result against a configured or hardcoded value, exiting if it doesn’t match.

Hmm, that's tricky because most of the parameters to the genesis block are "a prioris" that the node can't know anything about. But maybe the node can do some sort of validation on it.

kim:
> the parameters to the genesis block are "a prioris" that the node can't
know anything about

We could either start with an empty genesis, or supply the inputs at
startup along with the expected hash(es).

PR #415, opened 2019-03-19
Better internal representation for Data.Sequence.Circular

adinapoli-mndc:
Builds on top of #402, which is recommended we merge first.

This is mostly for @kim as I know he is a speed/performance junkie :trollface: .

This PR switch the internal representation for `Data.Sequence.Circular` to a very efficiently (and deadly dangerous) mutable vector implementation. I have also drastically simplified that module to retain only things we need.


src/Data/Sequence/Circular.hs
```diff
@@ -1,16 +1,11 @@
 {--
- This module models a Circular Buffer using Data.Sequence's `Seq` as foundation.
+ This module models a Circular Buffer using a vectore a as foundation.
```
kim: > a vectore

Bene così! :)

adinapoli-mndc: Eccellente! 😁

cloudhead: :joy:

src/Data/Sequence/Circular.hs
```diff
@@ -29,114 +26,48 @@ import qualified Data.Sequence as Seq
...
-instance Foldable CircularSeq where
-  foldMap f (CSeq _ _ x) = foldMap f x
+instance Show a => Show (CircularSeq a) where
+   show = show . toList
 
 {------------------------------------------------------------------------------
   Main API
 ------------------------------------------------------------------------------}
 
+toList :: CircularSeq a -> [a]
```
geigerzaehler: Could you document the ordering of the elements in the list? I.e. the most recently pushed element is at the end while the oldest is at the beginning

src/Data/Sequence/Circular.hs
```diff
@@ -1,16 +1,11 @@
 {--
- This module models a Circular Buffer using Data.Sequence's `Seq` as foundation.
+ This module models a circular buffer using a vector as a foundation.
```
geigerzaehler: It would be great to have tests for this module too so that we don’t need rely on the block cache tests.

adinapoli-mndc: I can surely do that (I think that's a good idea anyway).

src/Data/Sequence/Circular.hs
adinapoli-mndc: ps. note that the ordering is the opposite of what you are describing -- due to the fact we expose only a "cons" operation (i.e. <|), elements are added in front of the sequence, i.e:

```haskell
> import qualified Data.Sequence.Circular as CSeq
> let c = CSeq.empty 3
> 1 CSeq.<| c
[1]
> 2 CSeq.<| it
[2,1]
```

We could expose a `|>` function as before, but I have removed it as it's not needed for our use case (i.e. the block cache).

adinapoli-mndc:
@geigerzaehler This should be better now 😉 Let me know what you think! 

Screenshot 2019-03-20 at 16 00 45

PR #416, opened 2019-03-20
Fix broken block mining

adinapoli-mndc:
*This builds on top of #402.*

This commit fixes the broken block mining by acting on the Nakamoto
validation function. The culprit was simple and subtle:

One of the checks for a block to be valid, is that it shouldn't be more
than 2 hours "in the future" with respect to the parent block. However,
when mining the very *first block on top of genesis*, this check is
going to always fail, as genesis has a timestamp which is baked inside
`data/genesis.yaml` and it's the time when this genesis file has been
generated.

I have added some band aid to mitigate the immediate issue, but I suspect the morally correct thing to do would be to stop baking the timestamp inside the file, but rather have a CLI-passed parameter like "system start time" that can be used to assemble the genesis block in memory, so that we don't need that special case (this resonates a bit with what @kim said a while ago about Genesis, I believe).


cloudhead:
Great catch. I don't think we should merge this though, since it breaks the protocol rules and we'll have to remember to revert it. I think until we have a permanent solution to this, we should just locally regenerate the genesis when testing.

adinapoli-mndc:
I agree -- let me close this one and open a ticket to do this "properly" i.e. for now, by lack of better designs, we do allow the node to be start with a `fullValidation` function different depending on the `Environment` the node is starting with, so that we have more flexibility.

PR #422, opened 2019-03-21
Fixed #421

### LINKED ISSUE #421: Allow different validation function depending on the `Environment`
See: https://github.com/oscoin/oscoin/pull/416#issuecomment-475180591

adinapoli-mndc:
This patch tries to fix #421 by renaming `validateFull` to be a sort of facade to a zoo of validation functions, the specific one selected by the input `Environment`.

This allows us to now fix the broken mining without compromising the full, production-like validation.

Ps. Not very happy with the duplication of `validateFullDev` and `validateFullProd`, but factoring out their code looks like would make this PR grow bigger, but I can certainly do it if people think that's worth it.


adinapoli-mndc:
@cloudhead right, I am not terribly happy with the design either but I cannot see the wood for the trees at the moment.

> or to have a separate consensus that uses all the parts of Nakamoto, but looser validation

Good idea, I will cook up something following your suggestion 😉

adinapoli-mndc:
@cloudhead Ok, attempt n2, tell me if you like it better this time!

I have also seized the chance to remove the duplication from the two validation functions by splitting out the "standalone validation functions". At the moment this results in a more verbose module, but I guess we gain is composability and testability, as now we can test each separate validator in isolation.

Tell me what you think!

cloudhead:
This is better! I'd split out the `Lenient` into its own module though, eg `Nakamoto.Lenient` - I think the difference between the two should be made super obvious.

adinapoli-mndc:
>  I think the difference between the two should be made super obvious. The lenient validation functions can then also go in there, and the module can depend on the base Nakamoto module.

I took a slightly different approach and ended up moving out those functions into a separate `Oscoin.Consensus.Validation` module. My line of reasoning here is that those are not _lenient_ validation functions per se, they are "the real deal", so it would feel awkward to have them in the `Lenient` module.  Furthermore, now these small lego blocks are completely independent from any specific consensus algorithm, so it might be handy in the future to have them in a standalone module.

@cloudhead , do you feel this is an acceptable design?

adinapoli-mndc:
CI yielded not one, but two (unrelated) failures I have logged as part of #428 and #429 😨 .

PR #424, opened 2019-03-21
Tame the bracket spaghetti in node/Main

kim:


exe/node/Main.hs
```diff
@@ -23,16 +23,16 @@ import qualified Oscoin.P2P as P2P
...
-import qualified Oscoin.Storage.Block.STM as BlockStore.Concrete.STM
+import qualified Oscoin.Storage.Block.STM as BlockStore.STM
 import qualified Oscoin.Storage.State as StateStore
 import qualified Oscoin.Telemetry as Telemetry
 import           Oscoin.Telemetry.Logging (withStdLogger)
 import qualified Oscoin.Telemetry.Logging as Log
 import           Oscoin.Telemetry.Metrics
 
 import qualified Control.Concurrent.Async as Async
+import           Control.Monad.Managed
```
adinapoli-mndc: Apparently there is a `Control.Monad.Managed.Safe` module which is even more stringent -- do we have any chance of using that?

exe/node/Main.hs
```diff
@@ -115,40 +115,55 @@ main = do
...
+                       P2P.NodeAddr { P2P.nodeId   = pure nid
+                                    , P2P.nodeHost = host
+                                    , P2P.nodePort = gossipPort
+                                    }
+                       seeds
+                       (storage node)
+                       (Handshake.simpleHandshake keys)
+
+        liftIO . Log.withExceptionLogged lgr $ do
+            forkEkgServer metricsStore ekgHost ekgPort
```
adinapoli-mndc: Just to make sure my telemetry code relied in some crazy ordering of event -- are you still able to run the node and open the EKG dashboard? (the `--help` should tell you which port EKG listens by default).

kim: Yeah this works. It would be preferable to be able to manage the ekg thread in `main`, and make sure it starts up _only_ after the other services have started successfully. But I'll leave that for another time :)

exe/node/Main.hs
kim: It's the same thing, except that the `Safe` module doesn't export `with`, which let's you leak the "managed" resource

adinapoli-mndc: Gotcha 😉

exe/node/Main.hs
adinapoli-mndc: Right, I was dwelling upon that as well myself, and wrote a half-baked comment which I scrapped which essentially was wondering if we should have called that function `runEkgServer` and then make it run in the `runConcurrently` block with the rest, for consistency.

Problem being we piggyback on EKG's `forkServerWith`, and _I think_ `ekg-core` itself doesn't expose the relevant function to start up the ekg server synchronously.

exe/node/Main.hs
```diff
@@ -115,40 +115,66 @@ main = do
...
+                           (storage node)
+                           (Handshake.simpleHandshake keys)
+
+            liftIO $ do
+                forkEkgServer metricsStore ekgHost ekgPort
+                Async.runConcurrently $
+                         Async.Concurrently (HTTP.run (fromIntegral apiPort) node)
+                      <> Async.Concurrently (miner node gossip)
+
+    either (const $ exitFailure) (const $ exitSuccess) res
```
cloudhead:
```suggestion
    either (const exitFailure) (const exitSuccess) res
```
Tame the :money_with_wings:

PR #425, opened 2019-03-21
Support building the project with cabal-new

adinapoli-mndc:
This PR ships a `cabal.project` file suitable to build the project using `cabal` and `cabal new-build`.

It leverages [stack2cabal](https://github.com/edsko/stack2cabal) to generate this starting `cabal.project`. No effort is made to keep this file up-to-date by CI, but this is rather left to the common sense of the developers interested (I can surely nurture it).

Generating this initial file was a bit cumbersome due to the fact we are using a custom snapshot and `stack2cabal` doesn't support that. I ended up temporarily "unrolling" the `snapshot.yaml` file inside `stack.yaml` and kickoff things from there. Furthermore, `stack2cabal` comes with a clever mechanism which predates the introduction of the [source-repository-package](https://cabal.readthedocs.io/en/latest/nix-local-build.html#specifying-packages-from-remote-version-control-locations) directive: basically it checks out all the remote deps into a folder and treat them as "local". Therefore I had to strip these off and replace them with the right directives (I might end up firing a patch to `stack2cabal` some day).

Last but not least , I had to install `hpack` due to an unfortunate event: due to the fact more and more maintainers are switching to use `stack` only and thus `hpack`, the habit of _not_ checking out the hpack-generated `.cabal` file is spreading. As a result, `cabal new-build` won't be able to checkout and install the remote dependency:

```
monadic/oscoin (adinapoli/cabal-new●) > cabal new-build                                                               
Cloning into '/Users/adinapoli/work/clients/monadic/oscoin/dist-newstyle/src/radicle-7d30549013eba4a4'...
warning: redirecting to https://github.com/oscoin/radicle/
remote: Enumerating objects: 126, done.
remote: Counting objects: 100% (126/126), done.
remote: Compressing objects: 100% (77/77), done.
remote: Total 5893 (delta 76), reused 73 (delta 48), pack-reused 5767
Receiving objects: 100% (5893/5893), 1.47 MiB | 1.71 MiB/s, done.
Resolving deltas: 100% (3602/3602), done.
HEAD is now at eaf70a9 Merge pull request #514 from oscoin/b/base-pattern-matching
NoCabalFileFound
```

To circumvent this, it's possible to `cd` into the download repo and invoke `hpack package.yaml` there, and that is enough to please `cabal`.

Overall, it's a bit clunky at the moment to get started with `cabal new-*`, but it's do-able.

This requires `hpack`, `ghc`, `ghc-pkg`, `hsc2hs` and `markdown-unlit` to all be in your `PATH`.

`cabal new-test` is a bit clunky to orchestrate, due to the fact our integration tests on executables requires `oscoin` and `oscoin-cli` to be run. Now, with `stack test` this will "automagically work" as `stack` is probably silently adding the stuff from `.stack-work/dist` in the PATH, but this wasn't the case here, so I ended up running something like:

```
PATH=$PATH:dist-newstyle/build/x86_64-osx/ghc-8.6.3/oscoin-0.1.0.0/x/oscoin-cli/build/oscoin-cli/:dist-newstyle/build/x86_64-osx/ghc-8.6.3/oscoin-0.1.0.0/x/oscoin/build/oscoin/ cabal new-test
``` 

To get everything working.

It's also possible to use `cabal new-repl` during development, if you are using `ghcid`. This would be something like (i.e. in neovim):

```
:Ghcid -c "cabal new-repl"
```


cabal.project
```diff
@@ -0,0 +1,2389 @@
...
+  type: git
+  location: https://github.com/oscoin/ipfs
+  tag: 30a67159d5b58cf63be903920d6349f0be15cd46
+  subdir: multihash-serialise
+
+package sandi
+  flags: -with-conduit 
+
+constraints:
+-- lts-13.10 (except for extra dependencies)
```
kim: 😱 

Have you tried just placing the `cabal.config` next to this (https://www.stackage.org/lts-13.10/cabal.config)? Would be much easier to maintain...

cabal.project
```diff
@@ -0,0 +1,2389 @@
+packages:
+-- local packages
+    ./oscoin.cabal
+-- remote packages
```
kim: Does this clone the repos multiple times?

adinapoli-mndc:
> Is this only radicle?

I think for _now_ is just radicle, but I guess it's just a matter of time before other projects will do the same (we, for example, indeed do the same, hehe).

cabal.project
adinapoli-mndc: Oh, this bit is done automatically by `stack2cabal` if that is of any relief -- I didn't write any of that!

cabal.project
adinapoli-mndc: you mean whether or not the use of multiple `subdir` would result in multiple clones? That didn't seem to be the case, but I wouldn't know, on top of my head (I was busy tried to patch up stuff with `hpack` that didn't pay attention to this detail!)

cabal.project
cloudhead: Mhm.. :trollface:

adinapoli-mndc: And ps, that's essentially what `stack2cabal` does, it [pulls](https://github.com/edsko/stack2cabal/blob/master/src/Stack2Cabal/Resolver.hs#L18-L28) the `cabal.config` from Stackage and dumps it into the `.project` file 😉 It also explains why I had to (provisionally) butcher our original `stack.yaml` -- the tool expects the input `stack.yaml` to contain a `resolver` entry, which, in case of our `stack.yaml`, points to another, nested yaml file (snapshot.yaml), not the actual name of the LTS 😁   

I am open to alternative suggestions, of course, heh.

kim: Yeah I'm not sure if it's a good idea to rely on `stack2cabal` if we want to make `cabal` a first-class citizen. Dumping the cabal.config into a separate file (which you can even choose to ignore) seems like a cleaner and more maintainable solution.

cabal.project
kim: Yeah it seems to not to. Bit of a nuisance that `subdir` is of type `Maybe FilePath` and not `[FilePath]`.

PR #419, opened 2019-03-20
Grand Unified Peer Discovery

kim:
... which is a bold way of saying "we use DNS for peer discovery".

This turned out to be a bit more pedestrian than anticipated, mainly due to the `network` package being in a transitional phase (but us currently being unable to go straight to version 3.x), and some manual DNS juggling due to mDNS + testing (without modifying `/etc/resolv.conf`).

**tl;dr**:

* There is a new type and CLI parameter, `Network` (`--network`), which determines the logical oscoin network of the node ("mainnet", "testnet", etc.). This is used to minimise accidential joining of incompatible networks, but of course doesn't prevent it. The gossip handshake could ensure this in the future.
* As before, seed nodes can be given statically, if and when their IP addresses, host names, _and_ ports are known. Host names are resolved via DNS A/AAAA records, where _all_ returned addresses are considered as distinct nodes.
* It is possible to advertise a seed node farm (running on non-standard ports) via DNS SRV records. The format is fixed:

    `_gossip._tcp.oscoin.<network>.<search domain>`
    `_api._tcp.oscoin.<network>.<search domain>` 

   where `<network>` is the `Network` above, and `<search domain>` is a domain name owned by the node operator (e.g. "_gossip._tcp.oscoin.testnet.oscoin.io")
* A similar scheme can be enabled for "zeroconf" local area discovery via multicast DNS (`--enable-mdns`).

**Security Considerations**

DNSSEC is not currently supported, so using domain names is inherently insecure. Since we intend to rely on some form of authorisation during gossip handshake, this is of lesser concern.

PR #431, opened 2019-03-25
Make a start on the Address format

adinapoli-mndc:


adinapoli-mndc:
Here are some notes I am jotting down for later.

If we use a conventional scheme `base58 ( <version_prefix>, <base58(hash(pk))>, <checksum>)`, the pretty-printed address ends up being quite long, i.e. 119 chars:

```
4ANWmLdRacbs3muC4CjefXb3qPchq5KMNBJBPeuWU51i5iRVwQEj6oguCM9Ld1xUa49tMArTzW5uQnvRQrzohkJ6aWPWxqmsGLd21vubkspNXCyTG82JaM
```

If we scrap the internal `base58` encoding (do we really need it?) we don't lose anything (on top of my head) and this brings down the size to 94 chars:

```
JGe2DuBcB38oakR3ZGSkbJ277eemNt7cxPfBRVqiJ2UAipamQeDmRXXtYhcZdYeTEtDLhLtxghUrWJjsmHGtDjm3ovYp1
```

So far, comparatively:

| Crypto        | Average length (chars)    |
|---------------|---------------------------|
| Bitcoin       | 34                        |
| Parity (SS58) | 34 (?)                    |
| Ethereum      | 40                        |
| Cardano       | 60 - 93 (variable length) |
| Tezos | 37 (?) |
| Ripple | 35 | 
| Oscoin        | 94 (so far)               |

adinapoli-mndc:
Ok, with a bit of elbow grease we are down to 45 bytes for a typical Oscoin address. It looks like this:

```
JN955wakX2QPb9DDnCunHwFAwW3EYa3AcjbiMScNjyRT
```

xla:
@adinapoli-mndc You probably aware of this one, but I'm not sure it's the one you referring to when listing bitcoin. Leaving it here for consideration: https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki

adinapoli-mndc:
Thanks, @xla ! Yes, with @kim we were pondering wether or not we should take the plunge and implement directly the checksum mechanism described as part of BIP-173 rather than the simpler one, and looks like we should, judging from your comment 😀

xla:
@adinapoli-mndc Don't want to bias but the format proposed seem to have a lot of upsides and positively affects usability from a human perspective. Only wanted to raise awarenss but you seem to be on-top of all the addresses. :>

adinapoli-mndc:
> Don't want to bias but the format proposed seems to have a lot of upsides and positively affects usability from a human perspective. Only wanted to raise awarenss but you seem to be on-top of all the addresses. :>

Please do bias 😉 We have a great chance of implementing this from scratch and I think the more we can piggyback on good ideas taken from other sources, the better our end result will be 😀

xla:
Bias train rolling in :steam_locomotive::train::train::train::train::train:

As it supports **human readable prefixes** and **checksums** it gives a lot of safety to the user when handling addresss as the risk of cnfusion (either as human error or via man-in-the-middle attack) is vastly reduced. Additionally it allows for easy distinction of different account types for that matter and can be thought of/used at a form of namespacing for either later extension or clear separation from the get-go (think account distinction for: project, org and user).

adinapoli-mndc:
Ok, this lacks extensive documentation when it comes to the literate Haskell bit and on how we can migrate an existing `Address` into a new format without breaking backwards compatibility, but I think this is ready for an initial review. There are still some small bits of the API to iron out (which will be apparent when reviewing the code), but I think the format is starting to take shape.

I have been doing some deep thinking over the past few days on the format and how I could accommodate the various design concerns, last but not least @xla 's suggestions and here are my thoughts (which are obviously just the stepping stone for a broader, team-baked discussion): 

The format follows a pretty standard schema which is reminiscent of [SS58](https://wiki.parity.io/External-Address-Format-(SS58)) and `Base58Check` (the original BTC one). Basically, an Oscoin Address is formed as follow:

```
-- * /5/ bytes for the version prefix, so allocated:
-- | 00 | 00 | 00 | 00 | 00 |
--   ^    ^     ^   ^    ^
--   |    |     |   |    |_ reserved (unused)
--   |    |     |   |_ reserved (unused)
--   |    |     |_ address format (ensure backwards compatibility between revisions)
--   |    |_ address type (mainnet, testnet, devnet, multisig, etc)
--   |_  protocol version (bumped on hard forks)
--
-- * /36/ bytes for the CBOR-encoded (short) hash of the 'PublicKey'.
--
-- * /6/ bytes for the checksum, computed from the <prefix + payload>.
```  

In a nutshell, we have a fixed prefix which allows us to modify & evolve the schema in the future, a var-size payload (encoded as CBOR) and a checksum at the end to guard against tampering and transmission errors. Everything is then base58 encoded. The end result, is something like this (48 bytes long):

```
6oybzx6zr5P6rEZKGYXuJHCP83YKPYUD5F9ezsqYHoRcTn7
```

The initial revision of this schema was using as checksum the standard "double SHA256 hash", which, as pointed out by @kim, had some shortcomings. This also echoes @xla 's suggestion to take a look at what BTC has been doing in recent years, and that led to the [BIP-173](https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki#bech32) proposal, i.e. `Bech32` addresses.

After thinking about pro & cons, I have decided the following: I am borrowing the good things from that proposal (namely the improved checksum), but for now I am not taking the plunge on the _full_ format, and here is my line of reasoning:

- The proposal rightly identifies the "double sha" checksum to be potentially slow and not very good.
   Point taken, and this is why I have now implemented checksum verification for Oscoin addresses
   using the BCH codes described in the proposal;

- The proposal makes a point on usability, and how base58 is potentially slow and the mix of upper
   and lower cases makes typing addresses on mobile keyboards clunky.

I do hear where the proposal is coming from. However, I think switching to the full-fat Bech32 for our addresses will have a number of implications to be discussed:

1. The size (in bytes) of each address will go up. The proposal allows up to 90 bytes per address,
    which is quite a lot. This also seems to be contradictory; the proposal talks about ease of typing
    on mobile keyboards, but it also says "[...]The lack of mixed case makes it more efficient to read
    out loud or to put into QR codes. It does come with a 15% length increase, but that does not 
    matter when copy-pasting addresses.[...]" ;

2. The proposal is using _not_ the RFC-based base32 encoding, but rather an ad-hoc
    [implementation](https://gist.github.com/maaku/8996338#file-bip-ecc32-mediawiki) which is
    neither standardised by IETF nor a "de-facto" encoding, by which I mean it might be confusing for
    developers wanting to write applications on top of oscoin understand that they must use not one of
    the widespread base32-encoding library, but one geared towards BTC;

3. The human-readable part is nice, and I agree with @xla that is compelling to use the HRP 
    (_human readable part_) as a sort of namespace per org/project/user, but I also think for our use
    case this is not _too_ important, as we have a mechanism to distribute rewards automatically and
    in general micro-transactions are less frequent than a standard cryptocurrency, i.e. the
    namespacing will end up being more of a "vanity" feature rather than an actually useful one;

4. If the HRP is abused, this can effectively lead to the generation of 
    [vanity addresses](https://allprivatekeys.com/vanity-address.php) which could be easily be
    exploited. I am thinking about draconian cases when somebody would generate addresses like
    `osclinuxfoundationn1.....` and the distracted user might be fooled into making a donation to a
     scam address.

5. Even if we don't allow full "namespacing", we will need to pick the HRP tag and
    [claim it](https://github.com/satoshilabs/slips/blob/master/slip-0173.md) which might be good to
    do regardless we end up using Bech32 or not, but it feels a bit constricting.

6. The proposal introduces a lot of "conditionals", i.e. when uppercase vs lowercase must be used.
     For example, it says:

> The lowercase form is used when determining a character's value for checksum purposes.
>
>Encoders MUST always output an all lowercase Bech32 string. If an uppercase version of the encoding result is desired, (e.g.- for presentation purposes, or QR code use), then an uppercasing procedure can be performed external to the encoding process.
>
> Decoders MUST NOT accept strings where some characters are uppercase and some are lowercase (such strings are referred to as mixed case strings).
>
> For presentation, lowercase is usually preferable, but inside QR codes uppercase SHOULD be used, as those permit the use of alphanumeric mode, which is 45% more compact than the normal byte mode.

I am concerned that this will rapidly make the use of our addresses complicated, both for users and 3rd-party implementors.

I am fairly new to Bech32 myself and I might have misunderstood the BIP completely, so please feel free to debunk me. My 2 cents is that, unless @cloudhead or @xla have a very strong opinion on this
and they think we should absolutely go with Bech32, I think that with the current schema it should be possible to start from here and in the future migrate to a different format (in a compatible way), once we will have a better understanding of how addresses are being used in Oscoin and which are users' pain points with the revision currently in use.

Let me know what you think!

cloudhead:
Great work on this! Here's my 2c:

1. The benefits to sticking to BIP 173 would be that we have a spec to point developers to, vs. mix and matching which means we need our own spec. This means existing decoders for that format will already exist in the wild.
2. I don't think CBOR-encoding the payload is a good idea, or is necessary. It adds a lot of complexity to the address format for no gain - so I'd prefer if we just dumped the binary form of the key straight into the payload, like bitcoin does it. Otherwise, this will require CBOR just for encoding an address.

adinapoli-mndc:
> The benefits to sticking to BIP 173 would be that we have a spec to point developers to, vs. mix and matching which means we need our own spec. This means existing decoders for that format will already exist in the wild.

I hear what you are saying here, but is this really true, though? If developers are going to decode our Bech32-encoded-address, what they will decode will be `(hrp, dat)` , where `dat` will be  _our_ format, not a segwit address, which means they will need to read our spec _anyway_. Plus, Bech32 has some stuff which is very Bitcoin specific (i.e. the witness [version](https://github.com/sipa/bech32/blob/master/ref/haskell/src/Codec/Binary/Bech32.hs#L151)) which doesn't really make sense for us (I think).

What we can surely do, is something like [the Haskell reference implementation](https://github.com/sipa/bech32/blob/master/ref/haskell/src/Codec/Binary/Bech32.hs#L156-L159): you follow Bech32 but your data is not `base32 witProg` but `base32 ourAddressPrefixPlusPayload`, which means this will still be Bech32 compliant and existing decoders will work. However, this means that decoders out in the wild will decode the input address into `(hrp, base32blob)`, but at that point they _will_ need to tap into our spec to understand how to decode each individual field (if that's something they want to do).

The TL;DR is that I am not 100% convinced that the blow up in size and the extra level of complication (plus the need to implement & maintain a BTC-specific-base32-encoder ourselves) is worth the gain, but I am happy to stick to majority here. I am simply challenging a bit the proposal to really make sure we are aware of the full implications before pulling the trigger.

> I don't think CBOR-encoding the payload is a good idea, or is necessary. It adds a lot of complexity to the address format for no gain so I'd prefer if we just dumped the binary form of the key straight into the payload, like bitcoin does it.

Let me make sure I understand (this is what you get when you don't have ☕️ straight away 😁 ): if I understand correctly, you propose essentially to call `CBOR.serialise publicKey` and then dump that directly as a `ByteString` into our `Address`, so that later we can use `CBOR.deserialiseOrFail` to get it back. Essentially, you are proposing to eliminate a layer of indirection and avoid the "intermediate" `CBOR.serialise payload`. I _think_ that should work, but we need to be careful when extending the address, then. I am very, very skeptical that our Address format will remain this way, and sure enough the moment will arrive when we will need to make the `AddressPayload` more complicate than just the `PublicKey`. At that point, I think this should still work (reasoning out loud here) if we write our decoders correctly and we pattern match on the `addressFormat` field of the `AddressPrefix`, so that we can decode either the full `PublicKey` or something else depending on which conditional path we took. _Alfredo ponders if this is really true in practice_.

src/Oscoin/Crypto/Address.hs
```diff
@@ -0,0 +1,97 @@
...
+          , protocolVersion = ProtocolVersion ProtocolVersion_V1
+          }
+        payload = toShort
+                . toS
+                . CBOR.serialise @(ShortHash c)
+                . shortHash
+                $ pk
+    in Address prefix (AddressPayload . Tagged $ payload)
+  where
+    -- NOTE(adn) This is a bit iffy, as we might want to create
```
adinapoli-mndc: I think this is what I *really* need, alongside the environment: https://github.com/oscoin/oscoin/pull/419/files#diff-2b4a1052f78920146b335fda52b51cd2R68

adinapoli-mndc:
@cloudhead ok, your point 2. has been addressed now.

After some thinking I think I'm ok with 1. as well, provided we state clearly that we provide bech32-compliant addresses but this is limited to the encoding format, and it's up to the particular application to read the Oscoin spec to interpret the `dat` part correctly.

_Me proceeds to code the above before it's morning in the west coast._

xla:
> Me proceeds to code the above before it's morning in the west coast.

:running_man: :running_woman:

cloudhead:
> I hear what you are saying here, but is this really true, though? If developers are going to decode our Bech32-encoded-address, what they will decode will be `(hrp, dat)` , where `dat` will be _our_ format, not a segwit address, which means they will need to read our spec _anyway_. Plus, Bech32 has some stuff which is very Bitcoin specific (i.e. the witness [version](https://github.com/sipa/bech32/blob/master/ref/haskell/src/Codec/Binary/Bech32.hs#L151)) which doesn't really make sense for us (I think).

You might be right here - maybe it's not a clear win after all. If it adds complexity that is bitcoin specific, then probably not the best idea.

> Let me make sure I understand (this is what you get when you don't have  straight away  ): if I understand correctly, you propose essentially to call `CBOR.serialise publicKey` and then dump t...

That's not what I meant actually - I mean why use CBOR at all. CBOR adds tags which we don't need here, and requires CBOR to decode/encode. What I'm suggesting is to just put the binary data as-is in the address, the way other blockchains do it. Since everything is a fixed length, this is quite easy.

adinapoli-mndc:
> You might be right here - maybe it's not a clear win after all. If it adds complexity that is bitcoin specific, then probably not the best idea.

There _are_ good ideas to be stolen, though. For example, the use of z-Base32. If we need to go with our own format _anyway_, it might make sense to use z-base32, as that helps a lot with typing characters on a mobile keyboard, as everything will be lower case by default. I will play with different ideas and we will discuss them when you are back in full service.

>  I think we should make sure our address format doesn't change, this is why we have the meta-data fields in there.

But that's simply delusional, my friend 😁 even BTC had to change things over the years, and sure enough we _will_ need to amend the `AddressPayload`. What we shouldn't change is the `AddressPrefix`, and this is why I am allocating two extra unused bytes for future evolution.

> I mean why use CBOR at all. CBOR adds tags which we don't need here, and requires CBOR to decode/encode. What I'm suggesting is to just put the binary data as-is with no "format", the way other blockchains do it. Since everything is a fixed length, this is quite easy.

I am still going to stick to my guns on CBOR (or any other serialisation format) unless you can make a very strong case this is not needed (pinging @kim here as he surely has a lot of experience with coding/decoding data over the wire). To me, not using a serialisation format but instead bump some binary data is not a good idea, for a number of reasons:

a. In order to convert a `Hash (PublicKey c)` into some blob of bytes, you have only two choices: you either go via its `Buildable` instance (i.e. you use one of the `encodeBaseX` function) or you use a proper serialisation library like CBOR. You might be surprised that actually the _size_ of the base58 encoded `hash(pk)` in bytes was _greater_ than its size in CBOR. If you want hard numbers:

```
> import qualified Codec.Serialise as CBOR
> import Oscoin.Crypto.Address
> import Oscoin.Crypto
> import Oscoin.Crypto.PubKey
> import Oscoin.Crypto.PubKey.RealWorld 
> import Oscoin.Crypto.Hash.RealWorld 
> import Oscoin.Crypto.Hash
> import qualified Data.ByteString as BS
> import Data.ByteString.BaseN
> (pk :: PublicKey Crypto, _) <- generateKeyPair
> let (h :: Hashed Crypto (PublicKey Crypto)) = hash pk
> Prelude.show (fromHashed h)
"FKLWsfUVgX2BzNp4a6eo3i1L1VT4kmACPQCBwityaBvt"
> length it
44
> let ser = CBOR.serialise (fromHashed h)
> BS.length (toS ser)
39
```

b. CBOR (or protobuf, or whatever) gives us extensibility and backward compatibility. So I am not totally convinced this is going to "add complexity for no gain" 😉 . I am still convinced our `AddressPayload` will change over time, and once you will have anything more complex than just a `hash(pk)`, things will start to get messy: now you have a binary blob and have to interpret it, how do you do that? You chop the first X bytes from the blob to get the base58 string and then you parse the rest? But wait, a base58 string is of variable size and we want to change the algorithm used to hash our `PublicKey`, so now you have to do conditional logic in order to get exactly how many bytes you need to recostruct the key ... [..] . I mean, you get my point 😅 I would argue this is _more_ complex than just using plain CBOR. _Of course_, the fact we have the meta data fields means we can always write some (pseudo)code like this:

```haskell
case tag of
  AddressFormatTag_V1 -> -- decodeBase58 the Hash(PublicKey)
  AddressFormatTag_V2 -> -- do something else, maybe the HashAlgorithm used to compute the hash changed
  AddressFormatTag_v3 -> do -- let's assume now you have a hash + an extra data structure, a list of delegation keys, because it's 2022 and we switched to PoS
      x <- parseHash
      -- ?? Now you have to start writing crazy hand-rolled code we parse all the keys, you are reinventing CBOR.
```

At this point, sure enough no 3rd party client would like to switch from CBOR (which has a spec) to a crazy hand-rolled binary format in order to decode `AddressFormatTag_v3` 😅 

c. Let's not forget that CBOR is an IETF standard, and has support for canonical form (a property you _do want_ encoders to respect) and generally speaking it gives me a nice, warm, fuzzy feeling 😁 

I obviously don't want to be _that guy_ and, as always, I am challenging the brains here to see if we can do better, but ultimately I am going to stick to what majority think is best. My humble proposal would be to let this marinade over the weekend, reconvene next week the full team, and take an informative decision. Sounds fair? 😉

adinapoli-mndc:
@cloudhead @kim @geigerzaehler unrelated to our bigger discussion re: CBOR vs no CBOR, I had a play with using `z-base32-bytestring` (the library I have polished from `base32-bytestring` which implements [human-friendly base32 encoding](https://philzimmermann.com/docs/human-oriented-base-32-encoding.txt)) and the result can be seen as part of  https://github.com/oscoin/oscoin/pull/431/commits/727c827458fe5f16ce7e55d00e5d7a69f0db7e9f.

Our addresses got a bit more beefier (54 bytes) but this should be hopefully compensated by the improved end-user ergonomics. Let me know what you think about it, we can either keep this or scrap it, I don't mind (but I think improved ergonomics is a good thing 😉 ). They look like this:

```
yrbonyyykikqb4w15bumi9zmtjtrwscrwbf8qkjjhtobndymdrfbe
```

Note there is no "human readable part", as we decided it wasn't worth the effort to go full-bech32.

cloudhead:
> There _are_ good ideas to be stolen, though. For example, the use of z-Base32. If we need to go with our own format _anyway_, it might make sense to use z-base32, as that helps a lot with typing characters on a mobile keyboard, as everything will be lower case by default. I will play with different ideas and we will discuss them when you are back in full service.
> 

Sounds good!

> > I think we should make sure our address format doesn't change, this is why we have the meta-data fields in there.
> 
> But that's simply delusional, my friend  even BTC had to change things over the years, and sure enough we _will_ need to amend the `AddressPayload`. What we shouldn't change is the `AddressPrefix`, and this is why I am allocating two extra unused bytes for future evolution.
> 
Ah if that's what you meant, then yes I agree.


> I am still going to stick to my guns on CBOR (or any other serialisation format) unless you can make a very strong case this is not needed (pinging @kim here as he surely has a lot of experience with coding/decoding data over the wire). To me, not using a serialisation format but instead bump some binary data is not a good idea, for a number of reasons:
> 
> a. In order to convert a `Hash (PublicKey c)` into some blob of bytes, you have only two choices: you either go via its `Buildable` instance (i.e. you use one of the `encodeBaseX` function) or you use a proper serialisation library like CBOR.

> You might be surprised that actually the _size_ of the base58 encoded `hash(pk)` in bytes was _greater_ than its size in CBOR. If you want hard numbers:

Is this surprising - maybe I'm missing something, but you're comparing a binary format (CBOR) with a human-readable format (base58). If you compare `hash(pk)` to `cbor(hash(pk))`, then it should be equal or smaller.

> ```
> > import qualified Codec.Serialise as CBOR
> > import Oscoin.Crypto.Address
> > import Oscoin.Crypto
> > import Oscoin.Crypto.PubKey
> > import Oscoin.Crypto.PubKey.RealWorld 
> > import Oscoin.Crypto.Hash.RealWorld 
> > import Oscoin.Crypto.Hash
> > import qualified Data.ByteString as BS
> > import Data.ByteString.BaseN
> > (pk :: PublicKey Crypto, _) <- generateKeyPair
> > let (h :: Hashed Crypto (PublicKey Crypto)) = hash pk
> > Prelude.show (fromHashed h)
> "FKLWsfUVgX2BzNp4a6eo3i1L1VT4kmACPQCBwityaBvt"
> > length it
> 44
> > let ser = CBOR.serialise (fromHashed h)
> > BS.length (toS ser)
> 39
> ```
> b. CBOR (or protobuf, or whatever) gives us extensibility and backward compatibility. So I am not totally convinced this is going to "add complexity for no gain"  . I am still convinced our `AddressPayload` will change over time, and once you will have anything more complex than just a `hash(pk)`, things will start to get messy: now you have a binary blob and have to interpret it, how do you do that? You chop the first X bytes from the blob to get the base58 string and then you parse the rest? But wait, a base58 string is of variable size and we want to change the algorithm used to hash our `PublicKey`, so now you have to do conditional logic in order to get exactly how many bytes you need to recostruct the key ... [..] . I mean, you get my point  I would argue this is _more_ complex than just using plain CBOR. _Of course_, the fact we have the meta data fields means we can always write some (pseudo)code like this:

Do you have an example where the address payload is not just `hash(pk)`? We already have a bunch of fields in-front of it, the payload should always just be the hash of the public key..

So the format you're proposing is something like:
```
<address-prefix><cbor(hash(pk))><checksum>
```
vs.
```
<address-prefix><hash(pk)><checksum>
```
which I'm proposing, correct?

kim:
> Do you have an example where the address payload is not just hash(pk)? We already have a bunch of fields in-front of it, the payload should always just be the hash of the public key..
> 

Obvious places to look ;):

https://github.com/input-output-hk/cardano-sl/blob/ea7e4319ea7651194e2a483bdef8c60606afe6f4/core/src/Pos/Core/Common/Address.hs
https://github.com/input-output-hk/cardano-ledger/blob/06e0f69f5e061d2b544f175411cfe87446af4652/src/Cardano/Chain/Common/Address.hs

The former has some visible traces of amending a previous version.

I’m clearly not up to snuff on what role addresses play in various approaches to crypto ledgers, but it is quite obvious that the format ought to be extensible and potentially even compatible in both directions. Thus, using an established tagged serialisation format for the “payload” part, which satisfies these properties seems sensible.

However, the 3rd byte in the prefix seems unnecessary, and potentially even counterproductive then: suppose an older version of the software sees an address with a higher version than it knows. There is nothing it can do than to reject it. Careful evolution of the payload “schema” could however allow to simply ignore the unknown tags, and still do something useful with the known ones.

adinapoli-mndc:
Morning good colleagues 😉 

> Is this surprising - maybe I'm missing something, but you're comparing a binary format (CBOR) with a human-readable format (base58). If you compare hash(pk) to cbor(hash(pk)), then it should be the other way around, unless cbor is identity.

Maybe I am the one missing something here, but at the moment we don't have a way to simply convert a `hash(pk)` into a raw sequence of bytes if not going via its `Buildable` instance, which does `base58` encoding. If you remember the design discussions we had with @kim , we debated a lot on the `Hash` API and we concluded that a `toByteArray` function would have be a code smell. Therefore, if you have a `RealWorld` hash (which is a cryptonite `Digest`) then yes, you can just take the raw bytes out of it, but for a `MockHash` that is not immediately obvious (it's a `Word32`). This is why we decided to route everything into the `build` function.

> Do you have an example where the address payload is not just hash(pk)? We already have a bunch of fields in-front of it, the payload should always just be the hash of the public key..

Kim posted a bunch of examples where that wasn't true 😉 Of course I don't have the crystal ball, so I can only try to speculate and use my previous experience as an indicator of whether or not we should try to be flexible with these things.

> From my understanding the base58 (or 32) is the whole address, why would you encode the payload in base58?

I think the culprit here is that if we assume a `hash(pk)` is just a blob of bytes, then yes, you don't need base58. However (as written above) we have an extensible crypto layer and we can have multiple implementations for the crypto, and not all map into a `ByteString`. So I think what we might be facing here (regardless of the CBOR discussion) is that our current API is not flexible enough (and perhaps we should remedy that?)

> It would make more sense to put the hash as-is until we need structure in the payload, and when that day comes, updated the prefix format number and think about how we want to encode the payload, no?

Albeit it wouldn't me my first choice (I guess at this point it boils down to different views on things) that's potentially doable, _if_ we amend our `HasHashing` class to support this `toRawBytes` function, assuming it makes sense to have for things which are not "real" cryptos. _Alfredo tries to think about the implications of different crypto hashes having different sizes, but that shouldn't affect decoding, in theory._

> However, the 3rd byte in the prefix seems unnecessary, and potentially even counterproductive then: suppose an older version of the software sees an address with a higher version than it knows. There is nothing it can do than to reject it. Careful evolution of the payload “schema” could however allow to simply ignore the unknown tags, and still do something useful with the known ones.

The original reason why I added that 3rd byte, wasn't so much as a replacement for CBOR (for which I agree that you want to use a "careful evolution of the schema", and for that you don't need that byte) but for more subtle things, like changes in the underlying crypto library (i.e. it has a bug and now the hashes it produces are different), changes in the underlying hash algorithm (Blake2b_256 is brute forced and we need to switch to Blake2b_512) etc. In a nutshell, _all those things which wouldn't be captured by a new AddressPayload datatype because their evolution is lost by the "loss of precision" of certain things being just binary blobs"_. But perhaps we don't need all of that, and perhaps the `formatVersion` is very, very misleading as a name 😉

cloudhead:
I see, I didn't realise we had to go through base58 - I think this would be the first thing to fix, as that would be a double base58 encoding: `base58(<prefix> || base58(payload))`, which isn't desirable, correct?

I think I've warmed up a little to encoding the payload as CBOR, but still feel that it is a somewhat unnecessary extra step right now, when we have designed an upgradable address format, and the current needs don't require a structured payload. We're forcing all clients that deal with addresses now to use CBOR, but we aren't making use of what CBOR has to offer, ie. encoding structured data.

adinapoli-mndc:
> I think this would be the first thing to fix, as that would be a double base58 (or 32) encoding: base58(<prefix> || base58(payload)), which isn't desirable, correct?

Correct -- I suspect this is something where @kim has a stronger opinion on, so I will appreciate his input. If we add an extra `toRawBytes :: Hash a -> ByteString` function to our `HasHashing` typeclass, would this be iffy or would you all be OKish with that?

kim:
@cloudhead @adinapoli-mndc Did my point get through? If we skip the versioning byte and use CBOR (or some other tagging) for the payload, the address can be backwards _as well_ as forwards compatible.

xla:
> Note there is no "human readable part", as we decided it wasn't worth the effort to go full-bech32.

Where was this concluded?

I'm fully aware this might be an orthogonal concern and not at all the right place to constrain the outcome of the address exploration. Yet the angle I'm coming from is that we don't yet established how we going to do identity in the system (as a usability concern/seen by the user) and it's a hard topic. One of the ways we can have a baseline of identity are addresses as they are omni-present so if we find a way to make them as telling as possible we could gain a lot of usabillity with little effort.

> * The human-readable part is nice, and I agree with @xla that is compelling to use the HRP
>   (_human readable part_) as a sort of namespace per org/project/user, but I also think for our use
>   case this is not _too_ important, as we have a mechanism to distribute rewards automatically and
>   in general micro-transactions are less frequent than a standard cryptocurrency, i.e. the
>   namespacing will end up being more of a "vanity" feature rather than an actually useful one;

This only comes to fruition once financial contracts are activated and for that use-case I absolutely agree. Yet there will be plenty of touchpoints where the user will be exposed to addresses, esp. during exploration which is one of the first exposures to the system.

> * If the HRP is abused, this can effectively lead to the generation of
>   [vanity addresses](https://allprivatekeys.com/vanity-address.php) which could be easily be
>   exploited. I am thinking about draconian cases when somebody would generate addresses like
>   `osclinuxfoundationn1.....` and the distracted user might be fooled into making a donation to a
>   scam address.

We ought to explore the thread models for sure, but as I understand the strength (please correct where my understanding falls short) of bech32 is that the user only needs to verify an `hrp` and a very short `checksum` (6 characters from the original proposal, can be tweaked for differnt guarantees) in order to gain confidence. In your example the user would have to ignore the corectness of the hrp as well as the checksum prefix. To iterate what we are optimising, it's not that the characters in the address are easy to enter on certain devices by a human, rather that with little error rate to detect if the address matches the address that is the designated one.

> * Even if we don't allow full "namespacing", we will need to pick the HRP tag and
>   [claim it](https://github.com/satoshilabs/slips/blob/master/slip-0173.md) which might be good to
>   do regardless we end up using Bech32 or not, but it feels a bit constricting.

This is only true if we operate on one of the bitcoin networks, unless I understand the current maintained hrp registry they have wrong.

cloudhead:
> @cloudhead @adinapoli-mndc Did my point get through? If we skip the versioning byte and use CBOR (or some other tagging) for the payload, the address can be backwards _as well_ as forwards compatible.

Hmm yeah that's interesting, so:
```
[{version: "v1", pk: <key>}]
```
instead of
```
["v1", <key>]
```
Essentially you're saying that "premature tagging" gives us forward-compatibilty, if I understand correctly.

I think this is a solid argument for *tagging* at least. I remain to be convinced about CBOR, if there are simpler tagging systems, but maybe CBOR is simple enough?

cloudhead:
> To iterate what we are optimising, it's not that the characters in the address are easy to enter on certain devices by a human, rather that with little error rate to detect if the address matches the address that is the designated one.

To rephrase this: random addresses all look alike, but if you have something like `alexis91389afadjf3ifj`, you won't confuse it with `alfredo9fsf89j2fkjh`

cloudhead:
Another thing that comes to mind with regards to having an explicit `version` prefix in the address format, is that if we use CBOR for example, and we *don't* have an address format prefix, we are stuck with CBOR for eternity.

adinapoli-mndc:
OK, me, @cloudhead and @kim gather together for a brainstorm session and we finally reached consensus. This is what we will do:

* For now we are not going to support full Bench-32 (i.e. the human readable part). That can always be added later as part of our schema/address migration strategy, if we want;

* We agreed that encoding the payload with CBOR is actually fine, mostly because application code has to depend on CBOR _anyway_ (in order to use the API or decode blocks) so this doesn't impose such a heavy dependency on people. Plus, CBOR has other nice advantages we discussed before;

* In order to have fine-grained control over the serialisation format, we agreed that it's a sensible idea to write the `Serialise` instances by hand for an `Address` rather than relying on the `Generic` deriving. At some point we should probably extend this schema to every single instance in the codebase, as that allows us to migrate them via the "tagged" approach;

* We agreed that having an `addressFormat` reserved byte is useful, but that at the moment is a bit ill-named, as in the types it's given the very opaque tag `AddressFormatTag_V1`. Instead, what we should do is to use this byte to future-proofing our serialisation format, i.e. use it in case we would like to switch to a different serialisation format. To do that, naming this like `AddressFormatTag_CBOR` might make more sense, as well as reflecting this choice in the spec (yet to write);

* We agreed to scrap the intermediate `encodeBase58` call when we serialise the `hash(pk)`.. In order to do that, it might be sufficient to act on the `Buildable` instance for the `RealWorldCrypto`. If not, we might need to add a `toRawBytes` function to the `HasHashing` typeclass, in order to accomplish that.

src/Oscoin/Crypto/Address.hs
```diff
@@ -0,0 +1,111 @@
...
+
+{------------------------------------------------------------------------------
+    Creating an Address
+------------------------------------------------------------------------------}
+
+data AddressCreationError =
+    UnsupportedConfiguration (Config.Network, Config.Environment)
+    deriving Show
+
+-- | NOTE(adn) It's conceivable that we might want to crate a type to carry
```
cloudhead: Can we move the note somewhere else and have a haddock for this?

cloudhead:
```suggestion
-- | NOTE(adn) It's conceivable that we might want to create a type to carry
```

cloudhead: Perhaps it shouldn't be in the code, but just serves for the PR?

adinapoli-mndc: Oh yeah, totally -- what I was planning to do was not to use this module for documentation purposes (as it's actually useful to keep technical notes in there) but rather write proper documentation at `docs/address_format.md`. How does that sound?

src/Oscoin/Crypto/Address.hs
```diff
@@ -0,0 +1,111 @@
...
+-- around the context needed to create an 'Address', but for now single
+-- arguments should do.
+fromPublicKey
+    :: forall c.
+       ( ByteArrayAccess (ShortHash c)
+       , Hashable c (PublicKey c)
+       )
+    => Config.Network
+    -- ^ The 'Network' the node is running on.
+    -> Config.Environment
```
cloudhead: Having to pass an environment every time you want to generate an address may become a problem, because now we have to thread it everywhere. I guess one option is to partially apply this function and store the partially-applied version somewhere in the node ... thoughts?

src/Oscoin/Crypto/Address/Bech32.hs
```diff
@@ -0,0 +1,58 @@
+{-- | Partial, opinionated support for Bech32.
```
cloudhead:
```suggestion
{-- | Partial, opinionated support for Bech32 (BIP 173).
```

src/Oscoin/Crypto/Address.hs
adinapoli-mndc: The way I see this module, it's that it's a low-level one, which exposes just the building blocks you need to construct more abstractions on top of it. As such, we can surely do the "partially applied" trick elsewhere, methinks. Wdyt?

src/Oscoin/Crypto/Address/Bech32.hs
cloudhead:
```suggestion
-- | Partial, opinionated support for Bech32 (BIP 173).
```

src/Oscoin/Crypto/Address/Bech32.hs
```diff
@@ -0,0 +1,58 @@
...
+-- Mostly taken and adapted from the reference implementation available
+-- <here https://github.com/sipa/bech32/blob/master/ref/haskell/src/Codec/Binary/Bech32.hs>
+--}
```
cloudhead:
```suggestion
--
```

src/Oscoin/Crypto/Address/Bech32.hs
```diff
@@ -0,0 +1,58 @@
...
+module Oscoin.Crypto.Address.Bech32
+    ( createChecksum
+    , verifyChecksum
+
+    -- * Interals
```
cloudhead:
```suggestion
    -- * Internals
```
`:set spell` :wink:

src/Oscoin/Crypto/Address/Bech32.hs
```diff
@@ -0,0 +1,58 @@
...
+
+import           Data.Bits
+import qualified Data.ByteString as BS
+import           Oscoin.Crypto.Address.Internal (Checksum(..))
+
+(.>>.), (.<<.) :: Bits a => a -> Int -> a
+(.>>.) = unsafeShiftR
+(.<<.) = unsafeShiftL
+
+type HumanReadablePart = ByteString
```
cloudhead: Note sure what this type synonym gives us - It makes the function signatures slightly more confusing to me.

src/Oscoin/Crypto/Address/Bech32.hs
adinapoli-mndc: You might be surprised by I _do_ code with `:set spell` on. The problem is -- it doesn't persist across buffer changes, so after I while, I forget to re-enable it 😂

src/Oscoin/Crypto/Address/Bech32.hs
adinapoli-mndc: we can surely scrap it -- what it gives you is essentially a bit of consistency with the original Bench32 spec, as they explicitly introduces the concept of "HumanReadablePart". I'm not attached to it, so I'm gonna scrap it 😬

src/Oscoin/Crypto/Address/Internal.hs
```diff
@@ -0,0 +1,125 @@
...
+    | AddressTypeTestNet -- (Testnet, Production)
+    | AddressTypeDevnet  -- (Devnet,  Production)
+    -- Other exotic permutations
+    | AddressTypeMainnetDev -- (Mainnet, Development)
+    | AddressTypeMainnetTst -- (Mainnet, Testing)
+    | AddressTypeTestNetDev -- (Testnet, Development)
+    | AddressTypeTestNetTst -- (Testnet, Testing)
+    | AddressTypeDevnetDev  -- (Devnet,  Development)
+    | AddressTypeDevnetTst  -- (Devnet,  Testing)
+    deriving (Eq, Show)
```
cloudhead: Going through this extra type seems to add a lot of boilerplate, why not just always convert from `(network, env) -> Word8` and vice versa?

src/Oscoin/Crypto/Address.hs
cloudhead: Yeah sounds good.

src/Oscoin/Crypto/Address/Internal.hs
adinapoli-mndc: We could -- however, I think this is exactly the boilerplate you want in these cases, i.e. this is a layer of indirection that allows you to not couple yourself with the (network, env) tuple just yet (in case that grows in the future). Plus, it's quite scary to have to deal directly with these kind of conversions, as it's quite easy to mistype something and end up with the wrong isomorphism (despite our roundtrip tests helps here, of course).

src/Oscoin/Crypto/Address/Serialisation.hs
```diff
@@ -0,0 +1,249 @@
+{-# LANGUAGE UndecidableInstances #-}
+{-# OPTIONS_GHC -fno-warn-orphans  #-}
+module Oscoin.Crypto.Address.Serialisation
+    ( DeserialiseError(..)
+    , serialiseAddress
+    , deserialiseAddress
```
cloudhead: I'm not sure this is a real concern, and I partly blame the `serialise` library for the uk spelling, but the rest of our codebase uses american english, (grep `initialize`, `unauthorized`, etc.) - so might be easier to remember than having to think about which functions have uk spelling.

cloudhead: If you decide to change this, also grep for `deserialise`

adinapoli-mndc: I think consistency is important 😉 I am myself into the uk spelling camp (coming from a uk-heavy working environment legacy, hehe), so I need some nudge to keep me honest on that 😁

adinapoli-mndc: 84c5a8f

src/Oscoin/Crypto/Address/Internal.hs
adinapoli-mndc: 84c5a8f

src/Oscoin/Crypto/Address/Bech32.hs
adinapoli-mndc: 84c5a8f

src/Oscoin/Crypto/Address/Bech32.hs
adinapoli-mndc: 84c5a8f

src/Oscoin/Crypto/Address.hs
adinapoli-mndc: 84c5a8f

adinapoli-mndc: 84c5a8f

src/Oscoin/Crypto/Address/Bech32.hs
adinapoli-mndc: 84c5a8f

src/Oscoin/Crypto/Address.hs
cloudhead: Yeah, makes sense :+1:

src/Oscoin/Configuration.hs
```diff
@@ -0,0 +1,11 @@
+{-- | Umbrella module which gathers all the types which, when combined,
+--- determines the particular configuration an Oscoin node is running with.
+--}
```
cloudhead: The `{}` is not necessary here - same comment for the other module headers :jack_o_lantern:

src/Oscoin/Crypto/Address/Internal.hs
```diff
@@ -0,0 +1,112 @@
...
+  Miscellanea functions
+------------------------------------------------------------------------------}
+
+configToAddressTypeTag :: Network -> Maybe AddressTypeTag
+configToAddressTypeTag = \case
+  Mainnet -> Just AddressTypeMainnet
+  Testnet -> Just AddressTypeTestNet
+  Devnet  -> Just AddressTypeDevnet
+  -- NOTE(adn) The 'Somenet' constructor is not exported, which makes this
+  -- pattern-match non-exhaustive.
```
cloudhead: Somenet? :confused:

cloudhead: Ok I just found out about `Somenet` ... do we want to export it and reserve `0x0` for it? Or perhaps the `Word8` should be gotten from an `Enum` instance on `Network`? @kim ?

kim: `Somenet` came from the strange situation that, when using multicast discovery, you want to make sure to not accidentally joining a network you didn’t mean to. So, it’s essentially “randomnet”. 

The same applies if we want to allow namespaced deployments to k8s - the DNS discovery would assume the namespace to equal the network name.

Idk, maybe there’s some other way to distinguish these.

src/Oscoin/Configuration.hs
adinapoli-mndc: Old habits!

src/Oscoin/Crypto/Address/Internal.hs
adinapoli-mndc: If we do export `Somenet` that would already be a step forward, because in this pattern match (which can now be made exhaustive) we can yield `Nothing`. This would affect only a situation where the user (or the node) purposefully call `fromPublicKey` with a somennet, which will be rejected anyway from the system.

Forward compatibility in not an issue -- when `Ubernet` will be added (say) we will simply associate a fresh `Word8` to it. I think that the morally correct thing is to _not_ assign `somenet` any word8. It's essentially "user error" and as such it shouldn't be handled/represented in the system.

src/Oscoin/Configuration.hs
adinapoli-mndc: ps. It shouldn't matter too much, as @kim's PR [here](https://github.com/oscoin/oscoin/pull/459) introduces essentially the same module. I'll tell you what, I am going to merge the aforementioned PR (as it has 2 green reviews) and gonna deal with the ~pain~ conflicts 😅

adinapoli-mndc:
@cloudhead I was also thinking that in order to not inflate this PR too much, it might be worthwhile dedicating a separate ticket to the task of writing proper documentation, so we can merge this one relatively soon. Thoughts?

adinapoli-mndc:
Also fixed the Haddock docs a bit as they were rendering like crap before 😂 

Screenshot 2019-04-04 at 10 30 14

I do realise how writing _proper_ Haddock-annotated Haskell code is more art than science, and a full time job 😁

src/Oscoin/Crypto/Address/Internal.hs
cloudhead: :+1:

src/Oscoin/Crypto/Address.hs
```diff
@@ -0,0 +1,106 @@
...
+          let prefix = AddressPrefix
+                { addressFormat   = AddressFormat AddressFormatTagCBOR
+                , addressType     = AddressType at
+                , protocolVersion = ProtocolVersion ProtocolVersion_V1
+                }
+              payload = Tagged . CBOR.fromByteString . convert . shortHash @c $ pk
+          pure $ Address prefix (AddressPayload payload)
+
+-- | Parses a base32z-encoded binary block back into an 'Address'.
+parseAddress
```
cloudhead: Should be `decodeAddress` if it's operating on binary data

src/Oscoin/Crypto/Address.hs
```diff
@@ -0,0 +1,106 @@
...
+
+-- | Renders an 'Address' into its textual form (i.e. a z-base-32-encoded
+-- blob).
+renderAddress
+    :: ( CBOR.Serialise (ShortHash c)
+       , ByteArrayAccess (ShortHash c)
+       )
+    => Address c
+    -> Text
+renderAddress = F.sformat fmtAddress
```
cloudhead: Can we call this `showAddress` to keep with common vocabulary?

src/Oscoin/Crypto/Address/Internal.hs
```diff
@@ -0,0 +1,114 @@
...
+    } deriving (Eq, Show)
+
+data AddressPrefix = AddressPrefix
+    { protocolVersion :: ProtocolVersion
+    , addressType     :: AddressType
+    , addressFormat   :: AddressFormat
+    } deriving (Eq, Show)
+
+-- | A 'ProtocolVersion' is a byte representing the current version of the
+-- Oscoin protocol running. This makes possible to issue forks of the system
```
cloudhead:
```suggestion
-- Oscoin protocol running. This makes it possible to issue forks of the system
```

src/Oscoin/Crypto/Address/Internal.hs
```diff
@@ -0,0 +1,114 @@
...
+-- 0x01 Mainnet
+-- 0x02 Testnet
+-- 0x03 Devnet
+-- ..
+newtype AddressType =
+    AddressType AddressTypeTag
+    deriving (Eq, Show)
+
+data AddressTypeTag =
+      AddressTypeMainnet -- (Mainnet, Production)
```
cloudhead:
```suggestion
      AddressTypeMainnet -- Mainnet
```

cloudhead: etc.

src/Oscoin/Crypto/Address/Internal.hs
```diff
@@ -0,0 +1,114 @@
...
+-- Oscoin protocol running. This makes possible to issue forks of the system
+-- by simply acting on this value, so that nodes running the old protocol
+-- won't be able to use addresses from the new fork. Conversely, nodes
+-- running the fork can be programmed to support both formats.
+newtype ProtocolVersion =
+    ProtocolVersion ProtocolVersionTag
+    deriving (Eq, Show)
+
+data ProtocolVersionTag =
+      ProtocolVersion_V1
```
cloudhead: If we have an underscore here, we should make it consistent and have `AddressFormatTag_CBOR` no? Or remove the underscore here.

src/Oscoin/Crypto/Address/Internal.hs
```diff
@@ -0,0 +1,114 @@
...
+    AddressFormat AddressFormatTag
+    deriving (Eq, Show)
+
+data AddressFormatTag =
+    AddressFormatTagCBOR
+    deriving (Eq, Show)
+
+-- | The 'AddressPayload' is simply the hash of a 'PublicKey'.
+--
+-- NOTE(adn): The use of 'Tagged' here is due to the fact we cannot encode
```
cloudhead: General comment on these notes: Since they are meant for haddocks, I don't think it makes sense to "sign" them :grin:

src/Oscoin/Crypto/Address/Serialisation.hs
```diff
@@ -0,0 +1,237 @@
...
+      NotEnoughInput  ByteString
+      -- ^ The deserialisation failed as the decoder was expecting more
+      -- input.
+    | DeserializePayloadError CBOR.DeserialiseFailure
+      -- ^ The deserialisation process failed while deserialising the
+      -- 'AddressPayload'.
+    | UnknownProtocolVersion Word8
+      -- ^ When decoding the 'ProtocolVersion', we encountered an unknown tag.
+    | UnknownAddressType Word8
+      -- ^ When decoding the 'AddressType'', we encountered an unknown tag.
```
cloudhead:
```suggestion
      -- ^ When decoding the 'AddressType', we encountered an unknown tag.
```

src/Oscoin/Crypto/Address/Serialisation.hs
```diff
@@ -0,0 +1,237 @@
...
+    | UnknownAddressFormat Word8
+      -- ^ When decoding the 'AddressFormat, we encountered an unknown tag.
```
cloudhead:
```suggestion
      -- ^ When decoding the 'AddressFormat', we encountered an unknown tag.
```
Not sure how the missing quote ended up one line above

src/Oscoin/Crypto/Address.hs
adinapoli-mndc: Probably more effective, yup.

src/Oscoin/Crypto/Address/Internal.hs
```diff
@@ -0,0 +1,114 @@
...
+-- 0x02 Testnet
+-- 0x03 Devnet
+-- ..
+newtype AddressType =
+    AddressType AddressTypeTag
+    deriving (Eq, Show)
+
+data AddressTypeTag =
+      AddressTypeMainnet -- (Mainnet, Production)
+    | AddressTypeTestNet -- (Testnet, Production)
```
cloudhead:
```suggestion
    | AddressTypeTestnet -- (Testnet, Production)
```

src/Oscoin/Crypto/Address.hs
adinapoli-mndc: We do have a bunch of "render" stuff, i.e. [renderEnvironment](https://github.com/oscoin/oscoin/blob/790c8cbe3c369a1875a9dcb29a0220d38e04feb7/src/Oscoin/Configuration.hs#L10) and [renderNetwork](https://github.com/oscoin/oscoin/blob/790c8cbe3c369a1875a9dcb29a0220d38e04feb7/src/Oscoin/Configuration.hs#L16). I think the "unspoken" rule I have been using is that `showBlah` is for things which either return a `String` or are close to the `Show` class (i.e. machine readable things), whereas `render` is for human.

I am flexible, though 😉

src/Oscoin/Crypto/Address/Internal.hs
adinapoli-mndc: Right, this is one of those hlint controversy: I tend to like the underscore better in these cases as at glance it would be easy to scan the different suffixes. However, when I did introduce the CBOR one, hlint complained so I did change it to camelcase.. and for some reason hlint didn't bark for these other 🤦‍♂️ .

Back to your original question: my pref would be underscores all the way down (with Hlint annotations to suppress the warnings) but I appreciate this might not be everyones' cup of tea, so please vote 😅

src/Oscoin/Crypto/Hash.hs
```diff
@@ -92,7 +108,12 @@ class
     -- | The zero hash. Also the minimum hash value.
     zeroHash  :: Hash crypto
 
-    shortHash :: Hash crypto -> ByteString
+    -- | A short hash, in the similar spirit of the BTC one. This is typically
+    -- ripemd160(some_other_hash).
```
cloudhead: :thinking: - what do you mean by "some other hash"?

src/Oscoin/Crypto/Address/Internal.hs
adinapoli-mndc: ha, good point. It's actually quite a shame that haddocks renders them ( we might possibly hack around that by putting a white space between the "real" comment and the note). I guess signing is an habit which makes one accountable to either cleanup (if it's a FIXME) or it serves as a point of contact for other developers wanting to shed some lights on the comment itself (like: who the hell wrote this piece of crap? Oh, it's Alfredo 😅 ).

I do concur though for Haddock this is a no go. Do you have any tips? Embed them _inside_ the function? 🤔

src/Oscoin/Crypto/Hash.hs
adinapoli-mndc: Good question: this is essentially predated from BTC, see [here](https://en.bitcoin.it/wiki/Base58Check_encoding). In the section "Encoding a Bitcoin address" it reads:

```
Pay-to-pubkey-hash (p2pkh): payload is RIPEMD160(SHA256(ECDSA_publicKey)) where ECDSA_publicKey is a public key the wallet knows the private key for; version 0x00 (these addresses begin with the digit '1')
``` 

Apparently RIPEMD160 is very well suited for these things due to its "avalanche effect". However, we are not doing it 100% BTC-compliant , as we don't use SHA_256 but Blake2b_256.

This is what I meant with "some other hash" 😁

adinapoli-mndc: I think that in the comment, I should have said "This is typically RIPMD160(HashAlgorithm crypto)" or something like that 😅

src/Oscoin/Crypto/Address/Bech32.hs
```diff
@@ -0,0 +1,55 @@
...
+    values = bech32HRPExpand hrp <> dat
+    polymod = bech32Polymod (values <> [0, 0, 0, 0, 0, 0]) `xor` 1
+
+verifyBech32checksum :: ByteString -> [Word8] -> Bool
+verifyBech32checksum hrp dat = bech32Polymod (bech32HRPExpand hrp <> dat) == 1
+
+-- Top-level functions
+
+createChecksum :: ByteString -> Checksum
+createChecksum = Checksum . BS.pack . createBech32checksum mempty . BS.unpack
```
cloudhead: By convention we would say `mkChecksum` or simply `checksum` here, since the function has no side-effects. Same for `createBech32checksum`

adinapoli-mndc: Can do -- I probably like `mkChecksum` a bit better as there is already a `checksum` function is the `Serialisation` module, albeit we import `Bech32` qualified so it shouldn't really matter.

src/Oscoin/Crypto/Hash.hs
cloudhead: Ohhhhhhh. Maybe yeah even `eg. ripemd160(blake2b(x))`

src/Oscoin/Crypto/Address/Internal.hs
cloudhead: I think if it's intended to be cleaned up, then I'd keep the signature, but put it *inside* the function. If it's intended as a permanent note, I'd keep it in the haddock and remove the signature. There's some examples of this in the codebase, we usually use `Nb. Here is my awesome note` style, as it looks nicer in the haddocks.

src/Oscoin/Crypto/Address/Serialisation.hs
adinapoli-mndc: hah, was wondering the same 😂

src/Oscoin/Crypto/Address/Internal.hs
adinapoli-mndc: I would say this is an internal, permanent note, so it might make sense to put it inside. However, I do have a clear example of a "awesome note" style 😁 : in `fmtAddress`, when we warn the user about the function being heavy, that's exactly the kind of "unsigned" `Nb.` content we want to share on Haddock (methinks).

src/Oscoin/Crypto/Address/Internal.hs
cloudhead: Can we generally tell hlint to shutup about this? I think underscores for this makes sense, otherwise ItsALotOfCamelCase

adinapoli-mndc: my hlint-scripting-foo is fairly weak, but I can look up something on ~StackOverflow~ the manual, to silence the entire "use camelcase" rule (should be fairly trivial , jokes aside).

src/Oscoin/Crypto/Address.hs
cloudhead: Ah my bad, keep it as-is :+1:

src/Oscoin/Crypto/Address/Bech32.hs
adinapoli-mndc: 35cf933

src/Oscoin/Crypto/Hash.hs
adinapoli-mndc: 35cf933

src/Oscoin/Crypto/Address/Internal.hs
adinapoli-mndc: 35cf933

src/Oscoin/Crypto/Address/Serialisation.hs
adinapoli-mndc: 35cf933

src/Oscoin/Crypto/Address/Internal.hs
adinapoli-mndc: 35cf933

src/Oscoin/Crypto/Address/Internal.hs
adinapoli-mndc: 35cf933

src/Oscoin/Crypto/Address/Internal.hs
adinapoli-mndc: 35cf933 ended up removing the comments entirely -- now there weren't really adding anything meaningful to the mix.

src/Oscoin/Crypto/Address/Internal.hs
adinapoli-mndc: 35cf933

src/Oscoin/Crypto/Address.hs
adinapoli-mndc: 35cf933

adinapoli-mndc:
> Great! Only issue remaining is the Somenet one, but if you prefer to handle that separately, that's fine 

It's up to us -- as I am reluctant to expose the constructor (as that's @kim territory and he's partially unavailable today) I am also fine to merge this one first, and tackle this second bit separately once Kim gave us his blessing 😁

adinapoli-mndc:
Gonna merge this guy before it's too late, and fire a second PR to amend the `Somenet` stuff. Any extra review comment @kim has on this PR he's free to add, and will spin up a 3rd PR to incorporate his feedback 😁

PR #433, opened 2019-03-26
tests: Only test CBOR HTTP content type

geigerzaehler:
We only test the CBOR encoding in the HTTP tests and drop all other combinations. This reduces the test time dramatically. On CI from 55 seconds to 35 seconds and even more when running them in the repl locally.

Testing more combinations than CBOR/CBOR only covers more of the encoding/decoding and content type negotiation code. For the near future we will only use the CBOR content type and so this loss of coverage is acceptable.


kim:
> For the near future we will only use the CBOR content type

Are we?

adinapoli-mndc:
Provocative question for the audience: does the 20 seconds cut warrant the loss in test coverage? This seems to go against our key results (which are _extending_ the code coverage, not decreasing it 😅 ).

If running the whole tests every time is slow, is running only a part of the suite a viable option (as in `stack test --ta "--pattern Foo" ` ?

geigerzaehler:
> > For the near future we will only use the CBOR content type
> 
> Are we?

Yes. I briefly talked with @cloudhead about this.Having the JSON content type offers nothing at the moment. But it requires us to provide `ToJSON` and `FromJSON` instances for everything we want to send over the wire and a significant amount of code to do content type negotiation.

> If running the whole tests every time is slow, is running only a part of the suite a viable option (as in `stack test --ta "--pattern Foo" ` ?

The problem is that these tests are test a lot of components. So for a change in any of the tested components we to run these tests. So the tests that we would select with `--pattern` are actually slow.

kim:
Perhaps the right move is then to remove JSON encoding altogether?

kim:
Might be me, but I also can’t find tests which would test the content negotiation and serialisation of all possible response types in isolation. If we had that, I guess stripping down the integration tests in IO would be acceptable.

cloudhead:
Afaik, the only place where JSON is useful is for writing the yaml genesis file, but this requires ToJSON on `Block` and `Tx`. I think it makes sense to keep our API simple with only one encoding and look at how we can simplify the rest too (maybe hard-coding the genesis in a `.hs`) - but not as part of this PR.

kim:
> Afaik, the only place where JSON is useful is for writing the yaml genesis file, but this requires ToJSON on Block and Tx. 

Hm, not necessarily: if this is really the only place we want to use JSON, then perhaps it would even _improve_ things to just define a newtype with hand-coded instances. It is obvious then what fields exist, and what they mean.

> I think it makes sense to keep our API simple with only one encoding and look at how we can simplify the rest too (maybe hard-coding the genesis in a .hs) - but not as part of this PR.

Sure, if clients can deal with CBOR, I have no issue with that. But if a PR removes tests for codepaths because they are considered dead, it makes sense to also remove the dead code (or test it in a less resource-intensive way, as alluded to in https://github.com/oscoin/oscoin/pull/433#issuecomment-476752094). No?

cloudhead:
> Hm, not necessarily: if this is really the only place we want to use JSON, then perhaps it would even improve things to just define a newtype with hand-coded instances. It is obvious then what fields exist, and what they mean.

Yeah that's true and might indeed be a good thing. We potentially don't even need a newtype/instance for this and could just use a pair of functions.

> Sure, if clients can deal with CBOR, I have no issue with that. But if a PR removes tests for codepaths because they are considered dead, it makes sense to also remove the dead code (or test it in a less resource-intensive way, as alluded to in #433 (comment)). No?

Yeah, naturally! I wouldn't completely remove the instances though for now, but rather cleanup what's under `Oscoin.API.HTTP.*` from any ToJSON/FromJSON reference.

geigerzaehler:
Thanks for the input. I’m going to extend this PR by removing the JSON instances from the HTTP API.

PR #434, opened 2019-03-27
Relax crypto hashing constraints

geigerzaehler:
We rename `fromByteArray` to `hashByteArray` and relax the constraint from `ByteArray` to `ByteArrayAccess`.

### DIFF
--- a/src/Oscoin/Crypto/Hash/Mock.hs
+++ b/src/Oscoin/Crypto/Hash/Mock.hs
@@ -50,7 +50,7 @@ instance HasHashing MockCrypto where
         deriving (Eq, Ord)
 
     -- | Hash anything which is an instance of 'ByteArray' into a 'Word64' using
-    -- the xxhash.
+    -- FNV-1.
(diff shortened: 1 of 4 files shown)


src/Oscoin/Crypto/Hash/Mock.hs
```diff
@@ -49,7 +49,9 @@ instance HasHashing MockCrypto where
     newtype Hash MockCrypto = FnvHash { fromMockHash :: FnvHash64 }
         deriving (Eq, Ord)
 
-    fromByteArray = FnvHash . hashByteArray
+    -- | Hash anything which is an instance of 'ByteArray' into a 'Word64' using
+    -- the xxhash.
```
adinapoli-mndc: Note to myself: this comment is now stale, as we _used_ to use `xxhash` as non-cryptographic hashing function, but now we are using `FNV`, and we should probably amend it 😉

kim:
```suggestion
    -- FNV-1.
```

😬

PR #435, opened 2019-03-27
Replace RadTx with DummyTx in BlockStore tests

geigerzaehler:
We replace the use of `RadTx` with `DummyTx` in the block store tests. This is one step towards getting rid of Radicle. The change also improves the run time of the block store tests by roughly 40%.

To achieve the transition we start by extracting `DummyTx` into the `Test.Oscoin.DummyLedger` module.

The interface for the SQLite storage had a `SQLite.ToRow tx` constraint for all transactions that was implemeted for `Tx c Radicle.Value`. A naive implementation of `SQLite.ToRow DummyTx` would lead to incompatibilities with the database schema. For this reason I introduced the `StoreTransaction` class. This constraint needs to be satisfied in order to store transactions in the database.

### DIFF
--- a/src/Oscoin/Storage/Block/SQLite/Transaction.hs
+++ b/src/Oscoin/Storage/Block/SQLite/Transaction.hs
@@ -1,0 +1,143 @@
+{-# LANGUAGE QuasiQuotes          #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+-- | Store and retrieve transaction.
+--
+-- The store can store a transaction type @tx@ with a crypto @c@ if
+-- they satisfy the @'StorableTx' c tx@ constraint. This requires @tx@
+-- to implement 'IsTxRow'. We provide such an instance for 'Tx'.
+module Oscoin.Storage.Block.SQLite.Transaction
+    ( StorableTx
+    , IsTxRow(..)
+    , TxRow(..)
+    , TxRowDecodeError
+    , getTx
+    , getBlockTxs
+    , storeTxs
+    ) where
+
+import           Oscoin.Prelude
+
+import           Oscoin.Crypto.Blockchain.Block (BlockHash)
+import           Oscoin.Crypto.Hash (Hash, Hashable, hash)
+import           Oscoin.Crypto.PubKey
+import           Oscoin.Data.Tx
+
+import           Codec.Serialise
+import qualified Data.ByteString.Lazy as LBS
+
+import           Database.SQLite.Simple as Sql (Connection, Only(..))
+import qualified Database.SQLite.Simple as Sql
+import           Database.SQLite.Simple.FromRow (FromRow, field)
+import           Database.SQLite.Simple.Orphans ()
+import           Database.SQLite.Simple.QQ
+import           Database.SQLite.Simple.ToField (ToField, toField)
+
+
+data TxRow = TxRow
+    { txRowMessage :: ByteString
+    , txRowAuthor  :: ByteString
+    , txRowChainId :: Word16
+    , txRowNonce   :: Word32
+    , txRowContext :: ByteString
+    }
+
+instance FromRow TxRow where
+    fromRow = do
+        txRowMessage <- field
+        txRowAuthor <- field
+        txRowChainId <- field
+        txRowNonce <- field
+        txRowContext <- field
+        pure $ TxRow{..}
+
+
+newtype TxRowDecodeError = TxRowDecodeError String
+    deriving (Show, Eq)
+
+instance Exception TxRowDecodeError where
+    displayException (TxRowDecodeError msg) =
+        "Failed to decode transaction row: " <> msg
+
+
+
+-- | Contraints that need to be satisfied for storing and retriving
+-- transactions. @c@ is the crypto and @tx@ the transaction data type.
+-- This implies the @'IsTxRow' tx@ constraint.
+type StorableTx c tx = (Hashable c tx, ToField (Hash c), Serialise (Hash c), IsTxRow tx)
+
+class IsTxRow a where
+    toTxRow :: a -> TxRow
+    fromTxRow :: TxRow -> Either TxRowDecodeError a
+
+instance
+    ( Serialise (Hash c)
+    , Serialise (PublicKey c)
+    , Serialise (Signature c)
+    , Serialise msg
+    ) => IsTxRow (Tx c msg) where
+    toTxRow Tx{..} = TxRow
+        { txRowContext = LBS.toStrict $ serialise txContext
+        , txRowMessage = LBS.toStrict $ serialise txMessage
+        , txRowAuthor = LBS.toStrict $ serialise txPubKey
+        , txRowChainId = txChainId
+        , txRowNonce = txNonce
+        }
+
+    fromTxRow TxRow{..} = first toDecodeError $ do
+        txMessage <- deserialiseOrFail $ LBS.fromStrict txRowMessage
+        txPubKey <- deserialiseOrFail $ LBS.fromStrict txRowAuthor
+        txContext <- deserialiseOrFail $ LBS.fromStrict txRowContext
+        pure $ Tx
+            { txMessage
+            , txPubKey
+            , txChainId = txRowChainId
+            , txNonce = txRowNonce
+            , txContext
+            }
+      where
+        toDecodeError :: DeserialiseFailure -> TxRowDecodeError
+        toDecodeError = TxRowDecodeError . displayException
+
+getTx :: forall c tx. (StorableTx c tx) => Sql.Connection -> Hash c -> IO (Maybe tx)
+getTx conn hash' = do
+    rows <- Sql.query conn
+        [sql| SELECT message, author, chainid, nonce, context
+                FROM transactions
+               WHERE hash = ? |] (Only $ serialise hash')
+    forM (listToMaybe rows) fromTxRowThrow
+
+
+storeTxs :: forall c tx. (StorableTx c tx) => Sql.Connection -> Hash c -> [tx] -> IO ()
+storeTxs conn blockHash txs =
+    -- Nb. To relate transactions with blocks, we store an extra block hash field
+    -- for each row in the transactions table.
+    Sql.executeMany conn
+        [sql| INSERT INTO transactions  (message, author, chainid, nonce, context, hash, blockhash)
+              VALUES                    (?, ?, ?, ?, ?, ?, ?) |] (map toTxRowValues txs)
+  where
+    toTxRowValues tx =
+        let TxRow {..} = toTxRow  tx
+        in
+            [ toField txRowMessage
+            , toField txRowAuthor
+            , toField txRowChainId
+            , toField txRowNonce
+            , toField txRowContext
+            , toField $ serialise $ hash @c tx
+            , toField blockHash
+            ]
+
+-- | Get the transactions belonging to a block.
+getBlockTxs :: (StorableTx c tx) => Connection -> BlockHash c -> IO [tx]
+getBlockTxs conn h = do
+    txRows <- Sql.query conn
+        [sql| SELECT message, author, chainid, nonce, context
+                FROM transactions
+               WHERE blockhash = ? |] (Only h)
+    traverse fromTxRowThrow txRows
+
+fromTxRowThrow :: (IsTxRow tx) => TxRow -> IO tx
+fromTxRowThrow txRow = case fromTxRow txRow of
+    Left err -> throw err
+    Right tx -> pure tx
--- a/src/Database/SQLite/Simple/Orphans.hs
+++ b/src/Database/SQLite/Simple/Orphans.hs
@@ -8,7 +8,6 @@ import           Oscoin.Prelude
 import           Oscoin.Crypto.Blockchain.Block
 import qualified Oscoin.Crypto.Hash as Crypto
 import qualified Oscoin.Crypto.PubKey as Crypto
-import           Oscoin.Data.RadicleTx
 
 import qualified Codec.Serialise as CBOR
 import qualified Data.ByteString.Lazy as LBS
@@ -39,18 +38,6 @@ instance ( Crypto.HasHashing c
         <*> field
         <*> field
 
-instance ( CBOR.Serialise (Crypto.PublicKey c)
-         , CBOR.Serialise (Crypto.Signature c)
-         , Typeable c
-         , FromField (BlockHash c)
-         ) => FromRow (RadTx c) where
-    fromRow = Tx
-        <$> field
-        <*> field
-        <*> field
-        <*> field
-        <*> field
-
 -- | Convert a SQL parent hash field to a 'Crypto.Hash'. The SQL value is
 -- 'SQLNull' in the case of the genesis block.
 fromPrevHashField
@@ -63,17 +50,6 @@ fromPrevHashField f =
     case fieldData f of
         SQLNull -> Ok Crypto.zeroHash
         _       -> fromField f
-
-instance ( Crypto.HasHashing c
-         , ToField (Crypto.Hashed c (RadTx c))
-         , ToField (BlockHash c)
-         , CBOR.Serialise (BlockHash c)
-         , CBOR.Serialise (Crypto.PublicKey c)
-         , CBOR.Serialise (Crypto.Signature c)
-         ) => ToRow (RadTx c) where
-    toRow tx@Tx{..} =
-        [ toField (Crypto.hash @c tx), toField txMessage, toField txPubKey
-        , toField txChainId, toField txNonce, toField txContext ]
 
 instance ( CBOR.Serialise (Crypto.Signature c)
          , CBOR.Serialise msg
--- a/src/Oscoin/Storage/Block/SQLite.hs
+++ b/src/Oscoin/Storage/Block/SQLite.hs
@@ -3,6 +3,9 @@
 -- | Disk-backed block storage using SQLite.
 module Oscoin.Storage.Block.SQLite
     ( withBlockStore
+    , StorableTx
+    , IsTxRow(..)
+    , TxRow(..)
     ) where
 
 import           Oscoin.Prelude
@@ -18,10 +21,8 @@ import           Oscoin.Storage.Block.SQLite.Internal
 import           Database.SQLite.Simple (Only(..))
 import qualified Database.SQLite.Simple as Sql
 import           Database.SQLite.Simple.FromField (FromField)
-import           Database.SQLite.Simple.FromRow (FromRow)
 import           Database.SQLite.Simple.QQ
 import           Database.SQLite.Simple.ToField (ToField)
-import           Database.SQLite.Simple.ToRow (ToRow)
 
 import           Codec.Serialise (Serialise)
 
@@ -34,14 +35,10 @@ import           Codec.Serialise (Serialise)
 withBlockStore
     :: ( ToField s
        , FromField s
-       , ToField (Crypto.Hash c)
        , FromField (Crypto.Hash c)
        , Serialise s
-       , Serialise (Crypto.Hash c)
        , Ord (BlockHash c)
-       , ToRow tx
-       , FromRow tx
-       , Crypto.HasHashing c
+       , StorableTx c tx
        )
     => String
     -- ^ The path where the DB will live on disk
@@ -74,12 +71,9 @@ withBlockStore path genesisBlock action =
 lookupBlock
     :: forall c tx s.
        ( Serialise s
-       , Serialise (Crypto.Hash c)
-       , ToField (Crypto.Hash c)
        , FromField (Sealed c s)
        , FromField (Crypto.Hash c)
-       , FromRow tx
-       , Crypto.HasHashing c
+       , StorableTx c tx
        )
     => Handle c tx s
     -> BlockHash c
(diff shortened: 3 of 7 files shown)


src/Oscoin/Storage/Block/SQLite/Transaction.hs
```diff
@@ -0,0 +1,118 @@
...
+    fromRow = do
+        transactionRowMessage <- field
+        transactionRowAuthor <- field
+        transactionRowChainId <- field
+        transactionRowNonce <- field
+        transactionRowContext <- field
+        pure $ TransactionRow{..}
+
+
+class (Hashable c a, ToField (Hash c), Serialise (Hash c)) => StoreTransaction c a where
```
kim: `StoreTransaction` confused me until I scrolled here, due to "Transaction" being overloaded, and I think we use "Tx" everywhere to refer to block chain transactions. How about "StoredTx" resp "TxRow"?

It's also kind of odd that the instance selection is deferred to runtime using the proxy argument - isn't the crypto fixed by the choice of `a`? So a fundep `a -> c`, or associated type should be possible, or am I missing something?

geigerzaehler: I adjusted the names. Regarding the crypto parameter: As it is currently implemented `DummyTx` is independent of the crypto but the tests are polymorphic over the crypto. Hence we need instances for `StorableTx c DummyTx` for all `c` such that `IsCrypto c`. This makes a functional dependency `a -> c` impossible.

kim: Well then why not add a dummy phantom parameter to `DummyTx`?

kim: Let me `sed` this for you:

```diff
diff --git a/src/Oscoin/Storage/Block/SQLite/Transaction.hs b/src/Oscoin/Storage/Block/SQLite/Transaction.hs
index a490cbca..01cfd29d 100644
--- a/src/Oscoin/Storage/Block/SQLite/Transaction.hs
+++ b/src/Oscoin/Storage/Block/SQLite/Transaction.hs
@@ -49,9 +49,9 @@ instance FromRow TxRow where
         pure $ TxRow{..}
 
 
-class (Hashable c a, ToField (Hash c), Serialise (Hash c)) => StorableTx c a where
-    toTxRow :: Proxy c -> a -> TxRow
-    fromTxRow :: Proxy c -> TxRow -> a
+class (Hashable c a, ToField (Hash c), Serialise (Hash c)) => StorableTx c a | a -> c where
+    toTxRow   :: a -> TxRow
+    fromTxRow :: TxRow -> a
 
 instance
     ( HasHashing c
@@ -61,7 +61,7 @@ instance
     , Serialise (Signature c)
     , Serialise msg
     ) => StorableTx c (Tx c msg) where
-    toTxRow _ Tx{..} = TxRow
+    toTxRow Tx{..} = TxRow
         { txRowContext = LBS.toStrict $ serialise txContext
         , txRowMessage = LBS.toStrict $ serialise txMessage
         , txRowAuthor = LBS.toStrict $ serialise txPubKey
@@ -69,7 +69,7 @@ instance
         , txRowNonce = txNonce
         }
 
-    fromTxRow _ TxRow{..} = Tx
+    fromTxRow TxRow{..} = Tx
         { txMessage = deserialise $ LBS.fromStrict txRowMessage
         , txPubKey = deserialise $ LBS.fromStrict txRowAuthor
         , txChainId = txRowChainId
@@ -85,7 +85,7 @@ getTx conn hash' = do
                WHERE hash = ? |] (Only $ serialise hash')
     pure $ case listToMaybe rows of
         Nothing    -> Nothing
-        Just txRow -> Just $ fromTxRow (Proxy @c) txRow
+        Just txRow -> Just $ fromTxRow txRow
 
 
 storeTxs :: forall c tx. (StorableTx c tx) => Sql.Connection -> Hash c -> [tx] -> IO ()
@@ -97,7 +97,7 @@ storeTxs conn blockHash txs =
               VALUES                    (?, ?, ?, ?, ?, ?, ?) |] (map toTxRowValues txs)
   where
     toTxRowValues tx =
-        let TxRow {..} = toTxRow (Proxy @c) tx
+        let TxRow {..} = toTxRow tx
         in
             [ toField txRowMessage
             , toField txRowAuthor
@@ -115,4 +115,4 @@ getBlockTxs conn h = do
         [sql| SELECT message, author, chainid, nonce, context
                 FROM transactions
                WHERE blockhash = ? |] (Only h)
-    pure $ map (fromTxRow $ Proxy @c) txRows
+    pure $ map fromTxRow txRows
diff --git a/test/Oscoin/Test/Consensus.hs b/test/Oscoin/Test/Consensus.hs
index 79d922b4..a75d3336 100644
--- a/test/Oscoin/Test/Consensus.hs
+++ b/test/Oscoin/Test/Consensus.hs
@@ -190,7 +190,7 @@ prettyCounterexample
        , IsCrypto c
        )
     => TestNetwork c s a
-    -> [DummyTx]
+    -> [DummyTx c]
     -> Float
     -> Text
 prettyCounterexample tn@TestNetwork{..} txsReplicated score =
diff --git a/test/Oscoin/Test/Consensus/Nakamoto.hs b/test/Oscoin/Test/Consensus/Nakamoto.hs
index 3787fce3..b5f81b65 100644
--- a/test/Oscoin/Test/Consensus/Nakamoto.hs
+++ b/test/Oscoin/Test/Consensus/Nakamoto.hs
@@ -71,7 +71,7 @@ mineBlockRandom stdGen blk =
            then Just $ sealBlock emptyPoW blk
            else Nothing
 
-type NakamotoNode c = NakamotoT DummyTx (TestNodeT c PoW Identity)
+type NakamotoNode c = NakamotoT (DummyTx c) (TestNodeT c PoW Identity)
 
 data NakamotoNodeState c = NakamotoNodeState
     { nakStdGen :: StdGen
diff --git a/test/Oscoin/Test/Consensus/Network.hs b/test/Oscoin/Test/Consensus/Network.hs
index 619cb07a..81fafdd5 100644
--- a/test/Oscoin/Test/Consensus/Network.hs
+++ b/test/Oscoin/Test/Consensus/Network.hs
@@ -70,7 +70,7 @@ import           Text.Show (Show(..))
 -- TestableNode ----------------------------------------------------------------
 
 class
-    ( MonadMempool c DummyTx m
+    ( MonadMempool c (DummyTx c) m
     , MonadStateStore c DummyState m
     , HasTestNodeState c s a
     , LiftTestNodeT c s m
@@ -78,9 +78,9 @@ class
   where
     testableInit         :: TestNetwork c s b -> TestNetwork c s a
 
-    testableTick         :: Timestamp -> m (Maybe (Block c DummyTx (Sealed c s)))
+    testableTick         :: Timestamp -> m (Maybe (Block c (DummyTx c) (Sealed c s)))
     testableRun          :: a -> m b -> (b, a)
-    testableScore        :: a -> Blockchain c DummyTx s -> Score
+    testableScore        :: a -> Blockchain c (DummyTx c) s -> Score
 
 testableLongestChain :: TestableNode c s m a => a -> [BlockHash c]
 testableLongestChain =
@@ -92,7 +92,7 @@ testableLongestChain =
 testableNodeAddr :: forall c s m a. TestableNode c s m a => a -> DummyNodeId
 testableNodeAddr nodeState = tnsNodeId $ nodeState ^. testNodeStateL @c
 
-testableIncludedTxs :: TestableNode c s m a => a -> [DummyTx]
+testableIncludedTxs :: TestableNode c s m a => a -> [DummyTx c]
 testableIncludedTxs =
       concatMap (toList . blockData)
     . blocks
@@ -101,7 +101,7 @@ testableIncludedTxs =
 testableShow :: (HasHashing c, TestableNode c s m a) => a -> Text
 testableShow = showChainDigest . testableBestChain
 
-testableBestChain :: TestableNode c s m a => a -> Blockchain c DummyTx s
+testableBestChain :: TestableNode c s m a => a -> Blockchain c (DummyTx c) s
 testableBestChain nodeState =
     let blockStore = nodeState ^. testNodeStateL . tnsBlockstoreL
      in unsafeToBlockchain $ Pure.getBlocks 10000 blockStore
@@ -110,13 +110,13 @@ testableBestChain nodeState =
 -- Msg ------------------------------------------------------------------------
 
 data Msg c s =
-      BlockMsg    (Block c DummyTx (Sealed c s))
+      BlockMsg    (Block c (DummyTx c) (Sealed c s))
     -- ^ Message containing a 'Block'.
-    | TxMsg       DummyTx
+    | TxMsg       (DummyTx c)
     -- ^ Message containing a @tx@.
     | ReqBlockMsg DummyNodeId (BlockHash c)
     -- ^ Message requesting a 'Block' with a certain hash. 'DummyNodeId' is the requesting node.
-    | ResBlockMsg DummyNodeId (Block c DummyTx (Sealed c s))
+    | ResBlockMsg DummyNodeId (Block c (DummyTx c) (Sealed c s))
     -- ^ Message sent in response to 'ReqBlockMsg'. 'DummyNodeId' is the requesting node.
 
 deriving instance (Eq s, Eq (Hash c)) => Eq (Msg c s)
@@ -146,7 +146,7 @@ data TestNetwork c s a = TestNetwork
     -- message we take the first element to determine when the message
     -- gets delivered.
     , tnRng        :: StdGen
-    , tnEval       :: DummyEval
+    , tnEval       :: DummyEval c
     , tnValidate   :: DummyValidate c s
     , tnMsgCount   :: Int
     -- ^ Number of messages delivered in the network
@@ -154,11 +154,11 @@ data TestNetwork c s a = TestNetwork
     }
 
 -- | Test evaluator.
-type DummyEval = Evaluator DummyState DummyTx ()
+type DummyEval c = Evaluator DummyState (DummyTx c) ()
 
 -- | Test block validator.
 type DummyValidate c s =
-    (Block c DummyTx (Sealed c s) -> Either (ValidationError c) ())
+    (Block c (DummyTx c) (Sealed c s) -> Either (ValidationError c) ())
 
 -- | Network partitions is a map of node id to a set of node ids _not_
 -- reachable from that node id.
@@ -256,7 +256,7 @@ applyMessage
     => Consensus.Config
     -> DummyNodeId
     -> DummyValidate c s
-    -> DummyEval
+    -> DummyEval c
     -> Msg c s
     -> m [Msg c s]
 applyMessage _ _ _ _ msg@(TxMsg tx) = do
diff --git a/test/Oscoin/Test/Consensus/Node.hs b/test/Oscoin/Test/Consensus/Node.hs
index 4a859a1a..45095df6 100644
--- a/test/Oscoin/Test/Consensus/Node.hs
+++ b/test/Oscoin/Test/Consensus/Node.hs
@@ -68,7 +68,7 @@ withTestBlockStore
        , Ord (BlockHash c)
        , Hashable c Word8
        )
-    => (BlockStore c DummyTx s n -> n b)
+    => (BlockStore c (DummyTx c) s n -> n b)
     -> n b
 withTestBlockStore action =
     let publicAPI = BlockStoreReader
@@ -105,11 +105,11 @@ stepTestProtocol
        , Ord s
        , Buildable (Hash c)
        )
-    => BlockStore c DummyTx s n
-    -> n (Maybe (Block c DummyTx (Sealed c s)))
+    => BlockStore c (DummyTx c) s n
+    -> n (Maybe (Block c (DummyTx c) (Sealed c s)))
     -- ^ An action to fetch the next block, if any.
-    -> Validate c DummyTx s
-    -> (Block c DummyTx (Sealed c s) -> Score)
+    -> Validate c (DummyTx c) s
+    -> (Block c (DummyTx c) (Sealed c s) -> Score)
     -> n ()
 stepTestProtocol fullBlockStore fetchNextBlock validateFull scoreBlock = do
    -- The Consensus config is hardcoded for now, but it's fairly easy
@@ -142,11 +142,11 @@ instance Arbitrary DummyNodeId where
 
 data TestNodeState c s = TestNodeState
     { tnsNodeId       :: DummyNodeId
-    , tnsMempool      :: Mempool.Mempool c DummyTx
+    , tnsMempool      :: Mempool.Mempool c (DummyTx c)
     , tnsStateStore   :: StateStore.StateStore c DummyState
-    , tnsBlockstore   :: BlockStore.Pure.Handle c DummyTx s
-    , tnsOrphanage    :: Orphanage c DummyTx s
-    , tnsReceiptStore :: ReceiptStore.Store c DummyTx DummyOutput
+    , tnsBlockstore   :: BlockStore.Pure.Handle c (DummyTx c) s
+    , tnsOrphanage    :: Orphanage c (DummyTx c) s
+    , tnsReceiptStore :: ReceiptStore.Store c (DummyTx c) DummyOutput
     }
 
 deriving instance Show (Hash c) => Show (TestNodeState c s)
@@ -154,18 +154,18 @@ deriving instance Show (Hash c) => Show (TestNodeState c s)
 class HasTestNodeState c s a | a -> s where
     testNodeStateL :: Lens' a (TestNodeState c s)
 
-instance ReceiptStore.HasStore c DummyTx DummyOutput (TestNodeState c s) where
+instance ReceiptStore.HasStore c (DummyTx c) DummyOutput (TestNodeState c s) where
     storeL = lens tnsReceiptStore (\s tnsReceiptStore -> s { tnsReceiptStore })
 
-tnsMempoolL :: Lens' (TestNodeState c s) (Mempool.Mempool c DummyTx)
+tnsMempoolL :: Lens' (TestNodeState c s) (Mempool.Mempool c (DummyTx c))
 tnsMempoolL = lens tnsMempool (\s a -> s { tnsMempool = a })
 
-tnsBlockstoreL :: Lens' (TestNodeState c s) (BlockStore.Pure.Handle c DummyTx s)
+tnsBlockstoreL :: Lens' (TestNodeState c s) (BlockStore.Pure.Handle c (DummyTx c) s)
 tnsBlockstoreL = lens tnsBlockstore (\s a -> s { tnsBlockstore = a })
 
 emptyTestNodeState
     :: IsCrypto c
-    => (Block c DummyTx Unsealed -> Block c DummyTx (Sealed c s))
+    => (Block c (DummyTx c) Unsealed -> Block c (DummyTx c) (Sealed c s))
     -> DummyNodeId
     -> TestNodeState c s
 emptyTestNodeState seal nid = TestNodeState
@@ -191,7 +191,7 @@ newtype TestNodeT c s m a = TestNodeT (StateT (TestNodeState c s) m a)
 instance ( Ord (Hash c)
          , Hashable c Word8
          , Monad m
-         ) => MonadMempool c DummyTx (TestNodeT c s m) where
+         ) => MonadMempool c (DummyTx c) (TestNodeT c s m) where
     addTxs txs = modify' (over tnsMempoolL (Mempool.insertMany txs))
     getTxs     = Mempool.toList <$> gets tnsMempool
     delTxs txs = modify' (over tnsMempoolL (Mempool.removeTxs txs))
@@ -206,7 +206,7 @@ instance ( Ord (Hash c)
 
 instance ( Ord (Hash c)
          , Monad m
-         ) => MonadReceiptStore c DummyTx DummyOutput (TestNodeT c s m) where
+         ) => MonadReceiptStore c (DummyTx c) DummyOutput (TestNodeT c s m) where
     addReceipt = ReceiptStore.addWithStore
     lookupReceipt = ReceiptStore.lookupWithStore
 
diff --git a/test/Oscoin/Test/Consensus/Simple.hs b/test/Oscoin/Test/Consensus/Simple.hs
index 16a9e9a5..132921a4 100644
--- a/test/Oscoin/Test/Consensus/Simple.hs
+++ b/test/Oscoin/Test/Consensus/Simple.hs
@@ -60,7 +60,7 @@ runSimpleT env lt (SimpleT ma) = runStateT (runReaderT ma env) lt
 
 -- Simple Node -----------------------------------------------------------------
 
-type SimpleNode c = SimpleT DummyTx DummyNodeId (TestNodeT c PoA Identity)
+type SimpleNode c = SimpleT (DummyTx c) DummyNodeId (TestNodeT c PoA Identity)
 
 data SimpleNodeState c = SimpleNodeState
     { snsPosition :: (Int, Int)
diff --git a/test/Oscoin/Test/Storage/Block/Equivalence.hs b/test/Oscoin/Test/Storage/Block/Equivalence.hs
index 931c35c3..57da0307 100644
--- a/test/Oscoin/Test/Storage/Block/Equivalence.hs
+++ b/test/Oscoin/Test/Storage/Block/Equivalence.hs
@@ -100,7 +100,7 @@ type StoresUnderTest c tx s m =
 
 -- | The monad whe stores runs in, which keeps around an 'Orphanage' to be
 -- used internally for the SQLite store implementation (cfr. 'withStores').
-type StoreM c = StateT (Orphanage c DummyTx DummySeal) IO
+type StoreM c = StateT (Orphanage c (DummyTx c) DummySeal) IO
 
 -- | Initialises both the SQL and the STM store and pass them to the 'action'.
 -- For the SQL store, we need to cheat: due to the fact we /do not/ want to
@@ -110,8 +110,8 @@ type StoreM c = StateT (Orphanage c DummyTx DummySeal) IO
 withStores
     :: forall c a.
        IsCrypto c
-    => Orphanage c DummyTx DummySeal
-    -> (StoresUnderTest c DummyTx DummySeal (StoreM c) -> StoreM c a)
+    => Orphanage c (DummyTx c) DummySeal
+    -> (StoresUnderTest c (DummyTx c) DummySeal (StoreM c) -> StoreM c a)
     -> IO a
 withStores orphanage action =
     SQLite.withBlockStore ":memory:" defaultGenesis $ \sqlStore ->
@@ -119,15 +119,15 @@ withStores orphanage action =
             evalStateT (action (wrapProto (hoistStore sqlStore), hoistStore stmStore)) orphanage
   where
       hoistStore
-          :: BlockStore c DummyTx DummySeal IO
-          -> BlockStore c DummyTx DummySeal (StoreM c)
+          :: BlockStore c (DummyTx c) DummySeal IO
+          -> BlockStore c (DummyTx c) DummySeal (StoreM c)
       hoistStore (public, private) =
           (hoistBlockStoreReader liftIO public, hoistBlockStoreWriter liftIO private)
 
       -- Overrides the 'insertBlock' to pass via chain selection.
       wrapProto
-          :: BlockStore c DummyTx DummySeal (StoreM c)
-          -> BlockStore c DummyTx DummySeal (StoreM c)
+          :: BlockStore c (DummyTx c) DummySeal (StoreM c)
+          -> BlockStore c (DummyTx c) DummySeal (StoreM c)
       wrapProto bs@(public, private) =
           let cfg = Consensus.Config 1024 2016
               noValidation _ _ = Right ()
diff --git a/test/Test/Oscoin/DummyLedger.hs b/test/Test/Oscoin/DummyLedger.hs
index af0464cf..ac0ce215 100644
--- a/test/Test/Oscoin/DummyLedger.hs
+++ b/test/Test/Oscoin/DummyLedger.hs
@@ -21,17 +21,17 @@ import qualified Database.SQLite.Simple.ToField as SQLite
 
 import           Test.QuickCheck
 
-newtype DummyTx = DummyTx ByteString
+newtype DummyTx c = DummyTx ByteString
     deriving (Eq, Ord, Show, Serialise)
 
-instance Arbitrary DummyTx where
+instance Arbitrary (DummyTx c) where
     arbitrary = DummyTx . BS.pack <$> vectorOf 32 arbitrary
 
-instance (HasHashing c) => Hashable c DummyTx where
+instance (HasHashing c) => Hashable c (DummyTx c) where
     hash (DummyTx payload) = Crypto.toHashed $ Crypto.hashByteArray payload
 
-instance (SQLite.ToField (Hash c), Serialise (Hash c), HasHashing c) => StorableTx c DummyTx where
-    toTxRow _proxy (DummyTx msg) = TxRow
+instance (SQLite.ToField (Hash c), Serialise (Hash c), HasHashing c) => StorableTx c (DummyTx c) where
+    toTxRow (DummyTx msg) = TxRow
         { txRowContext = mempty
         , txRowMessage = msg
         , txRowAuthor  = mempty
@@ -39,12 +39,12 @@ instance (SQLite.ToField (Hash c), Serialise (Hash c), HasHashing c) => Storable
         , txRowNonce   = 0
         }
 
-    fromTxRow _proxy TxRow{txRowMessage} =
+    fromTxRow TxRow{txRowMessage} =
         DummyTx txRowMessage
 
 type DummyState = ()
 
 type DummyOutput = ()
 
-dummyEval :: Evaluator DummyState DummyTx DummyOutput
+dummyEval :: Evaluator DummyState (DummyTx c) DummyOutput
 dummyEval _ s = Right ((), s)
```

geigerzaehler: This would work, too. Thanks.

I’m wondering: Is there a compelling reason to have that phantom on `DummyTx`?

kim: > Is there a compelling reason to have that phantom on DummyTx?

Because the typeclass instance is otherwise selected at runtime, based on the `Proxy` value. If the typeclass is some temporary measure, I don't care much, but it doesn't look like it.

kim: Also, I don't think the details of how tests are implemented should determine how the abstractions in the application should look like.

Looking closer, I realise that there are other problems with this code: the typeclass establishes an isomorphism between `TxRow` and `a`, where `TxRow` is a tagged representation of the _serialised_ fields of `a`. However, that is a lie: `deserialise` diverges if the stored value is not what the current version of the code expects. Since we are quite sloppy and just derive our `Serialise` instances generically in most places, someone just has to change some field, and tada: `_|_`.

Also, there is actually no reason to mention `c` in the class: the instance of `Tx` can produce the  `hash` field and store that in `TxRow` (which then maps exactly to the database row, which I thought was the whole purpose). The class hence becomes:

```haskell
class StorableTx a where
    toTxRow :: a -> TxRow
    fromTxRow :: TxRow -> Either DeserialiseFailure a
```

with straightforward instances. Instead of `DeserialiseFailure` the left should probably some wrapper type.

geigerzaehler:
@kim, I made the following changes to address your comments.

The `StorableTx` type class has been renamed to `IsTxRow` and only carries the transaction type parameter. To simplify the use of all the storage and lookup functions I introduced the `StorableTx c tx` constraint alias that includes the constraint for serialising crypto hashes.

I also changed the type of `fromTxRow` to return an `Either`.

PR #436, opened 2019-03-27
Oscoin Tx and Core types - Part 1.

cloudhead:
This is a first pass on the high level domain logic of the ledger. Since there's a lot of code here, I decided to create the basic structure and implement only a few of the transactions, leaving the rest unimplemented.

Merging this should not affect any of the existing code, as it is fully isolated. The plan is to switch the current `Tx` from `Data/Tx.hs` to this when we're ready.

Once we have a better idea of how this will integrate with the existing codebase, we'll have to come up with a testing strategy, but before that there might be some significant or not so significant changes to the data-types and functions required.


src/Oscoin/Data/OscoinTx.hs
```diff
@@ -0,0 +1,325 @@
...
+        -- Account doesn't exist. Create it.
+        Nothing -> Just . AccountVal $ (mkAccount b) { accountBalance = bal }
+
+        -- Not an account ...
+        val -> val
+
+-- | Map a handler result into a transaction result.
+handler :: Either Error a -> Either (TxError c) a
+handler (Left err) = Left (ErrHandlerFailed err)
+handler (Right r)  = Right r
```
cloudhead: Anyone have a better name for this?

src/Oscoin/Data/Core.hs
```diff
@@ -0,0 +1,496 @@
...
+    | ProjectVal (Project c)
+    | MemberVal  (Member c)
+    | NatVal     Natural
+
+-- | Like 'Map.adjust', but for 'WorldState'.
+adjust :: (StateVal c -> StateVal c) -> StateKey -> WorldState c -> WorldState c
+adjust f k ws =
+    case WorldState.lookup k ws of
+        Just v  -> WorldState.insert k (f v) ws
+        Nothing -> ws
```
cloudhead: Eventually these functions should be implemented generically in `avl-auth`

cloudhead:
There are some important questions not answered by this PR on purpose, since they involve larger discussions. This includes:

* Key space, eg. how exactly are state objects mapped in the world state to keys - currently objects are bundled, ie. the `Project` type is one big object with everything relevant to the project, which is optimized for ease of programming, but not optimized for all access patterns.
* Serialization: how are things serialized when sent through the wire - domain objects are currently not serializable.

src/Oscoin/Data/OscoinTx.hs
```diff
@@ -0,0 +1,322 @@
...
+
+-- | Lookup a member in the state.
+lookupMember
+    :: (ByteArrayAccess (Crypto.ShortHash c), CBOR.Serialise (Crypto.ShortHash c))
+    => Address c -> WorldState c -> Either (TxError c) (Member c)
+lookupMember addr ws =
+    case WorldState.lookup (addressKey addr) ws of
+        Just (MemberVal m) -> Right m
+        _                  -> Left (ErrAccountNotFound addr)
```
kim: It is a bit unsatisfying that the type of the key doesn't correspond to the value - one can store anything under an `Address`, and later try to get back a different thing using the same key. While this can be solved using singleton types, wouldn't it be more informative in the meantime to map the absence of a key to a single `ErrNotFound`, and signal the case of a type mismatch as a separate error constructor?

src/Oscoin/Data/OscoinTx.hs
```diff
@@ -0,0 +1,322 @@
+-- | The Transaction.
```
kim: nit: module haddock typically goes _after_ pragmas

src/Oscoin/Data/Core.hs
```diff
@@ -0,0 +1,502 @@
+-- | Core ledger data types.
+--
+{-# LANGUAGE UndecidableInstances #-}
+{-# OPTIONS_GHC -fno-warn-redundant-constraints #-}
+{-# OPTIONS_GHC -fno-warn-deprecations #-}
```
kim: Why is this?

cloudhead: Because of the `notImplemented`s :unamused:

src/Oscoin/Data/Core.hs
```diff
@@ -0,0 +1,502 @@
...
+-------------------------------------------------------------------------------
+
+type ReceiveReward' c =
+       HasDigitalSignature c
+    => Balance                    -- ^ The balance being rewarded
+    -> Epoch                      -- ^ The current epoch
+    -> Project c                  -- ^ The project dictionary
+    -> [(Address c, Balance)]     -- ^ A balance distribution
+
+-- | By default, burn the reward.
```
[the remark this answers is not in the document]
cloudhead: Hehe - in practice projects will have the option to choose a different contract before the project receives oscoin, but burning the reward is a nice way to support the network as a whole by reducing the total supply.

src/Oscoin/Data/OscoinTx.hs
```diff
@@ -0,0 +1,322 @@
...
+import           Data.ByteArray (ByteArrayAccess)
+import qualified Data.Map as Map
+
+-------------------------------------------------------------------------------
+-- Tx
+-------------------------------------------------------------------------------
+
+-- | A transaction on the ledger
+data Tx c = Tx
+    { txVersion :: Word16
```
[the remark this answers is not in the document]
cloudhead: :eyes: like this: 24e2048 ?

src/Oscoin/Data/Core.hs
geigerzaehler: This module does not use `notImplemented`. Can we enable the warning and the warning about redundant constraints?

src/Oscoin/Data/Core.hs
```diff
@@ -0,0 +1,502 @@
+-- | Core ledger data types.
+--
+{-# LANGUAGE UndecidableInstances #-}
+{-# OPTIONS_GHC -fno-warn-redundant-constraints #-}
+{-# OPTIONS_GHC -fno-warn-deprecations #-}
+module Oscoin.Data.Core where
```
geigerzaehler: This is a very generic module name. What do think about `Oscoin.World` or `Oscoin.Ledger`

geigerzaehler: Can we also have explicit exports to distinguish between internal and public methods?

src/Oscoin/Data/OscoinTx.hs
cloudhead: Good idea - something like this 7bd9b4f ? The error could be improved in the future to include expected/actual..

src/Oscoin/Data/Core.hs
```diff
@@ -0,0 +1,502 @@
...
+-------------------------------------------------------------------------------
+
+-- | A project checkpoint.
+data Checkpoint c = Checkpoint
+    { checkpointNumber        :: Natural
+    -- ^ The checkpoint number, starting from zero.
+    , checkpointStateHash     :: Crypto.Hash c
+    -- ^ The hash of the project state at this checkpoint.
+    , checkpointContributions :: [Contribution c]
+    -- ^ The new contributions since the last checkpoint.
```
geigerzaehler: Is the list required to be hash-linked? Meaning does the `contribParentHash` of an item need to match the `contribHash` of the parent?

src/Oscoin/Data/Core.hs
```diff
@@ -0,0 +1,502 @@
...
+    a == b = checkpointNumber a == checkpointNumber b
+
+instance Ord (Checkpoint c) where
+    a <= b = checkpointNumber a <= checkpointNumber b
+
+-- | A contribution to a project.
+data Contribution c = Contribution
+    { contribHash       :: Crypto.Hash c
+    -- ^ The hash of the off-chain contribution artefact.
+    , contribParentHash :: Maybe (Crypto.Hash c)
```
geigerzaehler: What do you think about introducing `data ContributionArtifact` and changing this to `Crypto.Hashed ContriutionArtifact c`?

src/Oscoin/Data/Core.hs
```diff
@@ -0,0 +1,502 @@
...
+    , pReceiveReward   :: ReceiveReward' c
+    , pCheckpoint      :: Checkpoint' c
+    , pUnregister      :: Unregister' c
+    , pAuthorize       :: Authorize' c
+    , pDeauthorize     :: Deauthorize' c
+    , pUpdateContract  :: UpdateContract' c
+    }
+
+defaultProject :: Ord (Crypto.Hash c) => Address c -> Project c
+defaultProject addr = Project
```
geigerzaehler: How about calling this `mkProject` or `emptyProject` to align with `mkAccount`?

src/Oscoin/Data/Core.hs
```diff
@@ -0,0 +1,502 @@
...
+    , pUpdateContract   = defaultUpdateContract
+    }
+
+-- | Get the address of a project.
+projectAddr :: Project c -> Address c
+projectAddr = accountAddr . pAccount
+
+-- | A delegation of oscoin between two accounts.
+data Delegation c = Delegation
+    { dDelegator  :: Address c
```
geigerzaehler: I fear that using one letter as the field prefix quickly lead to clashes if we introduce another data type starting with “D”.

src/Oscoin/Data/Core.hs
cloudhead: I can remove the redundant constraints suppression now though!

src/Oscoin/Data/Core.hs
```diff
@@ -0,0 +1,502 @@
...
+distribute :: Balance -> Set (Address c) -> ([(Address c, Balance)], Balance)
+distribute bal accs =
+    ([(acc, share) | acc <- Set.toList accs], remainder)
+  where
+    share     = bal `div` fromIntegral (Set.size accs)
+    remainder = bal `mod` share
+
+-- | Return 'Right ()' if the member is a project maintainer.
+allowAnyMaintainer :: Ord (Crypto.Hash c) => Project c -> Member c -> Either Error ()
+allowAnyMaintainer Project{..} Member{..} =
```
geigerzaehler: What do you think about calling this `ensureMaintainer` or `requireMaintainer`?

src/Oscoin/Data/Core.hs
```diff
@@ -0,0 +1,502 @@
...
+    , accountNonce   = 0
+    }
+
+-- | Increase the account balance.
+deposit :: Balance -> Account c -> Account c
+deposit bal acc = acc { accountBalance = accountBalance acc + bal }
+
+-- | Decrease the account balance.
+withdraw :: Balance -> Account c -> Account c
+withdraw bal acc = acc { accountBalance = accountBalance acc - bal }
```
geigerzaehler: We need to check against underflow here.

src/Oscoin/Data/Core.hs
```diff
@@ -0,0 +1,502 @@
...
+mkAccount :: Address c -> Account c
+mkAccount addr = Account
+    { accountAddr    = addr
+    , accountBalance = 0
+    , accountNonce   = 0
+    }
+
+-- | Increase the account balance.
+deposit :: Balance -> Account c -> Account c
+deposit bal acc = acc { accountBalance = accountBalance acc + bal }
```
geigerzaehler: We need to check against overflow.

src/Oscoin/Data/OscoinTx.hs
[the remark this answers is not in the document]
geigerzaehler: Unfortunately this does not ensure that the value level values match the type level values

src/Oscoin/Data/OscoinTx.hs
```diff
@@ -0,0 +1,333 @@
...
+    ( Show (Signature c)
+    , Show (BlockHash c)
+    , Crypto.HasHashing c
+    , Show (PublicKey c)
+    ) => Show (TxPayload c)
+
+deriving instance
+    ( Eq (Signature c)
+    , Eq (BlockHash c)
+    , Eq (PublicKey c)
```
geigerzaehler: Maybe it would be a good idea to add these constraints to the requirements for the `HasDigitalSignature` class like we did for `Hash`.

src/Oscoin/Data/Core.hs
```diff
@@ -0,0 +1,495 @@
...
+data ParamVal =
+      Integer  Int
+    -- ^ Expresses positive or negative whole numbers.
+    | Rational Int Int
+    -- ^ Expresses ratios, eg. 1/2.
+    deriving (Show, Eq, Ord)
+
+-------------------------------------------------------------------------------
+
+newtype Error = Error Text
```
geigerzaehler: I think this name is too generic, especially if this module is imported unqualified. What do you think about `ContractError` or `HandlerError`

src/Oscoin/Data/OscoinTx.hs
```diff
@@ -0,0 +1,333 @@
...
+-- Tx
+-------------------------------------------------------------------------------
+
+-- | A transaction on the ledger
+data Tx' (version :: Nat) (flags :: Nat) c = Tx'
+    { txVersion :: Word16
+    -- ^ Transaction version. Must be @1@.
+    , txFlags   :: Word16
+    -- ^ Feature flags. Must be @0@.
+    , txPayload :: TxPayload c
```
geigerzaehler: `txMessage`?

src/Oscoin/Data/OscoinTx.hs
geigerzaehler: `mapHandlerError`?

src/Oscoin/Data/OscoinTx.hs
```diff
@@ -0,0 +1,333 @@
...
+       , Ord (Crypto.Hash c)
+       )
+    => Tx c
+    -- ^ The transaction to apply.
+    -> Address c
+    -- ^ The author or \"signer\" of the transaction.
+    -> WorldState c
+    -- ^ The input world state.
+    -> Either (TxError c) (WorldState c, TxOutput)
+applyTx tx@Tx'{..} author ws = do
```
geigerzaehler: I think we should at least have tests for this function.

src/Oscoin/Data/Core.hs
cloudhead: Yes, this ensures contributions haven't been omitted and makes the ledger more useful.

src/Oscoin/Data/Core.hs
cloudhead: Indeed, maybe `Oscoin.Data.Ledger`?

src/Oscoin/Data/Core.hs
cloudhead: Oh drat, you're right, it's only the other one which does!

src/Oscoin/Data/Core.hs
cloudhead: :thinking: I like `requireMaintainer`

src/Oscoin/Data/OscoinTx.hs
cloudhead: It used to be that, but I've seen transactions as a whole being called "messages", so not sure :thinking:

src/Oscoin/Data/OscoinTx.hs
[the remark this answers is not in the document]
kim: Yeah I meant removing the value-level fields and demoting the `Nat`s where needed (`natVal`)

src/Oscoin/Data/Core.hs
geigerzaehler: Could you add that to the field documentation?

src/Oscoin/Data/Core.hs
cloudhead: 2f5fc95

src/Oscoin/Data/OscoinTx.hs
geigerzaehler: I’m also fine with `txPayload`.

src/Oscoin/Data/OscoinTx.hs
cloudhead: :+1:

src/Oscoin/Data/Core.hs
cloudhead: Hmm I thought about this, but since these will come from the network, we won't have a `ContributionArtifact` to instantiate the type - what do you think?

src/Oscoin/Data/OscoinTx.hs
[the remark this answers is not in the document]
cloudhead: Hmm, will that make it more clunky to decode/encode? Guess we'll find out :crossed_fingers:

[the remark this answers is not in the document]
cloudhead: How's this: 6293ce3

src/Oscoin/Data/Core.hs
geigerzaehler: Come to think of it the tag will likely be only used here and we’ll unwrap it anyways. So let’s leave it like it is.

src/Oscoin/Data/OscoinTx.hs
```diff
@@ -0,0 +1,331 @@
...
+    deposit  n acc = acc { accountBalance = accountBalance acc + n }
+    withdraw n acc = acc { accountBalance = accountBalance acc - n }
+
+    withAccount f v = case v of
+        -- Account exists. Update balance or delete account if zero balance.
+        Just (AccountVal acc) -> case f acc of
+            acc'@Account{..} | accountBalance == 0 -> Nothing
+                             | otherwise           -> Just (AccountVal acc')
+        -- Account doesn't exist. Create it.
+        Nothing -> Just . AccountVal $ (mkAccount receiver) { accountBalance = bal }
```
geigerzaehler: It confuses me that in this case `f` is not called but `bal` is used. What do think of the following
~~~haskell
withAccount f v = case v of
    Just (AccountVal acc) -> squashAccountVal (f acc)
    Nothing -> squashAccountVal (f (mkAccount receiver))
    val -> val
squashAccountVal account =
    if accountBalance account == 0
    then Nothing
    else Just (AccountVal account)

src/Oscoin/Data/OscoinTx.hs
```diff
@@ -0,0 +1,331 @@
...
+
+        -- Not an account ...
+        val -> val
```
geigerzaehler: This should probably be an error

src/Oscoin/Data/OscoinTx.hs
```diff
@@ -0,0 +1,331 @@
...
+    Left (ErrInvalidTransfer bal)
+applyTxPayload (TxTransfer sender receiver bal) author ws = do
+    senderAcc   <- lookupAccount sender   ws
+    receiverAcc <- lookupAccount receiver ws
+
+    when (author /= sender) $
+        Left (ErrNotAuthorized author)
+
+    when (bal > accountBalance senderAcc) $
+        Left (ErrInsufficientBalance (accountBalance senderAcc))
```
geigerzaehler: What do you think about throwing these errors as part of the `withdraw` function. Similar throwing `ErrEverflow` as part of `deposit`.

src/Oscoin/Data/OscoinTx.hs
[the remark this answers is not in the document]
geigerzaehler: @kim Can you elaborate on the benefit of having the version and the flags at the type level?

src/Oscoin/Data/OscoinTx.hs
cloudhead: Not a bad idea, but not sure it will make the code more readable vs. doing validation upfront, as it will complicate the logic. Happy to explore in Part 2. of this series though.

src/Oscoin/Data/OscoinTx.hs
cloudhead: I'm going to return `undefined` here, as this branch should never be reached (we lookup the accounts prior) and that's the idiomatic way to signal that with Protolude. In the next PR we can look into returning an Either from this function, and whether that will make things cleaner.

src/Oscoin/Data/OscoinTx.hs
cloudhead: Will also punt on this, if we're thinking of returning an `Either` from withAccount, as it will mean a larger change.

src/Oscoin/Data/OscoinTx.hs
cloudhead: To do this properly will be a lot of code, which I'd rather handle in one of the subsequent PRs, since this code is not currently in use. We'll need generators for all payloads for both valid and invalid transactions, etc.

src/Oscoin/Data/OscoinTx.hs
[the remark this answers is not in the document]
cloudhead: I'm going to leave it as-is for now, with the type-level fields, we can think about it more when we implement encoding/decoding and make a final call then. At the very least, I don't think it "costs" us anything to have this at the type level so far.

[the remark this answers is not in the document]
kim: @geigerzaehler The fields are always constants, as the comments indicated. The constructor is, however, exported, so anyone can set them to whatever value they please. By lifting those to the type-level, there is a nice upgrade path should the datatype evolve in the future. For example, by turning it into a GADT:

```haskell
data Tx v flags where
    TxV1 :: {- ... -} -> Tx 1 0
    TxV2 :: {- ... -} -> Tx 2 flags -- oh, we can allow setting flags here!
```

PR #437, opened 2019-03-28
Do not run protocol for simple block storage tests

geigerzaehler:
We do not run the protocol for the `getTip . insertBlock equivalance` test. The property under test is independent of running the protocol. This also reduces the time to run the test by a whopping 65%.

### DIFF
--- a/test/Oscoin/Test/Storage/Block/Equivalence.hs
+++ b/test/Oscoin/Test/Storage/Block/Equivalence.hs
@@ -56,10 +56,9 @@ classifyChain chain =
 -- and get the tip, and that both stores agree on the result.
 propInsertGetTipEquivalence :: forall c.  Dict (IsCrypto c) -> Property
 propInsertGetTipEquivalence Dict =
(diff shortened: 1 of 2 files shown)


test/Oscoin/Test/Storage/Block/Equivalence.hs
```diff
@@ -102,28 +101,33 @@ type StoresUnderTest c tx s m =
...
 -- SQLite store to pass through chain selection.
 withStores
+    :: forall c a.
+       IsCrypto c
+    => (StoresUnderTest c DummyTx DummySeal IO -> IO a)
+    -> IO a
+withStores action =
+    SQLite.withBlockStore ":memory:" defaultGenesis $ \sqlStore ->
+        STM.withBlockStore (fromGenesis defaultGenesis) blockScore $ \stmStore ->
+            action (sqlStore, stmStore)
```
adinapoli-mndc: It's a shame these two functions are ever-so-slightly different. I think trying to refactor this is going to be slightly tricky due to the callback-style nature of this function, and maybe is not worth it 🤔

PR #443, opened 2019-04-01
Use the threaded runtime for the `oscoin` executable

kim:


package.yaml
```diff
@@ -102,6 +102,7 @@ executables:
       - yaml
     ghc-options:
       - -with-rtsopts=-T
+      - -threaded
```
adinapoli-mndc: Was wondering why it wasn't turned on by default!

PR #445, opened 2019-04-01
Extract crypto tests from 'Oscoin.Tests'

geigerzaehler:
Move crypto tests from `Oscoin.Tests` to new `Test.Oscoin.Crypto.Hash` and `Test.Oscoin.Crypto.PubKey`.

### DIFF
--- a/test/Test/Oscoin/Crypto/PubKey.hs
+++ b/test/Test/Oscoin/Crypto/PubKey.hs
@@ -15,7 +15,6 @@ import           Oscoin.Test.Util (condensed)
 import           Test.QuickCheck.Extended
 import           Test.QuickCheck.Instances ()
 import           Test.Tasty
-import           Test.Tasty.HUnit.Extended
 import           Test.Tasty.QuickCheck hiding ((===))
 
 import qualified Codec.Serialise as CBOR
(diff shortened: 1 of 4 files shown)


geigerzaehler:
Could you take another look @adinapoli-mndc? I pushed one more commit.

> In particular, most of the moved tests seems to be dealing with some form of serialisation roundtripping, so I would be tempted to have these into a separate module (i.e. `Test.Oscoin.Serialisation.*`).
> 
> However, I think the above would violate your newly-proposed schema, where test modules should always be associated with a (library) module under test, so technically you did put them where they belong. I suppose it will get a bit to get used to this schema, as these cross-cutting tests feels a bit like a special snowflake 😁

Yes, the split is obviously not ideal. However the code for the instances is in the corresponding modules. So this change makes it easier to find the tests.

test/Test/Oscoin/Crypto/PubKey.hs
```diff
@@ -0,0 +1,59 @@
...
+tests d = testGroup "Oscoin.Crypto.PubKey"
+    [ testProperty "prop_jsonRoundtripSigned" (prop_jsonRoundtripSigned d)
+    , testProperty "prop_serialiseRoundtripSigned" (prop_serialiseRoundtripSigned d)
+    , testProperty "prop_jsonRoundtripPublicKey" (prop_jsonRoundtripPublicKey d)
+    , testProperty "prop_serialiseRoundtripPublicKey" (prop_serialiseRoundtripPublicKey d)
+    , testProperty "prop_hashSigned" (prop_hashSigned d)
+    ]
+
+prop_jsonRoundtripSigned :: forall c. Dict (IsCrypto c) -> Property
+prop_jsonRoundtripSigned Dict =
```
cloudhead: If we're not using auto-discovery-template-magic, do we need the funky prefix vs. standard camel-case?

geigerzaehler: No, we don’t really need this.This is just for consistency with the tests from the P2P layer.

kim: Well, I think we _want_ to spare us the typing down the road (hoping to get rid of `tasty` eventually). Module-local TH magic (like this: https://hackage.haskell.org/package/hedgehog-0.6/docs/Hedgehog.html#v:discover) is pretty acceptable for this imho, whereas the compiler plugin approach est un peu beaucoup trop.

PR #448, opened 2019-04-01
Reject a peer presenting one's own public key in the handshake.

kim:
Note that this doesn't prevent duplicate keys in the network.

### DIFF
--- a/src/Oscoin/P2P/Handshake/Simple.hs
+++ b/src/Oscoin/P2P/Handshake/Simple.hs
@@ -5,7 +5,11 @@ module Oscoin.P2P.Handshake.Simple
     -- * Combinators
     , signedPayloads
     , whitelist
+
+    , guardResult
+    , peerIdMatch
     , guardPeerId
+    , guardPeerIdNot
     ) where
 
 import           Oscoin.Prelude hiding (length, pi)
@@ -35,6 +39,7 @@ data SimpleError =
     | InvalidPayload
     | Unauthorized
     | IdMismatch
+    | DuplicateId
     | Timeout
     | Gone
     deriving (Eq, Show, Generic)
@@ -67,7 +72,7 @@ keyExchange (myPK, mySK) _role peerId = do
         Hai theirPK -> do
             modifyTransport $
                 Transport.framedEnvelope (sign mySK) (verify theirPK)
-            guardPeerId (Proxy :: Proxy c) peerId HandshakeResult
+            guardPeerId (Proxy @c) peerId  HandshakeResult
                 { hrPeerId   = theirPK
                 , hrPreSend  = idM
                 , hrPostRecv = idM
@@ -89,8 +94,8 @@ signedPayloads sk hr = mapOutput (sign sk) (verify (hrPeerId hr)) hr
 -- If the 'hrPeerId' is not found by the lookup action (ie. it returns 'IO
 -- False'), the handshake is aborted with 'Unauthorized'.
 whitelist
-    :: (Eq n, Serialise (Crypto.PublicKey c))
-    => Proxy c
+    :: forall i o n c proxy. (Eq n, Serialise (Crypto.PublicKey c))
+    => proxy c
     -> (n -> IO Bool) -- ^ 'True' if this peer id is granted access, 'False' otherwise
     -> Handshake SimpleError n i o
     -> Handshake SimpleError n i o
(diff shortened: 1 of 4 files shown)


src/Oscoin/P2P/Handshake/Simple.hs
```diff
@@ -108,17 +110,29 @@ mapRecvError = \case
...
 
 guardPeerId _ _ res = pure res
 
+guardPeerIdNot
+    :: forall i o n c proxy. (Eq n, (Serialise (Crypto.PublicKey c)))
+    => proxy c
+    -> n
+    -> HandshakeResult i o n
+    -> HandshakeT SimpleError IO (HandshakeResult i o n)
+guardPeerIdNot _ peerId res | peerId == hrPeerId res = do
```
adinapoli-mndc: methinks -- what about (pseudo code, won't compile):

``` haskell
guardPeer :: Proxy -> n -> (HashshakeResult -> n -> Bool) -> SimpleError -> HandshakeResult -> m () -> m ()
guardPeer _ peerId predicate err res action | predicate res peerId = do
  handshakeSend (Bai err :: HandshakeMessage c)
  throwError err
guardPeer _ _ _ _ action = action

peerIdMatch :: HandshakeResult -> NodeId -> Bool
peerIdMatch = (==) . hrPeerId
```

and later:

```
...
guardPeer (peerIdMatch)
guardPeer (not . peerIdMatch) 
```

Seems like a recurring pattern worth the refactoring? 🤔

test/Oscoin/Test/P2P/Handshake.hs
```diff
@@ -16,23 +16,25 @@ import           Hedgehog
...
+    [ ("prop_simple",            prop_simple d)
+    , ("prop_simpleRejectsSelf", prop_simpleRejectsSelf d)
+    , ("prop_secure",            prop_secure d)
+    , ("prop_secureRejectsSelf", prop_secureRejectsSelf d)
     ]
 
-prop_simpleHandshake :: forall c. Dict (IsCrypto c) -> Property
-prop_simpleHandshake Dict = withTests 1 . property $ do
+prop_simple :: forall c. Dict (IsCrypto c) -> Property
+prop_simple Dict = withTests 1 . property $ do
```
adinapoli-mndc: Orthogonal: now that we have a mock crypto implementation, can we remove the `withTests 1` if the reason it was there in the first place was tests running time?

test/Oscoin/Test/P2P/Handshake.hs
```diff
@@ -125,6 +147,26 @@ prop_secureHandshake Dict = withTests 1 . property $ do
             Left Noise.DecryptionError -> success
             _                          -> failure
 
+prop_secureRejectsSelf :: forall c. Dict (IsCrypto c) -> Property
+prop_secureRejectsSelf Dict = withTests 1 . property $ do
+    (pk, sk) <- evalIO (Crypto.generateKeyPair @c)
```
adinapoli-mndc: Tip: if you want a predictable key-pair, you can use the `FastRandomBytes` abstraction from `Oscoin.Test.Crypto.PubKey.Arbitrary` to write something like:

```
seed <- forAll (Gen.word64 Range.linearBounded)
(pk, sk) <- forAll (withFastRandomBytes (SplitMix.mkSMGen seed) generateKeyPair)
```

src/Oscoin/P2P/Handshake/Simple.hs
kim: Yeah... but... it's different error constructors

test/Oscoin/Test/P2P/Handshake.hs
kim: No it's because they use actual sockets, which is dawg slow on our current CI

src/Oscoin/P2P/Handshake/Simple.hs
adinapoli-mndc: oh, I see, so you might need an injective type family to map one to the other to use it inside `guardPeer`? At which point it feels.. bulky I suppose?

kim: It's bulky, yeah, because that error type ought to not be thrown from another module, imho. But I'm a good 🏌️ , so here you go: ec2c276898ce0d261c8517187d01f08507be971b

adinapoli-mndc: Good lad, thanks! 😁 ❤️

test/Oscoin/Test/P2P/Handshake.hs
```diff
@@ -133,13 +186,29 @@ tripping' x f g = do
...
+
+genSeed :: Gen SplitMix.SMGen
+genSeed = SplitMix.mkSMGen <$> Gen.word64 Range.linearBounded
+
+genKeyPair
+    :: forall c. (HasHashing c, HasDigitalSignature c)
+    => Gen (PublicKey c, PrivateKey c)
+genKeyPair = do
+    seed <- genSeed
+    pure $ withFastRandomBytes seed (generateKeyPair @c)
```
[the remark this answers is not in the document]
kim: For QC, yes. 

The issue is that if you want more than one keypair, the generator needs to use the same seed for all of them. I'm still trying to figure out how to do this best. Someone else is going to be bitten by this again, for sure.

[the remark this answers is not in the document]
adinapoli-mndc: Oh, good catch -- this is exactly the same pitfall I also fell into while working on that integration test to generate random ports -- me using `choose (0, x)` meant that the ports were the same for all the 100+ tests as they were all originated by the same seed.

Incidentally, I think I also fell into the same exact pitfall in my patch on the address format, drat!

> I'm still trying to figure out how to do this best.

That is indeed tricky -- if you come across some clever technique keep me posted, as it would be nice to standardise and distill in a separate module all can benefit from.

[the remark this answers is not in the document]
kim: So the best I could come up with for now was [this](https://github.com/oscoin/oscoin/pull/448/files#diff-a15a0d4342481cb0550dbfd07662d05bR71). It's a bit annoying having to pattern match on the result (because `-Wall`).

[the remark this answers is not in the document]
adinapoli-mndc: Oh, I see -- essentially, generating a bunch of keys beforehand, and using those in the tests. Yes, I cannot think about any better way either. How annoying!

PR #450, opened 2019-04-01
Explicitly tie the lifetime of the EKG server to the other main threads

kim:


exe/node/Main.hs
```diff
@@ -207,3 +205,8 @@ main = do
     miner nod gos = runGossipT gos . runNodeT nod $ Node.miner
     storage nod =
         hoistStorage (runNodeT nod) (Node.storage Nakamoto.validateBasic)
+
+    runEkg store host port = do
+        tid <- forkEkgServer store host port
+        withException (threadDelay maxBound) $ \(e :: SomeAsyncException) ->
```
adinapoli-mndc: Let's just hope that one of those cyberpunk hipsters won't find a way to keep a machine going for more than 292471.208677536 years 😂

kim: I briefly considered blocking on an empty [M|T]Var, but that steampunk compiler is too smart ;)

PR #451, opened 2019-04-02
Add 'Ord' and 'Eq' constraints for hashes

geigerzaehler:
We require the `Ord (Hash c)` and `Eq (Hash c)` constraints for `HasHashing c` instances. This allows us to eliminate this explicit constraints in functions


adinapoli-mndc:
@kim Yeah, we discussed this with @geigerzaehler over the wire. I am also ok-ish with that, i.e. I do see the gain in terms of boilerplate reduction, and I think this is enough to justify the extra constraints.

The only sour taste in my mouth is that, from a "lawful" perspective, really ordering an hash makes very little sense. `Eq`? I'm OK with that. But for `Ord` things are totally hairy: how is `098f6bcd4621d373cade4e832627b4f6` greater or equal of `69e153e4d7add22f245e24de590eec21`? You can't really tell.

However, it's a necessary evil to be able to work with the codebase, so I suppose we can merge this one, as it will clean up the code in some other places, and perhaps open a tech debt ticket where we explore whether or not we can remove this `Ord` constraint. In practice, I don't think this will be possible/feasible. To begin with, you do want to talk about ordered sets and maps containing `BlockHash`es, and the minute you do that, you lost already (due to `Data.Set.Set` and `Data.Map.Map` imposing you on `Ord` constraint on the keys 😢 )

geigerzaehler:
> @geigerzaehler Can we also remove `UndecidableInstances` in some places with this?

Unfortunately not.

kim:
>  But for Ord things are totally hairy

Yeah that’s why I generally prefer HashMap, even though it’s often more painful to work with.

PR #452, opened 2019-04-02
Protocol refactorings

geigerzaehler:
This is a collection of four commits that improve the code and prepare it for further changes.

### DIFF
--- a/src/Oscoin/Consensus/Mining.hs
+++ b/src/Oscoin/Consensus/Mining.hs
@@ -9,16 +9,19 @@ import           Oscoin.Consensus.Types
 import qualified Oscoin.Crypto.Hash as Crypto
 import           Oscoin.Storage.Block.Abstract (BlockStoreReader)
 import qualified Oscoin.Storage.Block.Abstract as BlockStore
+import           Oscoin.Storage.Receipt (MonadReceiptStore)
+import qualified Oscoin.Storage.Receipt as ReceiptStore
 import           Oscoin.Storage.State.Class (MonadStateStore)
 import qualified Oscoin.Storage.State.Class as StateStore
 import           Oscoin.Time.Chrono (toNewestFirst)
 
 import           Oscoin.Crypto.Blockchain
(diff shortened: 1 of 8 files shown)


src/Oscoin/Consensus/Mining.hs
```diff
@@ -34,20 +38,21 @@ mineBlock
        )
     => BlockStoreReader c tx s m
     -> Consensus c tx s m
-    -> Evaluator st tx b
+    -> Evaluator st tx o
     -> Timestamp
-    -> m (Maybe (Block c tx (Sealed c s), st, [Receipt c tx b]))
```
adinapoli-mndc: Interesting -- this _used_ to have this type, but when we discussed this with @kim during my work on #402, we made the miner return also the state and the list of receipts. The rationale was that block evaluation/validation might still fail, and you might not want to update the state _just yet_. What do you think?

src/Oscoin/Node.hs
```diff
@@ -256,6 +252,9 @@ getBlocks d =
 lookupTx :: (MonadIO m) => Hashed c tx -> NodeT c tx st s i m (Maybe (TxLookup c tx))
 lookupTx tx = withBlockStore (`BlockStore.lookupTx` tx)
 
+lookupReceipt :: (Ord (Hash c), MonadIO m) => Hashed c tx -> NodeT c tx st s i m (Maybe (Receipt c tx RadicleTx.Output))
```
cloudhead: :mag:  :straight_ruler:

src/Oscoin/Consensus/Mining.hs
geigerzaehler: I agree that `mineBlock` does all the wrong things here. However this is a stopgap measure to reduce the code in `Oscoin.Node.mineBlock` and still make the tests pass. I intend to move the interactions with the state/block/receipt store in `mineBlock` behind an abstraction. And the state and receipts will only be written after a mined block is accepted by the protocol.

PR #453, opened 2019-04-02
Transition 'MonadStateStore' to dictionary API

geigerzaehler:
We start transitioning the MTL-style `MonadStateStore` to a dictionary based API. For this we introduce a generic `ContentStore` dictionary that can be reused by `MonadReceiptStore`.


src/Oscoin/Node.hs
```diff
@@ -84,6 +85,7 @@ withNode hConfig hNodeId hMempool hStateStore hBlockStore hProtocol hEval hConse
     bracket open close
   where
     open = do
+        hStateStore' <- newIoContentStore
```
adinapoli-mndc: Perhaps `newContentStoreIO` is more fitting, as it's consistent to other similar APIs? (cfr. `newTVarIO` , etc)

src/Oscoin/Node/Trans.hs
```diff
@@ -58,6 +59,7 @@ data Handle c tx st s i = Handle
     { hConfig       :: Config
     , hNodeId       :: i
     , hStateStore   :: StateStore.Handle c st
+    , hStateStore'  :: ContentStore c st IO
```
adinapoli-mndc: I am fairly sure the answer is "yes", but let me ask anyway: do you plan to clean this up once the mtl-style stuff is completely replaced, so that we don't have these almost duplicate record fields? (i.e. `hStateStore` vs `hStateStore'`).

src/Oscoin/Node/Trans.hs
```diff
@@ -106,13 +108,5 @@ instance (Ord (Hash c), MonadIO m) => MonadReceiptStore c tx RadicleTx.Output (N
...
-        h <- asks hStateStore
-        lift $ StateStore.withHandle h (pure . StateStore.lookupState k)
-    storeState st = do
-        h <- asks hStateStore
-        liftIO $ StateStore.storeStateIO h st
-
-    {-# INLINE lookupState #-}
-    {-# INLINE storeState #-}
+instance (MonadIO m) => MonadStateStore c st (NodeT c tx st s i m) where
+    getStateStore = hoistContentStore liftIO <$> asks hStateStore'
```
adinapoli-mndc: I think this answers the above 😁 .

src/Oscoin/Storage/ContentStore.hs
```diff
@@ -0,0 +1,54 @@
...
+
+import qualified Oscoin.Crypto.Hash as Crypto
+
+import qualified Data.Map as Map
+import           Lens.Micro
+import           Lens.Micro.Mtl
+
+data ContentStore crypto a m = ContentStore
+    { storeContent  :: a -> m ()
+    , lookupContent :: Crypto.Hashed crypto a -> m (Maybe a)
```
adinapoli-mndc: Name brainstorming: how does `KeyValueStore` sound?

src/Oscoin/Storage/ContentStore.hs
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+hoistContentStore :: (forall x. m x -> n x) -> ContentStore crypto a m -> ContentStore crypto a n
+hoistContentStore lift' c = ContentStore
```
adinapoli-mndc: category theory nitpick: this is technically a natural transformation, but I can live with `lift'` 😬

src/Oscoin/Storage/ContentStore.hs
```diff
@@ -0,0 +1,54 @@
...
+
+newIoContentStore
+    :: forall crypto a m.
+       ( MonadIO m
+       , Crypto.Hashable crypto a)
+    => IO (ContentStore crypto a m)
+newIoContentStore = do
+    mapVar <- newMVar mempty
+    let lookupContent hash = liftIO $ readMVar mapVar <&> Map.lookup hash
+    let storeContent a = liftIO $ modifyMVar_ mapVar $ \m -> pure $ Map.insert (Crypto.hash @crypto a) a m
```
adinapoli-mndc: Should we worry about inserting/replacing the content? Do we need to keep a "list of previous contents" we can use for looking things up "in the past"?

src/Oscoin/Storage/ContentStore.hs
geigerzaehler: I think `KeyValueStore` is confusing because we fixed keys. `ContentStore` comes from “content addressable storage” and the fact that the content determines how we can look it up.

src/Oscoin/Storage/ContentStore.hs
geigerzaehler: I don’t think we need this. The keys is the content’s hash so we cannot replace content strictly speaking. If two values have the same hash then they are the same.

src/Oscoin/Node/Trans.hs
geigerzaehler: This has been cleaned up.

src/Oscoin/Storage/ContentStore.hs
adinapoli-mndc: > If two values have the same hash then they are the same.

I see -- and it can never happen that two values have the same hash but they were created/stored at different point in time, and perhaps we'll need to rollback to a previous version? I suppose not, but it's still good to think about these things when doing these reviews 😀

geigerzaehler: The store is indeed quite dumb now and does deal with rollbacks. Since this is intended to store states I don’t think we even need it at the moment. But it is a good idea to keep an eye on this and possibly extend it.

src/Oscoin/Storage/ContentHashStore.hs
```diff
@@ -0,0 +1,73 @@
...
+import qualified Oscoin.Crypto.Hash as Crypto
+
+import qualified Data.Map.Strict as Map
+import           Lens.Micro
+import           Lens.Micro.Mtl
+
+data ContentHashStore crypto value m = ContentHashStore
+    { _storeHashContent  :: value -> m ()
+    , _lookupHashContent :: Crypto.Hashed crypto value -> m (Maybe value)
+    }
```
cloudhead: By convention, we use `Handle` for this and then we would use it externally as `ContentHashStore.Handle`

src/Oscoin/Storage/ContentHashStore.hs
```diff
@@ -0,0 +1,73 @@
+-- | 'ContentHashStore' is an interface for storing values adressed by
+-- their hashes. Conceptually 'ContentHashStore' is a mutable 'Map'
+-- with hashes as keys.
+module Oscoin.Storage.ContentHashStore
```
cloudhead: More name bikeshedding: can we simply go for `HashStore`? The "Content" there doesn't really add any information.

kim: Sounds like Görlitzer Park 😆

src/Oscoin/Storage/ContentHashStore.hs
geigerzaehler: I assumed we only use `Handle` for the old MTL-style APIs, like `StateStore.Handle`. The new dictionary APIs, like `BlockStore` does not use the term `Handle`. Also this module is intended to be imported unqualified as all the names include `ContentHash`. If we want to keep it at that we would need to use the name `ContentHashStoreHandle` which would be quite long. `ContentHashStore` flows better I think.

src/Oscoin/Storage/ContentHashStore.hs
cloudhead: :joy:

src/Oscoin/Storage/ContentHashStore.hs
cloudhead: Hmm yeah I think that makes sense, since the Handle is not actually opaque here, as it is with mempool, telemtry etc.

exe/node/Main.hs
```diff
@@ -131,7 +131,8 @@ main = do
     gen          <- Yaml.decodeFileThrow genesis :: IO GenesisBlock
     genState     <- either (die . fromEvalError) pure $
                         evalBlock Rad.txEval Rad.pureEnv gen
-    stStore      <- StateStore.fromStateM genState
+    stStore      <- newHashStoreIO
+    storeHashContent stStore genState
```
adinapoli-mndc: Minor: in the future you might think about adding a `newEmptyHashStoreIO` and a `newHashStoreIO`, where the former will be exactly what you have now, and the latter would allow you to take an initial value as input, so that you could theoretically write the above as:

```
stStore <- newHashStoreIO genState
```

src/Oscoin/Storage/State.hs
```diff
@@ -1,62 +1,34 @@
...
 import           Oscoin.Prelude
 
 import           Oscoin.Crypto.Blockchain.Block (StateHash)
-import           Oscoin.Crypto.Hash (Hash)
 import qualified Oscoin.Crypto.Hash as Crypto
+import           Oscoin.Storage.HashStore
+
+type StateStore c st m = HashStore c st m
+
+class Monad m => MonadStateStore c st m | m -> st, m -> c where
```
adinapoli-mndc: Question: do we need to maintain this typeclass at all, after this patch? I thought one of the main objective of this PR was to get rid of this typeclass entirely. What am I missing? Perhaps this is just an intermedia step towards getting rid of it? 😉

exe/node/Main.hs
geigerzaehler: Yes, makes a lot of sense to add this in the future.

src/Oscoin/Storage/State.hs
geigerzaehler: We are very close to getting rid of the type class but it still requires some effort and I wanted to keep the PR small. I’ll follow up on this after I also transition the receipt store.

adinapoli-mndc: 👌 👏

PR #455, opened 2019-04-02
Fixed #446 (intermittent concurrency bug in SQLite)

### LINKED ISSUE #446: `Protocol` concurrency issue
_Some_ incantation of running multiple nodes (locally) can lead to one node terminating with the following error:

`SQLite3 returned ErrorError while attempting to perform step: cannot start a transaction within a transaction`

adinapoli-mndc:
Fixed #446 (see that issue for a detailed explanation of what's going on).

I have went down the `MVar` route, introducing this small `Transaction` API to make things slightly more composable.

I have tested this by leveraging a nifty tool called [hivemind](https://github.com/DarthSim/hivemind), which doesn't require `tmux` and it allows multiple processes to be run in the same window with all their output. All it takes is to write a single `Procfile` to instrument the experiment.

I have then added a mini-cluster of 3 nodes which can be run by typing:

```
./procfiles/run_devnet_cluster_demo.sh
``` 

(Requires `stack` and `hivemind`).


src/Oscoin/Storage/Block/SQLite.hs
```diff
@@ -78,22 +78,23 @@ lookupBlock
     => Handle c tx s
     -> BlockHash c
     -> IO (Maybe (Block c tx (Sealed c s)))
-lookupBlock Handle{hConn} h = Sql.withTransaction hConn $ do
-    result :: Maybe (BlockHeader c (Sealed c s)) <- listToMaybe <$> Sql.query hConn
-        [sql| SELECT parenthash, datahash, statehash, timestamp, difficulty, seal
-                FROM blocks
-               WHERE hash = ? |] (Only h)
+lookupBlock Handle{hConn} h = withMVar hConn $ \conn ->
```
adinapoli-mndc: I suspect we would benefit from using the `Transaction` abstraction here as well. On it.

procfiles/run_devnet_cluster_demo.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
```
kim: Uhm, well, I'm not sure I want those guys in version control just yet. Maybe a separate PR?

src/Oscoin/Storage/Block/SQLite/Internal.hs
```diff
@@ -58,29 +59,54 @@ import           Codec.Serialise
...
+-- foot when working with MVars. It ensures that 'withMVar' is called in a
+-- single place ('runTransaction') rather than being scattered in multiple places.
+-- It also allows transactions to be composed together.
+newtype Transaction m a = TransactionT (ReaderT Connection m a)
+    deriving (Functor, Applicative, Monad, MonadReader Connection, MonadIO)
+
 -- | A handle to an on-disk block store.
 data Handle c tx s = Handle
-    { hConn       :: Connection                    -- ^ Connection to on-disk storage for non-orphan blocks.
+    { hConn       :: MVar Connection               -- ^ Connection to on-disk storage for non-orphan blocks.
```
kim: How about `newtype`ing `MVar Connection`?

src/Oscoin/Storage/Block/SQLite/Internal.hs
```diff
@@ -156,15 +181,16 @@ getGenesisBlock
        , StorableTx c tx
        , FromField (Hash c)
        )
-    => Connection
+    => MVar Connection
```
kim: Should this also run in the `Transaction` monad?

src/Oscoin/Storage/Block/SQLite/Internal.hs
```diff
@@ -174,21 +200,23 @@ getChainSuffix
        , FromField s
        , FromField (Hash c)
        )
-    => Connection
+    => MVar Connection
```
kim: Dito

procfiles/run_devnet_cluster_demo.sh
adinapoli-mndc: yup, this is why I have a separate, standalone commit for that 😬

src/Oscoin/Storage/Block/SQLite/Internal.hs
adinapoli-mndc: Can do! What about `ConnectionLock` or something like that?

src/Oscoin/Storage/Block/SQLite/Internal.hs
adinapoli-mndc: oops, yes.

src/Oscoin/Storage/Block/SQLite/Internal.hs
kim: I'm not sure ... does it need to "leave" this module? I think just `Connection` would be fine.

adinapoli-mndc: At the moment yes, it does leave this module (cfr. the `WhiteBox` tests). I can simply type alias this though, if this will still aid the visibility/readability within the `Internal` module.

src/Oscoin/Storage/Block/SQLite/Internal.hs
adinapoli-mndc: d3378c5

src/Oscoin/Storage/Block/SQLite/Internal.hs
adinapoli-mndc: d3378c5

procfiles/run_devnet_cluster_demo.sh
adinapoli-mndc: d3378c5

src/Oscoin/Storage/Block/SQLite/Internal.hs
```diff
@@ -58,29 +59,54 @@ import           Codec.Serialise
 import           Oscoin.Storage.Block.Cache (BlockCache)
 import qualified Oscoin.Storage.Block.Cache as BlockCache
 
+-- Simple, composable transaction mechanism to avoid shooting ourselves in the
```
cloudhead:
```suggestion
-- | Simple, composable transaction mechanism to avoid shooting ourselves in the
```

src/Oscoin/Storage/Block/SQLite/Internal.hs
```diff
@@ -58,29 +59,54 @@ import           Codec.Serialise
...
+
 -- | Opens a connection to the SQLite DB.
-open
-    :: String
-    -> IO (Handle c tx s)
+open :: String -> IO (Handle c tx s)
 open path =
-    Handle <$> (Sql.open path >>= setupConnection)
+    Handle <$> (Sql.open path >>= newMVar >>= setupConnection)
            <*> BlockCache.newBlockCache 2016 -- TODO(and) pass this parameter externally
```
cloudhead:
```suggestion
           <*> BlockCache.newBlockCache 2016 -- TODO(adn) pass this parameter externally
```

adinapoli-mndc:
Will squash and merge once @kim approves.

PR #458, opened 2019-04-03
Move mempool, orphanage, API tests to new structure

geigerzaehler:
In three separate commits we
* move `Oscoin.Test.API` to `Test.Oscoin.API`
* move `Oscoin.Test.Storage.Block.Orphanage` to `Test.Oscoin.Storage.Block.Orphanage`
* extract `Test.Oscoin.Node.Mempool` from `Oscoin.Tests`

### DIFF
--- a/test/Test/Oscoin/API.hs
+++ b/test/Test/Oscoin/API.hs
@@ -1,6 +1,6 @@
 -- | Tests behavior of the Node through its API using a test
 -- implementation of 'MonadClient'.
-module Oscoin.Test.API
+module Test.Oscoin.API
     ( tests
     ) where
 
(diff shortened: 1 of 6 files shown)


test/Test/Oscoin/API.hs
```diff
@@ -24,8 +24,8 @@ import           Test.Tasty.HUnit.Extended
 import           Test.Tasty.QuickCheck
 
 
-tests :: forall c. Dict (IsCrypto c) -> [TestTree]
-tests Dict =
+tests :: forall c. Dict (IsCrypto c) -> TestTree
+tests Dict = testGroup "Oscoin.API"
```
kim: Wait .. shouldn't this always be _literally_ the module's name?

geigerzaehler: I’d be ok with that, too. But I omitted the `Test` prefix in all my changes and this is also what is documented in [“Test organization”](https://github.com/oscoin/oscoin#test-organization) in the readme.

kim: Ok, guess I missed that. I think mechanically the module name is preferable, as we can eventually just have it generated.

geigerzaehler: I’ll switch to that

geigerzaehler:
Can you have another look @kim?

Issue #389, opened 2019-03-04
Account Address Format


cloudhead:
How do oscoin account addresses look like?
* Haskell types
* Serialization
* Documentation of format

This is a pretty good starting point: https://wiki.parity.io/External-Address-Format-(SS58)

adinapoli-mndc:
Fixed by https://github.com/oscoin/oscoin/pull/431

PR #459, opened 2019-04-03
Shuffle around config and option parsing

### LINKED ISSUE #427: Move Args parser into its own module
This voices the concern raised [here](https://github.com/oscoin/oscoin/pull/426#issue-263548725). Essentially our `Args` parser in `exe/Main.hs` is growing a bit out of control, and I think it's time to move it in a standalone module.

@kim suggested we move everything into the library, as that would allow us to write better and more specific tests.

kim:
Fixes #427

Feel free to bikeshed, the color palette is limited to red, green, and blue. Ie. better to change stuff in subsequent patches (if desired), than to agonise over this one, as it changes a lot of files.


exe/node/Main.hs
```diff
@@ -4,7 +4,7 @@ import           Oscoin.Prelude hiding (option)
 
 import qualified Oscoin.API.HTTP as HTTP
 import           Oscoin.CLI.KeyStore (readKeyPair)
-import           Oscoin.CLI.Parser (environmentParser, keyPathParser)
+import           Oscoin.Configuration
```
adinapoli-mndc: Drat, I have introduced the same module in my address-format PR -- one of us will get conflicts 😁

src/Oscoin/Configuration/Options.hs
```diff
@@ -0,0 +1,25 @@
...
+                 , allEnvironments
+                 , readEnvironment
+                 , renderEnvironment
+                 )
+
+import           Options.Applicative
+
+
+environmentParser :: Parser Environment
+environmentParser = option (eitherReader readEnvironment)
```
adinapoli-mndc: Blue paint bikeshedding: Feels slightly artificial to have a `Oscoin.Configuration.Options` module exporting only a `Parser` -- perhaps this belongs in one of those `CLI/Parser/Whatever` submodules?

kim: This was accidentally committed - the parser is in `Oscoin.Configuration`.

Also consider the `CLI` is about the (administrative) commands, so the reason to have them in the "gobal" configuration is to be able to re-use between `Node` and `CLI` options.

adinapoli-mndc:
Gonna merge it, as Kim is only partially available and as it introduces the exact same module I am using as part of #389, I think it's simpler if I do handle the conflicts on my side.

PR #461, opened 2019-04-04
Use dictionary-style API for ReceiptStore

geigerzaehler:
Continuing the work of #453 we implement a dictionary style API intended to replace `MonadReceiptStore`. We do not yet eliminate the `MonadReceiptStore` class. This is left for a future PR.

To accomplish the change we introduce `ContentStore`, a generalized version of `HashStore`. `ContentStore` is an interface for a mutable key-value store where the key is computed from the value.


src/Oscoin/Storage/Receipt.hs
```diff
@@ -1,80 +1,67 @@
...
-    storeL :: Lens' r (Store c tx o)
+-- | 'ContentStore' for 'Receipt's where a receipt is adressed by the
+-- hash of the transaction the receipt is for.
+type ReceiptStore c tx o m = ContentStore (Crypto.Hashed c tx) (Receipt c tx o) m
 
-emptyStore :: Store c tx o
-emptyStore = Map.empty
+-- | Underlying data structure for pure versions of 'ReceiptStore'
+-- created with 'mkStateReceiptStore'.
+type ReceiptStoreMap c tx o = Map (Crypto.Hashed c tx) (Receipt c tx o)
```
adinapoli-mndc: Minor: reading the documentation makes clear what this is about, but I guess this might be confusing when encountered "in the wild" and might confuse the reader trying to understand the different between a `ReceiptStore` and a `ReceiptStoreMap`. Do you think moving this into a separate `Pure` module might makes sense, or name it like `PureContentStore` or something like that?

Follow up (as I keep writing something and scrapping it 😅 ):if I understand correctly, this is exported and used _only_ in tests (orthogonal thought: we should really tame those tests so that we don't need to introduce these ad-doc types). If that's true, methinks it's better if we name this _explicitly_ (as in `TestReceiptStore`) to convey the fact this is something which needs to be used _only_ for testing?

Thoughts?

geigerzaehler: What do you think about calling this `ReceiptMap`. I could also remove it (or not export it).

adinapoli-mndc: I am a firm believer of code which states intent, so my preference would still be to name it "TestBlah", or to move it into the test code (if that works) while keeping the name as-it-is, if you like it better 😁 

But I am not forcing refactorings on people, so definitely feel free to act according to your best judgement (and don't feel this should stop you from merging this PR altogether).

cloudhead: I had the same thought while reading. `ReceiptMap` would make more sense, or as @adinapoli-mndc proposed, moving it to a `Pure` module.

geigerzaehler: I’m going to rename it to `ReceiptMap`. I don’t think moving it into a `Pure` module would make sense since that module would only contain `ReceiptMap` and `mkStateReceiptStore`.

PR #463, opened 2019-04-04
Fix #462 (expose Somenet and make pattern match exhaustive)

### LINKED ISSUE #462: Expose Somenet constructor to get rid of non-exhaustive pattern match
As part of #389 we have a function `Network -> Maybe AddressTypeTag` which pattern match is non-exhaustive due to `Somenet` not being exposed. Non-exhaustive pattern matches can backfire in the future, so it's better to expose the constructor (with proviso , obviously).

adinapoli-mndc:
This PR exposes (with proviso) the `Somenet` constructor, so that we can make the pattern match in the `Address` module exhaustive.


adinapoli-mndc:
Closing this after a discussion with @kim , as he's now on top of it.

PR #465, opened 2019-04-04
Simplify orphanage tests

geigerzaehler:
* We eliminate the use of monadic properties
* We simplify `propInsertOrphans` replacing `genOrphanChainsFrom` with `arbitraryBlockChain`.

### DIFF
--- a/test/Oscoin/Test/Storage/Block/SQLite.hs
+++ b/test/Oscoin/Test/Storage/Block/SQLite.hs
@@ -3,30 +3,19 @@ module Oscoin.Test.Storage.Block.SQLite
     , DummySeal
     , withSqliteDB
     , withMemStore
-    , insertOrphans
-    , bestChainOracle
-    , shuffledOrphanChains
-    , Shuffled(..)
     ) where
 
 import           Oscoin.Prelude
-import qualified Prelude
 
-import           Codec.Serialise
-import qualified Data.Sequence as Seq
 
-import           Oscoin.Consensus.Nakamoto (blockScore)
 import           Oscoin.Crypto.Blockchain hiding (genesisBlock)
 import           Oscoin.Data.RadicleTx
 import qualified Oscoin.Storage.Block.Abstract as Abstract
-import           Oscoin.Storage.Block.Orphanage
 import           Oscoin.Storage.Block.SQLite as Sqlite
 import           Oscoin.Storage.Block.SQLite.Internal as Sqlite
 import qualified Oscoin.Time as Time
-import           Oscoin.Time.Chrono (toNewestFirst)
 
 import           Oscoin.Test.Crypto
-import           Oscoin.Test.Crypto.Blockchain.Generators
 import           Oscoin.Test.Data.Rad.Arbitrary ()
 import           Oscoin.Test.Data.Tx.Arbitrary ()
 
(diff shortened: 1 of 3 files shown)


test/Oscoin/Test/Storage/Block/Orphanage.hs
```diff
@@ -44,18 +44,12 @@ tests d =
...
-            ps <- forM chains $ \(orphanChain, _missingLink) -> do
-                     let actual = toList $ getOrphans (insertOrphans (toNewestFirst $ blocks orphanChain) orphanage)
-                     pure (sort actual === sort (map blockHash (toNewestFirst $ blocks orphanChain)))
-            pure $ foldl' (.&&.) (property True) ps
+propInsertOrphansGetOrphans :: forall c. Dict (IsCrypto c) -> Property
+propInsertOrphansGetOrphans Dict =
+    forAllShow arbitraryBlockchain (toS . showChainDigest) $ \(blockchain :: Blockchain c () ())->
+        let blks = toNewestFirst $ blocks blockchain
+            orphanage = insertOrphans blks (emptyOrphanage @c blockScore)
+        in Set.fromList (map blockHash blks) === getOrphans orphanage
```
adinapoli-mndc: I am actually not sure this simplification buys us anything, unfortunately 😢 When you call `arbitraryBlockchain`, you are going to create a _proper_ blockchain, i.e. a blockchain with no orphans (this might surprise you as `arbitraryBlockchain` used to do that, I think, but that generated all kinds of problems).

As a consequence of this, now you are effectively testing that you can store things inside the orphanage, but this is a bit of a moot test, as things gets interesting the minute you actually have orphans, and this is why we have the `genOrphanChainsFrom` tests 😉

test/Oscoin/Test/Storage/Block/Orphanage.hs
```diff
@@ -67,25 +61,24 @@ propSelectBestCandidateSingleChoice Dict = do
...
-                              . toNewestFirst
-                              . reverse
-                              . toBlocksOldestFirst orphanage''
-                              . snd <$> selectBestChain [blockHash defaultGenesis] orphanage''
-                let p2 = bestChain === Just ((toNewestFirst . blocks) orphanChain <> [missingLink])
-                pure [p1, p2]
-            pure $ classifyChainsByScore chains $
-                       foldl' (.&&.) (property True) (mconcat ps)
+            classifyChainsByScore chains
+                $ conjoin
```
adinapoli-mndc: I think the reason why I had added the monadic code in the first place was that it makes slightly easier to pinpoint errors, as `conjoin` will simply compose everything together but that won't necessary help to isolate the failing property (in term of stacktracking). Would you do a simple test for me? Try introducing a (trivial) failure in one of the properties, and compare the stacktraces you get from one or the other piece of code -- if they are equivalent, then I'm fine with that 👍

test/Oscoin/Test/Storage/Block/Orphanage.hs
geigerzaehler: I reverted the change to the generator. We keep using `genOrphanChainsFrom`. However I still have the feeling that we don’t actually need this. I’ll think about it and we can discuss it some other time.

test/Oscoin/Test/Storage/Block/Orphanage.hs
geigerzaehler: I checked: Using `monadic runIdentity` and `.&&.` over `conjoin` does not give us anything. The error reporting is improved by `===` from `Test.Quickcheck.Extended`.

geigerzaehler:
@adinapoli-mndc I added some further changes that should make the code more readable.

test/Oscoin/Test/Storage/Block/Orphanage.hs
adinapoli-mndc: Perfect 👍

geigerzaehler:
Incidently this also makes the tests about four times faster 🚀

Issue #460, opened 2019-04-04
Remove chain selection from the pure block store


adinapoli-mndc:
This ticket echoes what @geigerzaehler said a while ago: at the moment we _still_ have two agents that perform chain selection: the `Protocol` but also the pure block store (i.e. `Pure.hs`).

The reason why we have this, it's fairly simple: we do need to compare the result of chain selection that `Protocol` performs with the one `Pure.hs` does (which is way, way, simpler) in our "equivalence" tests, to make sure they behave the same (as the former is quite more complex due to the use of the Orphanage).

But if we think about it, what we really need is a _reference implementation_ for chain selection, not the _whole_ pure block store, i.e., the two can be separated out.

I propose the following:

1. We introduce a separate module (name to be bikeshedded) where only the _reference implementation_ of the chain selection will be provided;

2. We keep `Pure.hs` really dumb, as in `Map BlockHash Block` dumb (modulo domain-specific stuff still needed);

3. We amend the `Equivalence` tests accordingly so that (potentially) the stores can be mixed & matched but the result must match. We can even use only the pure store for both tests now, in theory (although I have to think about implications with switchToFork & co and how the pure block store might support it).

This has a bunch of practical advantages:

1. It gets rid of a bunch of corner cases in our application;
2. It allow us to start making the consensus tests a bit more realistic: at the moment they simply use the pure store to perform chain selection, but that's not what we want: we want to put our consensus at test by using the actual chain selection algorithm which will be used in production.

@cloudhead thoughts?

cloudhead:
@geigerzaehler is working on this if I'm not mistaken?

adinapoli-mndc:
@cloudhead When me and @geigerzaehler spoke about it, he didn't explicitly say so, but rather than PR #466 was sitting in his stash and as such he pushed it upstream.

I don't really mind who is working on what, provided we state our intent clearly to avoid stepping on each other toes 😉 If Thomas plans on tackling it in full, I would appreciate if he could assign himself to it 😃 .

adinapoli-mndc:
Me and @cloudhead had quite a productive design session this morning, which also resulted in a bit of a retrospective and a post-mortem on the entire `Protocol-BlockTree-BlockStore` saga which I am now going to share: essentially, the reason why we are finding difficult to separate out chain selection from `Pure.hs` is that the combination of main_chain+orphans+chain_selection was never meant to be separated out, as those component make sense only if considered holistically.

[...]

PR #466, opened 2019-04-04
Use dumb in-memory Blockstore in API tests

geigerzaehler:
Before, the Node API tests used the STM blockstore that does chain selection *and* the protocol that is now responsible for chain selection. We change the block store for the API tests to a dumb in-memory blocks store. This is a step towards resolving #460.

Also removes redundant `BlockStore.Pure` instances


src/Oscoin/Storage/Block/Memory.hs
```diff
@@ -0,0 +1,58 @@
...
+
+import           Oscoin.Crypto.Blockchain hiding (lookupTx)
+import           Oscoin.Crypto.Blockchain.Block (Block)
+import qualified Oscoin.Crypto.Hash as Crypto
+import qualified Oscoin.Storage.Block.Abstract as Abstract
+import qualified Oscoin.Time.Chrono as Chrono
+
+import qualified Data.List as List
+
+newBlockStoreIO
```
cloudhead: The `IO` suffix kind of suggests that there is a `newBlockStore` function :confused:

geigerzaehler: Yes, this is a tad unfortunate. This comes out of consistency with `ContentStore` and `HashStore`. I’m open to suggestions.

cloudhead: I guess I'd leave it as-is for now :+1: and we can revisit in the future if we want.

src/Oscoin/Storage/Block/Memory.hs
```diff
@@ -0,0 +1,58 @@
...
+        { getGenesisBlock =
+            List.last <$> getBlocks
+        , lookupBlock = \h ->
+            find (\b -> blockHash b == h) <$> getBlocks
+        , lookupTx = \txHash -> lookupTx txHash <$> getBlocks
+        , getBlocksByDepth = \d ->
+              Chrono.NewestFirst . take (fromIntegral d) <$> getBlocks
+        , getBlocksByParentHash = \h ->
+              Chrono.NewestFirst . takeWhile (\b -> blockHash b /= h) <$> getBlocks
+        , getTip = List.head <$> getBlocks
```
adinapoli-mndc: Do you think we should use a `NonEmpty` list here, to avoid the use of unsafe functions?

src/Oscoin/Storage/Block/Memory.hs
```diff
@@ -0,0 +1,58 @@
...
+        }
+    blockStoreWriter = Abstract.BlockStoreWriter
+        { insertBlock = \b -> modifyBlocks $ \bs -> b:bs
```
adinapoli-mndc: I might be wrong, but `modifyMVar` is not strict neither in the function it applies nor in the content it wraps, which means `b:bs` will leak memory as will build lots of thunks.

We should probably put a bang here, like `\!bs -> b:bs` or contemplate on a bigger scale the use of packages like [strict-concurrency](http://hackage.haskell.org/package/strict-concurrency)

src/Oscoin/Storage/Block/Memory.hs
```diff
@@ -0,0 +1,58 @@
...
+        , switchToFork = \depth newSuffix -> modifyBlocks $ \blks ->
+            toList (Chrono.toNewestFirst $ Chrono.reverse newSuffix) <> drop (fromIntegral depth) blks
```
adinapoli-mndc: Possibly the same malady here, as `<>` will build a thunk.

src/Oscoin/Storage/Block/Memory.hs
geigerzaehler: I replaced `[]` with `Seq` instead of using `NonEmpty`. But I’m requiring a `NonEmpty` blockchain when initializing the blockstore. This should enforce that we never have an empty blockstore.

src/Oscoin/Storage/Block/Memory.hs
kim: Why is it actually using an `MVar`? Can it not use a simple `IORef`? `atomicModifyIORef` (without the tick) is lazy too, though.

src/Oscoin/Storage/Block/Memory.hs
adinapoli-mndc: Right, but a `Seq` can itself be empty, no? Here I am arguing whether or not this is the right contract for this abstraction -- the blockchain must be non-empty by definition, so why not leverage that to get rid of annoying edge cases?

Any problems using `NonEmpty` here? Anything particular that you don't like?

src/Oscoin/Storage/Block/Memory.hs
geigerzaehler: Good point. I’ll switch to IORef

src/Oscoin/Storage/Block/Memory.hs
geigerzaehler: I just want to use `Seq` to get better performance. If you think that is not a concern I’m happy to switch to `NonEmpty`.

adinapoli-mndc: On top of my head I don't think this is going to be a concern, and we can always switch to `Seq` later once we do a performance testing round. Better to start off with sound interfaces and limit total functions, that at least will be my preference 😉

geigerzaehler: I tried to implement it with `NonEmpty`. Turns out that implementing `switchToFork` becomes a bit more involved. I’ll keep it as is (using `Seq`).

PR #468, opened 2019-04-04
Merge blockchain generator modules and test them

geigerzaehler:
Merge, simplify, and document the modules `Oscoin.Test.Crypto.Blockchain.Generator` and `Oscoin.Test.Crypto.Blockchain.Arbitrary`. Also add tests for the blockchain generators.


cloudhead:
Didn't we want to keep generators and arbitrary separate? Don't remember.

geigerzaehler:
> Didn't we want to keep generators and arbitrary separate? Don't remember.

This probably applied to keeping `Arbitrary` orphan instances separate. But here we just have functions that return `Gen`s. This just puts them into one place and uses consistent naming.

test/Test/Oscoin/Crypto/Blockchain/GeneratorsTest.hs
```diff
@@ -0,0 +1,50 @@
...
+import           Oscoin.Consensus
+import qualified Oscoin.Consensus.Nakamoto as Nakamoto
+import           Oscoin.Consensus.Validation
+import           Oscoin.Crypto.Blockchain
+import           Oscoin.Test.Crypto
+import           Oscoin.Test.Crypto.Blockchain.Generators
+import qualified Oscoin.Time.Chrono as Chrono
+
+import           Test.Tasty
+-- import           Test.Tasty.QuickCheck
```
cloudhead: Delete?

geigerzaehler: oops

PR #469, opened 2019-04-05
Refine (global) config

kim:
Amends #459 to align with anticipated deployment scenarios. Addresses the impedance mismatch between "logical" and "physical" network, as pointed out in #462 

## Context

Anticipated deployment scenarios are:

1. `mainnet` - public network, production settings
2. `testnet` - semi-public network, production settings
3. `devnet` - private network, continuous deployment from `master`, development settings
4. ad-hoc - local testing, developer nerd-out / CPU-based housewarming, namespaced deployment of unstable branch, development settings

In order to prevent potentially catastrophic transactions, the addresses for our ledger carry a network bit, such that addresses are not accepted across (logical) networks. This restriction can be lifted for any `devnet`, as nothing is at stake.

## Solution

There is a distinction between a logical network (`Oscoin.Configuration.Network`), and a "physical" network (`Oscoin.P2P.Types.Network`). The former is isomorphic to `Oscoin.Crypto.Address.Internal.AddressTypeTag`, ie. there is exactly one address type for every network. The latter allows for the creation of arbitrarily-named[0] ad-hoc networks, which all map to the logical `devnet`.

Note that the parameter `--network` refers to the physical network. There is also an anti-footgun measure in place, which prompts for confirmation when trying to run in either `mainnet` or `testnet` mode. When the time comes for actual deployment to these networks, this may be enhanced with a non-interactive confirmation (eg. compare the genesis hash to some known value or something).

Lastly, the `Testing` constructor for the `Environment` type is removed due to unclear semantics, and to keep the number of possible permutations down. It might be worth considering to drop the concept of `Environment` entirely in the future: it functions as a kind of "preset" for various settings, which are only meaningful in specific deployment scenarios. Perhaps providing those presets as actual dictionaries might be less surprising in the long run.

## WIP

~~This patch is still wip, as it's lacking documentation and testing.~~ Note that I'm _not_ willing to bikeshed over the naming of `Network`, unless I am provided with an ultrawide monitor supporting enterprise haskell naming (`AbstractConcretePhysicalNetworkImpl`) 👿 

[0]: the name must be a valid domain name label


adinapoli-mndc:
👏

test/Test/Oscoin/P2P/Types.hs
```diff
@@ -10,16 +10,19 @@ import           Oscoin.P2P.Types
...
+                 , renderNetwork
                  )
 
 import           Codec.Serialise (deserialiseOrFail, serialise)
 import           Data.IP (IP(IPv6))
 
 import           Oscoin.Test.Crypto
-import           Test.Oscoin.P2P.Gen (genHost, genHostname, genMsg, genMsgId)
+import           Test.Oscoin.P2P.Gen
+                 (genHost, genHostname, genMsg, genMsgId, genNetwork)
```
kim: I do despise this stylish-haskell rule to the fullest.

src/Oscoin/P2P/Disco/Options.hs
```diff
@@ -0,0 +1,144 @@
...
+
+    evalYesNo x = \case
+        Yes -> pure x
+        No  -> Left "Cancelled"
+
+evalOptions :: Options OptNetwork -> IO (Either String (Options Network))
+evalOptions opt =
+    map (\n -> opt { optNetwork = n }) <$> evalOptNetwork (optNetwork opt)
+
+-- | Interactively obtain a 'YedNo' value.
```
cloudhead:
```suggestion
-- | Interactively obtain a 'YesNo' value.
```

src/Oscoin/Node/Trans.hs
```diff
@@ -48,12 +50,18 @@ runNodeT env (NodeT ma) = runReaderT ma env
...
     , cfgNoEmptyBlocks   :: Bool
     , cfgConsensusConfig :: Consensus.Config
     }
 
+data GlobalConfig = GlobalConfig
+    { globalEnv             :: Environment
+    , globalLogicalNetwork  :: Network
+    , globalPhysicalNetwork :: P2P.Network
+    }
+
```
cloudhead: Any reason why we have a record for this, vs. having the fields directly in `Config`? What is it that makes this config 'global' and the other not? Is it because it's network-wide?

cloudhead: Perhaps this can be answered in a haddock :man_shrugging:

kim: idk, can't decide where to put this - the types are defined in the "global" `Oscoin.Configuration` module, and I expect them to be threaded to other places in the code (irrespective of the `Node`'s config).

src/Oscoin/P2P/Types.hs
```diff
@@ -50,19 +59,90 @@ import qualified Codec.Serialise as CBOR
...
+readNetworkText :: Text -> Either String Network
+readNetworkText "mainnet" = pure Mainnet
+readNetworkText "testnet" = pure Testnet
+readNetworkText "devnet"  = pure Devnet
+readNetworkText t
+  | T.length t > 63         = Left "Network name longer than 63 characters"
+  | Just c <- invalidChar t = Left $ "Invalid character in network name: " <> [c]
+  | otherwise               = Right $ Somenet' t
+  where
+    invalidChar = T.find (`elem` ("./:" :: String))
```
geigerzaehler: It is unclear what a “valid domain name label” is. This code seem to imply that it is anything that does not contain the characters `./:`. The tests however generate only alphanumeric network names.

I’d suggest to document what a valid network name is explicitly and make the test cover the full range.. I’d prefer to use just alphanumeric names (including dashes).

src/Oscoin/Node/Trans.hs
kim: It _should_ probably be in `Oscoin.Configuration` with ctor hidden as not all combinations make sense / should be allowed. Can't import the P2P types though, because cycle.

PR #470, opened 2019-04-08
Require matching network during handshake, as an additional sanity check

kim:
On top of #469


src/Oscoin/P2P/Handshake/Simple.hs
```diff
@@ -26,10 +26,10 @@ import           Data.ByteArray.Orphans ()
 
 
 data HandshakeMessage c =
-      Hai (Crypto.PublicKey c)
-    -- ^ Present a pubkey and random value.
+      Hai (Crypto.PublicKey c) Network
+    -- ^ Present a pubkey and network name.
     | Bai SimpleError
-    -- ^ At any point, either side may bail oout.
+    -- ^ At any point, either side may bail out.
```
cloudhead: :woman_shrugging:

PR #471, opened 2019-04-08
Callstacks for debug and development prelude functions

geigerzaehler:
We update `Oscoin.Prelude` to allow for a better development experience when using `notImplemented` and `undefined`.

During development `notImplemented` and `undefined` are often used to leave holes in the code. When running tests against code that contains multiple `notImplemented` statements we want to know which of these threw an error. For this we add the `HasCallstack` constraint to `notImplemented` and `undefined`.

We also start exporting `Protolude.error` from `Oscoin.Prelude` so that we can put some more information into errors during development.

### DIFF
--- a/src/Oscoin/Prelude.hs
+++ b/src/Oscoin/Prelude.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE NoImplicitPrelude #-}
+{-# OPTIONS_GHC -fno-warn-warnings-deprecations #-}
 
 -- | Custom Prelude based on <https://hackage.haskell.org/package/protolude Protolude>.
 --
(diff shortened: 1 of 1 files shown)


kim:
I don't understand why it is necessary to redefine `notImplemented` and `undefined`, afaics the [protolude versions](https://hackage.haskell.org/package/protolude-0.2.3/docs/src/Debug.html) are defined in terms of [Protolude.error](https://hackage.haskell.org/package/protolude-0.2.3/docs/src/Protolude.Error.html#error), which in turn does include a CallStack.

geigerzaehler:
I had to amend this even further: `Protolude.Error.error` does not provide call stacks at all and I don’t know why. But you can easily reproduce it with GHCi.

```
> Protolude.Error.error "foo"
*** Exception: foo
> Prelude.error "foo"
*** Exception: foo
CallStack (from HasCallStack):
  error, called at <interactive>:2:1 in interactive:Ghci1
```

geigerzaehler:
See https://github.com/sdiehl/protolude/pull/102

kim:
:picard:

adinapoli-mndc:
:double-picard:

adinapoli-mndc:
@geigerzaehler I am not opposed to this PR, but me just wonders -- would it make sense to just wait for the change in protolude to be effective and mark this PR moot? Although, I have to say, I like the `notImplemented` thing, even just as a support during development 🤔

geigerzaehler:
I added documentation that we should switch to the upstream version again when the fix is released. I’ll monitor the package. Judging from previous releases it is going to take a while until the fixed version is released.

PR #477, opened 2019-04-09
Make 'Signed' self-contained

cloudhead:
This makes a few changes in preparation for #436 :
* Instead of passing a `PublicKey` around for signature verification, we include it in `Signed`, so that it is self-contained. This is the first step towards recovering the PublicKey from the signature (see: https://crypto.stackexchange.com/questions/18105/how-does-recovering-the-public-key-from-an-ecdsa-signature-work for how this works). This will save us a bunch of space in the blockchain, instead of having to always include the PublicKey.
* We make some improvements to the mock crypto to actually check that the signature belongs to the signed message.
* We implement a `derivePublicKey` which derives a pubkey from a secret key.
* We remove PublicKey from `Tx`, since it is no longer needed there.

The only thing I'm unsure of are the changes I had to make to the P2P verification stuff to make the tests pass, which don't feel *quite* right - maybe @kim can shed some light.


kim:
I don't think this is the right move: we're now sending the pk over the wire with _every_ signed message, which is not only wasteful, but pretty much defeats the purpose of validating a signature (you can send me a different pk every time at your convenience).

If the goal is just to save some space in the transactions table, it seems to me like it is sufficient to remove the `author` field, and defer implementing the key recovery to when we actually would want to validate persistent transactions (note Section 4.1.6 of https://www.secg.org/sec1-v2.pdf, and [this comment](https://crypto.stackexchange.com/questions/18105/how-does-recovering-the-public-key-from-an-ecdsa-signature-work#comment40258_18138))

geigerzaehler:
The cited Stackoverflow answer says that

> Actually, it is not possible to uniquely recover the public key from an ECDSA signature.

There are two keys that can be derived from a signature. This means that when we evaluate a transaction we don’t know what account hash of the sender is. It could be either of the two.

I also second @kim's concerns about validating signatures. Removing the `PubKey` parameter from `verify` prevents us from authenticating messages for an already known key.

src/Oscoin/Crypto/PubKey.hs
```diff
@@ -49,33 +51,37 @@ class HasDigitalSignature c where
...
 
     sign   :: (ByteArrayAccess msg, MonadRandom m) => PrivateKey c -> msg -> m (Signed c msg)
-    verify :: ByteArrayAccess msg => PublicKey c -> Signed c msg -> Bool
+    verify :: ByteArrayAccess msg => Signed c msg -> Bool
 
     -- | Generate a new random keypair.
     generateKeyPair :: (Crypto.HasHashing c, MonadRandom m) => m (KeyPair c)
 
+    -- | Derive the public key of a private key.
+    derivePublicKey :: PrivateKey c -> PublicKey c
```
kim: This seems generally useful, however we don't need `generateKeyPair` as a class method anymore. It should be replaced by `generateSecretKey`, and a top-level convenience function `generateKeyPair` could be provided which is implemented in terms of `generateSecretKey` and `derivePublicKey`.

cloudhead:
> There are two keys that can be derived from a signature. This means that when we evaluate a transaction we don’t know what account hash of the sender is. It could be either of the two.

Yeah, you have to add a byte to the signature to determine which of the two it is.

src/Oscoin/Crypto/PubKey.hs
cloudhead: Good point!

cloudhead:
> I don't think this is the right move: we're now sending the pk over the wire with _every_ signed message, which is not only wasteful, but pretty much defeats the purpose of validating a signature (you can send me a different pk every time at your convenience).
> 
Indeed, perhaps the right way is to have a `SignedTx` type that is only used for transactions, and not for all signed messages.

> If the goal is just to save some space in the transactions table, it seems to me like it is sufficient to remove the `author` field, and defer implementing the key recovery to when we actually would want to validate persistent transactions (note Section 4.1.6 of https://www.secg.org/sec1-v2.pdf, and [this comment](https://crypto.stackexchange.com/questions/18105/how-does-recovering-the-public-key-from-an-ecdsa-signature-work#comment40258_18138))

The goal was more to save space in blocks on the wire, which (without public key recovery) need to store the full public key in each transaction.

src/Oscoin/Crypto/PubKey.hs
cloudhead: Hmm annoyingly, cryptonite doesn't expose a `generatePrivateKey`, meaning our implementation of it would have to generate a pair and discard the public key (which maybe will never be computed due to laziness :thinking: )

kim:
I see. Seems to call for spec of how and when signatures are actually verified. Announcing your public key to verify a subsequent signed message should always be subject to a challenge, though (I'm working on supporting noise to talk to the API a bit on the side, so that should be covered).

cloudhead:
Going to close this, thanks for the feedback. I may extract some of it into a different proposal when needed, but I'm going to approach it in a different way.

PR #479, opened 2019-04-10
Cleanup 'Oscoin.Crypto.Blockchain.Block'

geigerzaehler:
This commit includes changes that improve the code in the `Oscoin.Crypto.Blockchain.Block` module.

We eliminate `genesisBlock`. This function was only used in one place where it was easily replaced by `emptyGenesisFromState`.

We eliminate the unused `withHeader` function.

We rewrite `linkParent` and `sealBlock` using lenses. This simplifies the code an reduces duplication. But more importantly we now guarantee the correctness of the `blockHash` field by using `blockHeaderL` instead of relying on us setting the field manually.

### DIFF
--- a/src/Oscoin/Crypto/Blockchain/Block.hs
+++ b/src/Oscoin/Crypto/Blockchain/Block.hs
@@ -23,11 +23,9 @@ module Oscoin.Crypto.Blockchain.Block
     , blockData
     , emptyGenesisBlock
     , emptyGenesisFromState
-    , genesisBlock
     , isGenesisBlock
     , headerHash
     , parentHash
-    , withHeader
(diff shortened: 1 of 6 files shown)


adinapoli-mndc:
> May I suggest to take a look at generic-lens? It's pretty awesome.

Seconded. Never used it but I have seen a presentation at the last ZuriHac and it's quite an impressive project (especially as they managed to optimise stuff away in a way that the generate GHC.Core is the same as if you were using plain lenses).

geigerzaehler:
> May I suggest to take a look at `generic-lens`? It's pretty awesome.

I’d be up to using it too if we all agree that it is a good idea.

PR #482, opened 2019-04-10
Allow tweaking the time between empty blocks

kim:
1 second is often too noisy. Cannot be set for `Production`, obviously.

### DIFF
--- a/src/Oscoin/Consensus/Nakamoto.hs
+++ b/src/Oscoin/Consensus/Nakamoto.hs
@@ -37,7 +37,7 @@ import           Database.SQLite.Simple.ToField (ToField)
 
 -- | Target block time.
 blockTime :: Duration
-blockTime = 1 * seconds
+blockTime = 1 * minutes
 
 -- | The default target difficulty at genesis.
 defaultGenesisDifficulty :: Difficulty
--- a/src/Oscoin/Node.hs
+++ b/src/Oscoin/Node.hs
@@ -45,7 +45,6 @@ import           Oscoin.Data.Query
 import qualified Oscoin.Data.RadicleTx as RadicleTx
 import           Oscoin.Node.Mempool (Mempool)
 import qualified Oscoin.Node.Mempool as Mempool
-import qualified Oscoin.Node.Mempool.Class as Mempool
 import           Oscoin.Node.Trans
 import qualified Oscoin.P2P as P2P
 import           Oscoin.Storage (Storage(..))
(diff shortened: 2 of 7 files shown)


src/Oscoin/Node.hs
```diff
@@ -127,23 +130,39 @@ miner
...
                 liftIO $ tele (BlockMinedEvent bhash)
                 lift (tryBroadcast b) >>= liftIO . tele . \case
                     Left  e  -> BlockBroadcastFailedEvent bhash (toException e)
                     Right () -> BlockBroadcastEvent bhash
+
+        done  <- liftIO $ getTime Monotonic
+        let
+            !delta     = fromInteger . toNanoSecs $ diffTimeSpec start done
+            isProd     = Production == env
+            shouldWait = nTxs == 0 && not isProd && delta < emptyBlockTime
```
kim: Maybe it could apply regardless of the number of txs for this block, idk

cloudhead:
It would make more sense if this was adjusting the `blockTime` (https://github.com/oscoin/oscoin/blob/master/src/Oscoin/Consensus/Nakamoto.hs#L39) as a consensus parameter, since that is exactly the purpose of it - otherwise we have two notions of "block time"

cloudhead:
Alternatively, you can just update `blockTime` to `1 * minute` :joy: since that is the actual block time we're aiming for. This will work so long as our difficulty adjustment code works :crossed_fingers:

kim:
If that only had the same effect on avoiding useless CPU cycles lol. Open to suggestions for alternative naming to avoid clashing with jargon.

cloudhead:
Haha true. Maybe a better approach then would be to have a miner function which has a configurable delay in it? I'm reluctant to have debug/testing logic such as this in the main code path :thinking: - it would be nice if we could rather do something like `nakamotoConsensus { miner = delayedNakamotoMiner (1000 * seconds) }`, and have that isolated there.

kim:
Hm yeah, we can move this to the `Lenient` module

src/Oscoin/Node/Options.hs
```diff
@@ -58,6 +60,14 @@ nodeOptionsParser cps = Options
...
        <> showDefault
         )
+    <*> option (map (* 1000000000) auto)
+        ( long "empty-block-time"
+       <> help "Wait approx. this many seconds between empty blocks. Has no \
+               \effect when running with production settings."
+       <> metavar "SECONDS"
+       <> value 1
+       <> showDefault
+        )
```
cloudhead: Should we prefix this option to communicate that it affects the lenient consensus? Eg. `lenient.empty-block-time` or something?

cloudhead: We can also punt on this and revisit our whole cli options scheme at some point in the future :man_shrugging:

kim:
Ait, it seems to me that consensus-affecting parameters are a bit of a mess at the moment, and it is easy to adjust one but miss out on others (and end up with non-meaningful permutations). I'm leaving this to someone else to consolidate in the future, including the decision if it should be possible to adjust the configuration from the outside (probably not).

Anyways, I'm amending this as follows:

* The target Nakamoto `blockTime` is set to 10 minutes (which seems to be the expectation elsewhere)
* `--no-empty-blocks` is removed, because correctness
* When running in `Development`, the (actual) block time can be tweaked using `--block-time-lower`, which imposes a lower bound. Ie. if there are no transactions to go into the block, the (lenient) miner waits for `min blockTimeLower blockTime` before actually mining the block. This avoids busy looping, but also control over how long to wait for a new block during development. Note that the main motivation here is the `devnet`, however, which will be idle most of the time, but shouldn't waste CPU (which is actually something cloud providers punish you for).

You may cast your vote for alternative names of the parameter here:

* `--hodgepodge-higgeldy-piggeldy`
* `--wait-mine-empty-block-seconds`
* `--lenient-empty-block-wait-seconds`
* `--seconds-before-mining-empty-block-in-development-environment`
* `--wat-gimme-a-break`

xla:
`--lemme-save-some-trees`

`--abiding-cloud-citizen`

src/Oscoin/Consensus/Nakamoto.hs
```diff
@@ -37,7 +37,7 @@ import           Database.SQLite.Simple.ToField (ToField)
 
 -- | Target block time.
 blockTime :: Duration
-blockTime = 1 * seconds
+blockTime = 10 * minutes
```
cloudhead: The target for the mainnet will likely be `1 * minutes` - that's what we've been modeling our supply curves from, with `60 osc` per block. This gives you a nice and intuitive `1 osc / second` emission rate.

cloudhead:
This does simplify things! I'm still in favor of `--lenient.*` as a namespace, but why not just have `--lenient.block-time`, and have it just be a way to overwrite the block-time? Ie. instead of using `min`, just have it be an override?

kim:
> but why not just have --lenient.block-time, and have it just be a way to overwrite the block-time?

Here's why:

> I'm leaving this to someone else to consolidate in the future, including the decision if it should be possible to adjust the configuration from the outside (probably not).

PR #483, opened 2019-04-10
[Experiment] Exploration of ways to replace radicle

adinapoli-mndc:
@cloudhead @kim @geigerzaehler as I am off shortly for the rest of the week, I wanted to seize the chance to take stock and plant the seed in each of your brains on how do we approach dropping `radicle` from our stack.

What you see here is a half-baked, fully-broken-wont-compile PR that serves just as a discussion kickstarter. The problem we have is that the use of `RadTx c` is quite pervasive in the codebase; not only that, but we use a concrete `tx` implementation in some `src` modules (so not just in the application code, but also in library code). This has a number of shortcomings:

* It doesn't make library code reusable (for example the HTTP stuff, which is usable only with a concrete `RadTx` type);

* It makes replacing one implementation with the other quite a heavy task.

Yes, we do have in a lot of places `tx` being a type parameter (which allows for code to be written in a generic way), but the problem is that we don't do this _everywhere_ in the library code.

For example:

```haskell
data Handle c tx st s i = Handle
    { hConfig       :: Config
    , hNodeId       :: i
    , hStateStore   :: HashStore c st IO
    , hBlockStore   :: Abstract.BlockStoreReader c tx s IO
    , hProtocol     :: Protocol.Handle c tx s IO
    , hMempool      :: Mempool.Handle c tx
    , hEval         :: Evaluator st tx RadicleTx.Message
    , hConsensus    :: Consensus c tx s (NodeT c tx st s i IO)
    , hReceiptStore :: ReceiptStore c tx RadicleTx.Output IO
    }
```

As you can see, the problem are the `Evaluator` and the `ReceiptStore`. We _could_ , in principle, make those type parameters, but the type signature for `Handle` is already a bit intimidating. Therefore, something I have been playing with was the possibility to have the following typeclass:

```haskell
class IsTx c tx where
    type family TxOutput c tx  = txout | txout -> tx
    type family TxPayload c tx = payload | payload -> tx

-- | A concrete Oscoin transaction.
-- N.B. this type and the inner payload is just stubbed.
type OscoinTx c = Types.Tx c (Payload c)

-- | A not-very-interesting 'TxPayload'.
newtype Payload c = Payload ByteString deriving (Serialise, ByteArrayAccess)

instance ToJSON (Payload c) where
    toJSON = serialiseToJSON

instance FromJSON (Payload c) where
    parseJSON = deserialiseParseJSON

type Output  c = Int

instance IsTx c (OscoinTx c) where
    type TxPayload c (OscoinTx c) = Payload c
    type TxOutput  c (OscoinTx c) = Output c

eval :: Hashable c (PublicKey c) => Evaluator () (OscoinTx c) (Output c)
eval _ _ = pure (42, initialState)

initialState :: ()
initialState = ()

```

So that you could write:

```haskell
data Handle c tx st s i = Handle
    { hConfig       :: Config
    , hNodeId       :: i
    , hStateStore   :: HashStore c st IO
    , hBlockStore   :: Abstract.BlockStoreReader c tx s IO
    , hProtocol     :: Protocol.Handle c tx s IO
    , hMempool      :: Mempool.Handle c tx
    , hEval         :: Evaluator st tx (OscoinTx.TxOutput tx)
    , hConsensus    :: Consensus c tx s (NodeT c tx st s i IO)
    , hReceiptStore :: ReceiptStore c tx (OscoinTx.TxOutput tx) IO
    }
```

Now, while this approach might work, I am not totally convinced it's worth it. A number of open questions I cannot answer (just yet):

- How invasive is this gonna be? We will probably need to add `IsTx` in quite a few places;
- Is it really true that fixing the `tx` fixes the output of the receipt store and the evaluator? How are they related?
- What about the CLI stuff? We have things like `readRadFile` which I am not sure how they will play out in this schema (or any other schema);
- When is the right moment to get concrete? Ideally in application code, but you will see that this gets invasive quite rapidly, as for the HTTP code if you want to be 100% polymorphic you would need to add an extra `tx` type variable to a bunch of types (including the `ApiAction`).

In a nutshell, I do wonder if this is really worth it. If we think a transaction is something we are going to swap again the future and that our work with the smart contract are going to make them more complicated, it might be a good time to reflect how we go with that.

If we thing this is a fool's errand, probably the best course of action is to park this up until @cloudhead finishes his work on the new Tx type, which hopefully might kick Radicle out of the codebase.

Thoughts?


kim:
> Is it [...] true that fixing the tx fixes the output of the receipt store and the evaluator

For posterity: this is the main question for me, really. It does _not_ seem to me as if the _type_ of `tx` determines the type of output, at least not in principle. If that is the case, there is not much choice but to thread the polymorphic `txout` wherever there is a polymorphic `tx`.

geigerzaehler:
I really like this approach!

> Is it really true that fixing the `tx` fixes the output of the receipt store and the evaluator? How are they related?

I think it is very true that `tx` fixes the output. In fact, `tx` also fixes the evaluator and the state so I’d argue that we should extend the class with theses.

Is the `TxPayload` type strcitly necessary? Is there a function with an `IsTx` contraint that will use it?

> In a nutshell, I do wonder if this is really worth it. If we think a transaction is something we are going to swap again the future and that our work with the smart contract are going to make them more complicated, it might be a good time to reflect how we go with that.

I don’t quite quite get your second sentence but I’d like to point out that it is not only about being able to swap the tx out but the class will also provide abstraction and separation between the domain logic and the rest of the system. Therefore I think this is very much worth it.

adinapoli-mndc:
> For posterity: this is the main question for me, really. It does not seem to me as if the type of tx determines the type of output, at least not in principle. If that is the case, there is not much choice but to thread the polymorphic txout wherever there is a polymorphic tx.

Right, I guess this is one of those things you wouldn't know even after you start extending the system, so there is always a bit of danger of taking a wrong turn there, even though @geigerzaehler 's comment is reassuring. 

> Is the TxPayload type strcitly necessary? Is there a function with an IsTx contraint that will use it?

That's a good question -- the original reason was because I _did_ use it in the `Handle`. In particular, if you observe the original definition of `Handle`:

```haskell
data Handle c tx st s i = Handle
    { hConfig       :: Config
    , hNodeId       :: i
    , hStateStore   :: HashStore c st IO
    , hBlockStore   :: Abstract.BlockStoreReader c tx s IO
    , hProtocol     :: Protocol.Handle c tx s IO
    , hMempool      :: Mempool.Handle c tx
    , hEval         :: Evaluator st tx RadicleTx.Message
    , hConsensus    :: Consensus c tx s (NodeT c tx st s i IO)
    , hReceiptStore :: ReceiptStore c tx RadicleTx.Output IO
    }
```

You will see that `Evaluator` used the `RadicleTx.Message` (which is conceptually what my `TxPayload` was meant to replace) whereas the `ReceiptStore` used the `RadicleTx.Output`). However, I think there is a silent bug in the current `master`: if you observe the call to `withNode`:

```haskell
withNode
    :: (Log.Buildable (Hash c), HasHashing c)
    => Config
    -> i
    -> Mempool.Handle c tx
    -> HashStore c st IO
    -> BlockStoreReader c tx s IO
    -> Protocol.Handle c tx s IO
    -> Evaluator st tx RadicleTx.Output
    -> Consensus c tx s (NodeT c tx st s i IO)
    -> (Handle c tx st s i -> IO a)
    -> IO a
withNode hConfig hNodeId hMempool hStateStore hBlockStore hProtocol hEval hConsensus =
[...]
```

This now wants an `Evaluator` that yields a `RadicleTx.Output`, and obviously `RadicleTx.Output` and `RadicleTx.Message` unify _only because they are aliased to the same type_, which is a `Rad.Value`. So either:

1. This is a genuine bug;
2. I didn't understand the code and indeed the `Evaluator` output can be changed/hoisted and `withNode` is not using the stuff from the `Handle`;

Either way, I did fix the unification problem by changing `RadicleTx.Message` into `TxOutput` into the definition of `Handle`, and that effectively made `TxPayload` used, even though it's still unclear whether or not we need it 😁

cloudhead:
I like this approach. I wouldn't complicate it further, and I'd keep any concrete `RadicleTx` concrete, eg. in the HTTP API.

> This now wants an Evaluator that yields a RadicleTx.Output, and obviously RadicleTx.Output and RadicleTx.Message unify only because they are aliased to the same type, which is a Rad.Value. So either:

This is by design, since the output of a transaction is just another radicle value. When removing radicle though, I suspect that the following will make more sense:

```haskell
class IsTx c tx where
    type family TxOutput c tx  = txout | txout -> tx
    type family TxState c tx = st | st -> tx
```

Then the evaluator is `Evaluator (TxState tx) tx (TxOutput tx)`

adinapoli-mndc:
Let me close this and open another PR as rebasing would have been a nightmare.

PR #484, opened 2019-04-10
Update 'evalBlock' to make it more powerful

geigerzaehler:
`evalBlock` is updated to produce the new state and the receipts from block transactions. This functionality will be used by the Ledger abstraction in the future.

We also add tests for `evalBlock` and move the tests for `buildBlock` into the same module.

Finally, we eliminate the unused `evalBlockchain` functions.

### DIFF
--- a/exe/node/Main.hs
+++ b/exe/node/Main.hs
@@ -10,7 +10,7 @@ import qualified Oscoin.Consensus.Config as Consensus
 import qualified Oscoin.Consensus.Nakamoto as Nakamoto
 import           Oscoin.Crypto (Crypto)
 import           Oscoin.Crypto.Blockchain.Block (Block, Sealed)
-import           Oscoin.Crypto.Blockchain.Eval (evalBlock, fromEvalError)
+import           Oscoin.Crypto.Blockchain.Eval (evalBlock)
 import           Oscoin.Data.RadicleTx (RadTx)
 import qualified Oscoin.Data.RadicleTx as Rad (pureEnv, txEval)
 import           Oscoin.Node (runNodeT, withNode)
@@ -57,8 +57,7 @@ main = do
     nid          <- pure (mkNodeId $ fst keys)
     mem          <- Mempool.newIO
     gen          <- Yaml.decodeFileThrow (genesisPath optPaths) :: IO GenesisBlock
-    genState     <- either (die . fromEvalError) pure $
-                        evalBlock Rad.txEval Rad.pureEnv gen
+    let (genState, _receipts) = evalBlock Rad.txEval Rad.pureEnv gen
     stStore      <- newHashStoreIO
     storeHashContent stStore genState
     metricsStore <-
--- a/src/Oscoin/Crypto/Blockchain/Eval.hs
+++ b/src/Oscoin/Crypto/Blockchain/Eval.hs
@@ -9,17 +9,14 @@ module Oscoin.Crypto.Blockchain.Eval
     , buildBlockStrict
     , buildGenesis
     , evalBlock
-    , evalBlockchain
     ) where
 
 import           Oscoin.Prelude
 
 
-import           Oscoin.Crypto.Blockchain (Blockchain(..), blocks)
 import           Oscoin.Crypto.Blockchain.Block hiding (parentHash)
 import           Oscoin.Crypto.Hash (Hash)
 import qualified Oscoin.Crypto.Hash as Crypto
-import           Oscoin.Time.Chrono (toNewestFirst)
 
 import           Codec.Serialise (Serialise)
 import qualified Crypto.Data.Auth.Tree.Class as AuthTree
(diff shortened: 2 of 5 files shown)


src/Oscoin/Crypto/Blockchain/Eval.hs
```diff
@@ -159,32 +156,27 @@ buildGenesis eval tick txs s =
...
-
--- | Evaluate all transactions of a 'Blockchain' and return the result.
-evalBlockchain
-    :: Evaluator st tx o
-    -> st
-    -> Blockchain c tx s
-    -> ([(tx, Either EvalError o)], st)
-evalBlockchain eval st (toNewestFirst . blocks -> blks) =
-    evalTraverse eval (concatMap (toList . blockData) (reverse blks)) st
+    -> (st, [Receipt c tx o])
```
kim: It is a bit obscure that any `EvalError`s are buried in the `Receipt`s 🤔

cloudhead: Indeed - maybe making this more explicit by having it be a composition of two functions, like `intoReceipt . eval`

geigerzaehler: We don’t return information about evaluation errors explicitly because blocks are valid even if a tx produces an eval error. This is why the `receiptTxOutput` field of a receipt has type `Either EvalError output` and not just `output`. If a transaction in the block produces an error we leave the state untouched and include the error in the receipt. But the transaction is a “valid” part of the blockchain.

However there’s a good case to be made that blocks should only contain valid transactions which would mean that `receiptTxOutput` just get’s type `output` and `evalBlock` would return `Either EvalError ...`. @cloudhead and I discussed this briefly and we are leaning towards this variant. Making this change requires a concerted effort though and should be done in another PR.

@cloudhead: Could you elaborate on your suggestion? I don’t get what you mean.

cloudhead: I just mean that the `EvalError` is "swallowed" in the current implementation and not visible in the types, so having a more explicit `Either EvalError Output -> Receipt` step would surface the mechanism and be self-documenting.

kim: I'm not sure: is an empty block actually valid? Also, what happens to transactions which fail to evaluate, are they discarded?

Intuitively, I'd expect a signature like this:

```haskell
evalBlock 
    :: Hashable c tx
    => Evaluator st tx o
    -> st
    -> Block c tx s
    -> (st, Validation (NonEmpty EvalError) ([(EvalError, tx)], NonEmpty (Receipt c tx o))
```

ie. either a non-empty list of `EvalError`s if none of the transactions could be evaluated, or a pair of a possibly empty list of `EvalError`s along with the transactions which failed to evaluate, and a non-empty list of valid receipts. This type could be wrapped up in a `EvalResult` sum.

geigerzaehler: > I'm not sure: is an empty block actually valid?

Depends on what you consider as valid. An empty block is accepted by `evalBlock`. Nakamoto consensus also needs empty blocks to work if the network does not produce transactions.

> I just mean that the `EvalError` is "swallowed" in the current implementation and not visible in the types, so having a more explicit `Either EvalError Output -> Receipt` step would surface the mechanism and be self-documenting.

This is just `evalTraverse` which is used by `evalBlock`.

To clarify I’m going to give an example. Assume a block has three transactions _tx₁_, _tx₂_, _tx₃_. Applying _tx₁_ to the start state _s₀_ produces _s₁_,  _tx₂_, fails and produces an error and _tx₃_ applied to  _s₁_ gives _s₂_. Then the block evaluates to the state _s₂_ and produces three receipts. The receipt for _tx₁_ includes the output of that transaction, the receipt for _tx₂_ includes an error and the receipt for _tx₃_ again holds the output of the transaction. All transactions are valid parts of the block and can be evaluated.

geigerzaehler: I updated the tests to make this clearer

kim: > Nakamoto consensus also needs empty blocks to work if the network does not produce transactions.

Fair enough. I guess partitioning the receipts list to find erroneous transactions is fine then.

exe/node/Main.hs
```diff
@@ -57,8 +57,7 @@ main = do
     nid          <- pure (mkNodeId $ fst keys)
     mem          <- Mempool.newIO
     gen          <- Yaml.decodeFileThrow (genesisPath optPaths) :: IO GenesisBlock
-    genState     <- either (die . fromEvalError) pure $
-                        evalBlock Rad.txEval Rad.pureEnv gen
+    let (genState, _receipts) = evalBlock Rad.txEval Rad.pureEnv gen
```
kim: On a second thought: it doesn't seem wise to ignore evaluation errors for the genesis - there is no guarantee that the file at the given path decodes to a valid block.

geigerzaehler: A genesis block can be valid even if it contains transactions that produce evaluation errors. In that respect it is no different than any other block.

I agree however that we also need to validate the genesis block, in particular that the state hash matches the produced state. With #486 this code here will be moved to the ledger which will take care of that validation.

kim:
> it is no different than any other block.

No

exe/node/Main.hs
cloudhead: :thinking: 
Independent of this PR, since we basically decide what goes in genesis though, we should probably validate that it has no evaluation errors.

PR #486, opened 2019-04-11
Introduce 'Ledger' for Node, Miner, and tests

geigerzaehler:
We introduce the `Ledger` abstraction which wraps the block store, state store, and receipt store. The purpose of the `Ledger` is to provide consistent access to block artifacts (states and receipts). This commit only integrates the ledger with `Oscoin.Node`, `Oscoin.Consensus.Mining`, and the `Oscoin.Test.Consensus.Nakamoto`. It remains to integrate the ledger with `Protocol` where it will replace the `BlockStore`.

 Todo

 - [x] ~~Test and document `Ledger` constructors~~ Will do in a follow-up PR

### DIFF
--- a/src/Oscoin/Node.hs
+++ b/src/Oscoin/Node.hs
@@ -29,17 +29,8 @@ import           Oscoin.Consensus (Consensus(..), ValidationError)
 import qualified Oscoin.Consensus as Consensus
 import           Oscoin.Crypto.Blockchain (TxLookup)
 import           Oscoin.Crypto.Blockchain.Block
-                 ( Block(..)
-                 , BlockHash
-                 , BlockHeader(..)
-                 , Depth
-                 , Sealed
-                 , StateHash
-                 , blockHash
-                 )
-import           Oscoin.Crypto.Blockchain.Eval (Evaluator, Receipt)
-import           Oscoin.Crypto.Hash
-                 (HasHashing, Hash, Hashable, Hashed, formatHash)
+import           Oscoin.Crypto.Blockchain.Eval (Receipt)
+import           Oscoin.Crypto.Hash (Hash, Hashable, Hashed, formatHash)
 import qualified Oscoin.Crypto.Hash as Crypto
 import           Oscoin.Data.Query
 import qualified Oscoin.Data.RadicleTx as RadicleTx
@@ -49,7 +40,7 @@ import           Oscoin.Node.Trans
 import qualified Oscoin.P2P as P2P
 import           Oscoin.Storage (Storage(..))
 import qualified Oscoin.Storage as Storage
-import           Oscoin.Storage.HashStore
+import qualified Oscoin.Storage.Ledger as Ledger
 import           Oscoin.Telemetry (NotableEvent(..))
 import qualified Oscoin.Telemetry as Telemetry
 import           Oscoin.Telemetry.Logging (ftag, stext, (%))
@@ -57,10 +48,7 @@ import qualified Oscoin.Telemetry.Logging as Log
 import qualified Oscoin.Time.Chrono as Chrono
 
 import qualified Oscoin.Protocol as Protocol
-import           Oscoin.Storage.Block.Abstract
-                 (BlockStoreReader, hoistBlockStoreReader)
 import qualified Oscoin.Storage.Block.Abstract as BlockStore
-import qualified Oscoin.Storage.Receipt as ReceiptStore
 
 import           Codec.Serialise
 import           Control.Monad.IO.Class (MonadIO(..))
@@ -70,23 +58,20 @@ import qualified Crypto.Data.Auth.Tree.Class as AuthTree
 import           Lens.Micro ((^.))
 
 withNode
-    :: (Log.Buildable (Hash c), HasHashing c)
+    :: (Log.Buildable (Hash c))
     => Config
     -> i
     -> Mempool.Handle c tx
-    -> HashStore c st IO
-    -> BlockStoreReader c tx s IO
+    -> Ledger.Ledger c s tx RadicleTx.Message st IO
     -> Protocol.Handle c tx s IO
-    -> Evaluator st tx RadicleTx.Output
     -> Consensus c tx s (NodeT c tx st s i IO)
     -> (Handle c tx st s i -> IO a)
     -> IO a
-withNode hConfig hNodeId hMempool hStateStore hBlockStore hProtocol hEval hConsensus =
+withNode hConfig hNodeId hMempool hLedger hProtocol hConsensus =
     bracket open close
   where
     open = do
-        hReceiptStore <- ReceiptStore.newReceiptStoreIO
-        gen <- liftIO (BlockStore.getGenesisBlock hBlockStore)
+        gen <- liftIO (BlockStore.getGenesisBlock $ Ledger.blockStoreReader hLedger)
         let
             GlobalConfig { globalEnv             = env
                          , globalLogicalNetwork  = lnet
@@ -169,26 +154,19 @@ mineBlock
        )
     => NodeT c tx st s i m (Maybe (Block c tx (Sealed c s)))
 mineBlock = do
-    Handle{hEval, hConsensus, hProtocol} <- ask
-    stateStore <- getStateStore
+    Handle{hConsensus, hProtocol} <- ask
     time <- currentTick
-    bs <- getBlockStoreReader
+    ledger <- getLedger
     maybeBlock <- hoist liftIO $ Consensus.mineBlock
-        bs
-        stateStore
+        ledger
         hConsensus
-        hEval
         time
     -- NOTE(adn) Here we should dispatch the block and wait for the
     -- result: if the block hasn't been inserted (for example due to
     -- a validation error) we shouldn't proceed with all these other
     -- side effects. Ideally 'dispatchBlockSync' should return some kind
     -- of 'Either Error ()'.
-    forM maybeBlock $ \(blk, st, receipts) -> do
-        stateStore' <- getStateStore
-        storeHashContent stateStore' st
-        receiptStore <- getReceiptStore
-        for_ receipts $ ReceiptStore.storeReceipt receiptStore
+    forM maybeBlock $ \blk -> do
         liftIO $ Protocol.dispatchBlockSync hProtocol blk
         pure blk
 
@@ -231,23 +209,23 @@ getPath
     -> [Text]
     -> NodeT c tx st s i m (Maybe (QueryVal st))
 getPath stateHash p = do
-    stateStore <- getStateStore
-    result <- lookupHashContent stateStore (Crypto.toHashed stateHash)
+    ledger <- getLedger
+    result <- Ledger.lookupState ledger (Crypto.toHashed stateHash)
     pure $ query p =<< result
 
 -- | Get a value from the latest state.
 getPathLatest
     :: ( MonadIO m
        , Query st
+       , Crypto.Hashable c tx
        )
     => [Text]
-    -> NodeT c tx st s i m (Maybe (StateHash c, QueryVal st))
+    -> NodeT c tx st s i m (Maybe (QueryVal st))
 getPathLatest path = do
-    bs <- asks hBlockStore
-    stateHash <- blockStateHash . blockHeader <$> BlockStore.getTip (hoistBlockStoreReader liftIO bs)
-    result <- getPath stateHash path
-    for result $ \v ->
-        pure (stateHash, v)
+    ledger <- getLedger
+    Ledger.getTipWithState ledger >>= \case
+        Left _err -> pure Nothing
+        Right (_, st) -> pure $ query path st
 
 getBlocks
     :: MonadIO m
@@ -262,10 +240,10 @@ lookupTx tx = do
     bs <- getBlockStoreReader
     BlockStore.lookupTx bs tx
 
-lookupReceipt :: (MonadIO m) => Hashed c tx -> NodeT c tx st s i m (Maybe (Receipt c tx RadicleTx.Output))
+lookupReceipt :: (MonadIO m, Crypto.Hashable c tx) => Hashed c tx -> NodeT c tx st s i m (Maybe (Receipt c tx RadicleTx.Output))
 lookupReceipt txHash = do
-    receiptStore <- getReceiptStore
-    ReceiptStore.lookupReceipt receiptStore txHash
+    ledger <- getLedger
+    Ledger.lookupReceipt ledger txHash
 
 lookupBlock
     :: (MonadIO m)
--- a/src/Oscoin/Storage/Ledger.hs
+++ b/src/Oscoin/Storage/Ledger.hs
@@ -1,0 +1,257 @@
+{-# LANGUAGE UndecidableInstances #-}
+
+-- | A 'Ledger' provides access the states and receipts associated with
+-- blocks in a 'BlockStoreReader'.
+--
+-- This module should be imported qualified.
+module Oscoin.Storage.Ledger
+    ( Ledger
+    , hoist
+    , mkLedger
+    , newFromBlockStoreIO
+
+    , blockStoreReader
+    , getBlocksByDepth
+    , getTipWithState
+    , getTip
+    , lookupState
+    , lookupReceipt
+
+    , buildNextBlock
+    ) where
+
+import           Oscoin.Prelude
+
+import           Oscoin.Crypto.Blockchain
+import           Oscoin.Crypto.Blockchain.Eval
+import qualified Oscoin.Crypto.Hash as Crypto
+import           Oscoin.Storage.Block.Abstract (BlockStoreReader)
+import qualified Oscoin.Storage.Block.Abstract as BlockStore
+import           Oscoin.Storage.HashStore
+import qualified Oscoin.Storage.Receipt as ReceiptStore
+import qualified Oscoin.Time.Chrono as Chrono
+
+import           Codec.Serialise (Serialise)
+import           Control.Monad.Trans.Maybe
+import qualified Crypto.Data.Auth.Tree.Class as AuthTree
+
+-- | Errors that may be raised from the ledger's methods
+data LedgerError c
+    = StateNotInCache (BlockHash c)
+    -- ^ State for the given block is not materialized in the state
+    -- cache.
+    | BlockNotFound (BlockHash c)
+    -- ^ A block with the given hash was not found in the block store
+
+deriving instance (Crypto.HasHashing c) => Eq (LedgerError c)
+deriving instance (Show (Crypto.Hash c)) => Show (LedgerError c)
+
+data Ledger crypto seal tx output state (m :: * -> *) = Ledger
+    { ledgerBlockStoreReader :: BlockStoreReader crypto tx seal m
+    , ledgerStateStore       :: HashStore crypto state m
+    , ledgerReceiptStore     :: ReceiptStore.ReceiptStore crypto tx output m
+    , ledgerEvaluator        :: Evaluator state tx output
+    }
+
+hoist
+    :: (forall a. m a -> n a)
+    -> Ledger crypto seal tx output state m
+    -> Ledger crypto seal tx output state n
+hoist f l =
+    Ledger
+        { ledgerBlockStoreReader = BlockStore.hoistBlockStoreReader f $ ledgerBlockStoreReader l
+        , ledgerStateStore = hoistHashStore f $ ledgerStateStore l
+        , ledgerReceiptStore = ReceiptStore.hoistReceiptStore f $ ledgerReceiptStore l
+        , ledgerEvaluator = ledgerEvaluator l
+        }
+
+
+-- | Create a ledger backed by the given block store and in-memory
+-- storage of state and receipts.
+--
+-- @initialState@ is the state that the transactions of the
+-- genesisblock are applied to.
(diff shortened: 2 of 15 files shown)


src/Oscoin/Node.hs
```diff
@@ -71,23 +59,20 @@ import qualified Crypto.Data.Auth.Tree.Class as AuthTree
...
     -> (Handle c tx st s i -> IO a)
     -> IO a
-withNode hConfig hNodeId hMempool hStateStore hBlockStore hProtocol hEval hConsensus =
+withNode hConfig hNodeId hMempool hLedger hProtocol hConsensus =
     bracket open close
   where
     open = do
-        hReceiptStore <- ReceiptStore.newReceiptStoreIO
-        gen <- liftIO (BlockStore.getGenesisBlock hBlockStore)
+        gen <- liftIO (BlockStore.getGenesisBlock $ Ledger.blockStoreReader hLedger)
```
cloudhead: We could also have `Ledger.getGenesisBlock`

src/Oscoin/Storage/Ledger.hs
```diff
@@ -0,0 +1,282 @@
...
+import           Oscoin.Storage.HashStore
+import qualified Oscoin.Storage.Receipt as ReceiptStore
+import qualified Oscoin.Time.Chrono as Chrono
+
+import           Codec.Serialise (Serialise)
+import           Control.Monad.Trans.Maybe
+import qualified Crypto.Data.Auth.Tree.Class as AuthTree
+import qualified Data.List.NonEmpty as NonEmpty
+
+-- | Errors that maybe raised from the ledgers methods
```
cloudhead:
```suggestion
-- | Errors that may be raised from the ledger's methods
```

src/Oscoin/Storage/Ledger.hs
```diff
@@ -0,0 +1,282 @@
...
+    -- cache.
+
+deriving instance (Crypto.HasHashing c) => Eq (LedgerError c)
+deriving instance (Show (Crypto.Hash c)) => Show (LedgerError c)
+
+data Ledger crypto seal tx output state (m :: * -> *) = Ledger
+    { blockStore   :: BlockStore crypto tx seal m
+    , stateStore   :: HashStore crypto state m
+    , receiptStore :: ReceiptStore.ReceiptStore crypto tx output m
+    , ledgerEval   :: Evaluator state tx output
```
cloudhead: Should this just be `eval` or `evaluator`?

src/Oscoin/Storage/Ledger.hs
```diff
@@ -0,0 +1,282 @@
...
+
+
+-- | Create a ledger using in-memory storage for blocks, states, and
+-- receipts.
+--
+-- Uses 'newFromBlockStore' with 'newBlockStoreIO' initialized with the
+-- given blocks.
+--
+-- @initialState@ is the state that the transactions of the
+-- genesisblock are applied to.
```
cloudhead:
```suggestion
-- genesis block are applied to.
```

src/Oscoin/Storage/Ledger.hs
```diff
@@ -0,0 +1,282 @@
...
+    => Evaluator state tx output
+    -> Chrono.NewestFirst NonEmpty (Block crypto tx (Sealed crypto seal))
+    -> state
+    -> IO (Ledger crypto seal tx output state m)
+newMemoryIO ledgerEval blks initialState = do
+    blockStore <- newBlockStoreIO blks
+    newFromBlockStore ledgerEval blockStore initialState
+
+
+-- | Same as 'newMemoryIO' with just @genesisBlk@ in the blockchain.
```
cloudhead: The function parameter names won't be visible in the haddock, so maybe you could say "the given block"

src/Oscoin/Storage/Ledger.hs
```diff
@@ -0,0 +1,282 @@
...
+-- and receipt associated to a block in the blockchain.
+--
+-- This module should be imported qualified.
+module Oscoin.Storage.Ledger
+    ( Ledger
+    , hoist
+    , mkLedger
+    , newMemoryIO
+    , newMemoryFromGenesisIO
+    , newFromBlockStore
```
cloudhead: Since this function also operates in `IO`, shouldn't it be consistently named with the others?

src/Oscoin/Storage/Ledger.hs
```diff
@@ -0,0 +1,282 @@
...
+
+getTip :: Ledger crypto seal tx output state m -> m (Block crypto tx (Sealed crypto seal))
+getTip ledger = BlockStore.getTip (fst $ blockStore ledger)
+
+
+-- | Get the receipt for a given transaction contained in the
+-- blockchain.
+--
+-- Returns 'Nothing' if the transaction is not contained in the
+-- blockchain or the receipts was not properly build.
```
cloudhead:
```suggestion
-- blockchain or the receipt was not properly built.
```
What does it mean that the receipt wasn't built?

src/Oscoin/Storage/Ledger.hs
```diff
@@ -0,0 +1,282 @@
...
+    => Ledger crypto seal tx output state m
+    -> Crypto.Hashed crypto tx
+    -> m (Maybe (ReceiptStore.Receipt crypto tx output))
+getReceipt ledger txHash = runMaybeT $ do
+    -- Ensure that the transaction is in the block chain. Just calling
+    -- 'ReceiptStore.lookupReceipt' is not sufficient since
+    -- 'buildNextBlock' adds receipts without the block being part of
+    -- the blockchain.
+    _ <- MaybeT $ BlockStore.lookupTx (blockStoreReader ledger) txHash
+    MaybeT $ ReceiptStore.lookupReceipt (receiptStore ledger) txHash
```
cloudhead: I'm not sure I grok this function:
1) By "in the blockchain", do you mean the best chain?
2) Why do we care if it's in the blockchain or not? Why not return it if it exists in the store?

src/Oscoin/Storage/Ledger.hs
```diff
@@ -0,0 +1,282 @@
...
+    maybeState <- lift $ lookupHashContent (stateStore ledger) (Crypto.toHashed stateHash)
+    case maybeState of
+        Just st -> pure (block, st)
+        Nothing -> throwError (StateNotInCache $ blockHash block)
+
+getState
+    :: Ledger crypto seal tx output state m
+    -> Crypto.Hashed crypto state
+    -> m (Maybe state)
+getState Ledger{stateStore} stateHash = lookupHashContent stateStore stateHash
```
cloudhead: Just a minor nit: by convention it's nice to use `lookup` as a prefix for functions that can fail with `Nothing` or `Left`, and `get` for those that shouldn't fail or raise an exception if they do. This is how `containers` does it and how I've tried to name things in the codebase. This would apply to `getReceipt` and a few others too.

src/Oscoin/Storage/Ledger.hs
```diff
@@ -0,0 +1,282 @@
...
+    let (newState, receipts) = evalBlock (ledgerEval ledger) prevState blk
+    storeHashContent (stateStore ledger) newState
+    forM_ receipts (ReceiptStore.storeReceipt (receiptStore ledger))
+    pure newState
+
+
+-- | Computes all states and receipts from the blocks in the
+-- blockchain and adds them to the storage.
+--
+-- The genesis block is applied to @initialState@
```
cloudhead:
```suggestion
-- The genesis block is applied to the given initial state
```
(Makes it more useful in haddock)

test/Test/Oscoin/DummyLedger.hs
```diff
@@ -17,14 +25,21 @@ import           Oscoin.Storage.Block.SQLite
 
 import           Codec.Serialise (Serialise)
 import qualified Data.ByteString as BS
+import qualified Data.ByteString.BaseN as BaseN
 
 import           Test.QuickCheck
 
 newtype DummyTx = DummyTx ByteString
```
cloudhead: If it's 8 bytes, you could use a Word64 :man_shrugging:

src/Oscoin/Storage/Ledger.hs
```diff
@@ -0,0 +1,282 @@
...
+    -> IO (Ledger crypto seal tx output state m)
+newMemoryFromGenesisIO ledgerEval genesisBlk initialState =
+    let blks = Chrono.NewestFirst $ genesisBlk NonEmpty.:| []
+    in newMemoryIO ledgerEval blks initialState
+
+
+-- | Create a ledger backed by the given block store and in-memory
+-- storage of state and receipts.
+--
+-- @initialState@ is the state that the transactions of the
```
cloudhead: Same here ^

src/Oscoin/Storage/Ledger.hs
cloudhead: Perhaps to rephrase: in what circumstance would it not be properly built? What does that indicate?

src/Oscoin/Storage/Ledger.hs
```diff
@@ -0,0 +1,282 @@
...
+    -> m (Maybe state)
+getState Ledger{stateStore} stateHash = lookupHashContent stateStore stateHash
+
+
+-----------------------------------------------------------
+-- * Mutations
+-----------------------------------------------------------
+
+-- | Builds an unsealed block from a list of transactions applied to
+-- the state of the most recent block.
```
cloudhead: Worth being more precise here: by most recent, do you mean *best*?

src/Oscoin/Storage/Ledger.hs
```diff
@@ -0,0 +1,282 @@
...
+    _ <- MaybeT $ BlockStore.lookupTx (blockStoreReader ledger) txHash
+    MaybeT $ ReceiptStore.lookupReceipt (receiptStore ledger) txHash
+
+
+-- | Get the most recent block and the associated state.
+getLatestBlockWithState
+    :: (Monad m)
+    => Ledger crypto seal tx output state m
+    -> m (Either (LedgerError crypto) (Block crypto tx (Sealed crypto seal), state))
+getLatestBlockWithState ledger = runExceptT $ do
```
cloudhead: Does the ledger only ever have access to the best chain?

src/Oscoin/Storage/Ledger.hs
```diff
@@ -0,0 +1,282 @@
...
+       , Crypto.Hashable crypto tx
+       )
+    => Evaluator state tx output
+    -> BlockStore crypto tx seal IO
+    -> state
+    -> IO (Ledger crypto seal tx output state m)
+newFromBlockStore ledgerEval blockStore initialState = do
+    receiptStore <- ReceiptStore.newReceiptStoreIO
+    stateStore <- newHashStoreIO
+    storeHashContent stateStore initialState
```
cloudhead: I know this isn't part of this PR, but this function reads funny, as "storing the contents of a hash" :smile:

src/Oscoin/Storage/Ledger.hs
```diff
@@ -0,0 +1,282 @@
...
+    -- Ensure that the transaction is in the block chain. Just calling
+    -- 'ReceiptStore.lookupReceipt' is not sufficient since
+    -- 'buildNextBlock' adds receipts without the block being part of
+    -- the blockchain.
+    _ <- MaybeT $ BlockStore.lookupTx (blockStoreReader ledger) txHash
+    MaybeT $ ReceiptStore.lookupReceipt (receiptStore ledger) txHash
+
+
+-- | Get the most recent block and the associated state.
+getLatestBlockWithState
```
cloudhead: Since we have `getTip` also, this should probably be `getTipWithState`?

src/Oscoin/Storage/Ledger.hs
```diff
@@ -0,0 +1,282 @@
...
+--
+-- The genesis block is applied to @initialState@
+materialize
+    :: (Monad m, Crypto.Hashable crypto tx)
+    => Ledger crypto seal tx output state m
+    -> state
+    -> m ()
+materialize ledger initialState = do
+    genesisBlock <- BlockStore.getGenesisBlock (blockStoreReader ledger)
+    allFollowingBlocks <- BlockStore.getBlocksByParentHash (blockStoreReader ledger) (blockHash genesisBlock)
```
cloudhead: Seems like we need a `getBlocksFromHeight` and then we can do `getBlocksFromHeight 0`

src/Oscoin/Storage/Ledger.hs
```diff
@@ -0,0 +1,282 @@
...
+    _ <- foldlM (materializeBlock ledger) initialState (genesisBlock : Chrono.toOldestFirst (Chrono.reverse allFollowingBlocks))
+    pure ()
+
+
+-- | Adds a block to the end of the blockchain and adds the
+-- corresponding state and receipts.
```
cloudhead: Is this really true? Is it not possible that the block is added, but not as a new tip?

src/Oscoin/Storage/Ledger.hs
```diff
@@ -0,0 +1,282 @@
...
+
+-- | Get the @depth@ most recent blocks.
+getBlocksByDepth
+    :: Ledger crypto seal tx output state m
+    -> Depth
+    -> m (Chrono.NewestFirst [] (Block crypto tx (Sealed crypto seal)))
+getBlocksByDepth ledger depth = BlockStore.getBlocksByDepth (blockStoreReader ledger) depth
+
+getTip :: Ledger crypto seal tx output state m -> m (Block crypto tx (Sealed crypto seal))
+getTip ledger = BlockStore.getTip (fst $ blockStore ledger)
```
cloudhead: This always gets the tip of the best chain, correct? Remind me how that works? Ie. how the ledger is always operating on the best fork?

src/Oscoin/Storage/Ledger.hs
geigerzaehler: `eval` might be too ambiguous for a field name. I’ll use `ledgerEvaluator`.

src/Oscoin/Storage/Ledger.hs
geigerzaehler: > By "in the blockchain", do you mean the best chain?

The block store only stores the best chain so it does not make sense to talk about the “best chain” in that context.

> Why do we care if it's in the blockchain or not? Why not return it if it exists in the store?

The receipt store acts as a cache. As explained in the comment `buildNextBlock` will add receipts to the receipt store but this does not mean that the tx is accepted into the blockchain.

test/Test/Oscoin/DummyLedger.hs
geigerzaehler: Using `Word64` would complicate the `TxRow` instance because we would need to convert the Word64 to and from a BytesString. We would also need to add `Hashable c Word64` to `IsCrypto` and implement the instances.

src/Oscoin/Storage/Ledger.hs
geigerzaehler: As mentioned above the blockstore (and thus the ledger) only stores one blockchain so it does not make sense to talk about “best”.

src/Oscoin/Storage/Ledger.hs
geigerzaehler: Yes, indeed.

src/Oscoin/Storage/Ledger.hs
geigerzaehler: I think we need a `Blockstore.foldBlocks` method to do streaming from the SQLite backend. but I’ll leave that for another PR.

src/Oscoin/Storage/Ledger.hs
cloudhead: Oh no the reason I bring it up is because the other fields are not prefixed with `ledger`, but eval is. I think `ledgerEval` is fine, but thought that you might want `ledgerBlockStore` then etc.. :shrug:

src/Oscoin/Storage/Ledger.hs
geigerzaehler: See #490

src/Oscoin/Storage/Ledger.hs
geigerzaehler: That makes sense. I prefixed all fields with `ledger` now.

src/Oscoin/Storage/Ledger.hs
```diff
@@ -0,0 +1,289 @@
...
+data LedgerError c
+    = StateNotInCache (BlockHash c)
+    -- ^ State for the given block is not materialized in the state
+    -- cache.
+
+deriving instance (Crypto.HasHashing c) => Eq (LedgerError c)
+deriving instance (Show (Crypto.Hash c)) => Show (LedgerError c)
+
+data Ledger crypto seal tx output state (m :: * -> *) = Ledger
+    { ledgerBlockStore   :: BlockStore crypto tx seal m
```
adinapoli-mndc: I have discussed this with @geigerzaehler in a call and, in my humble opinion, this is the wrong design move. The `Ledger` is about the computation/application level, whereas the `Protocol` and the `Blockstore` deals exclusively with the settlement layer.

Therefore, it feels iffy to me that we are now shunting the operations on the main chain (i.e. `addBlock` and whatnot) to a component of the system what shouldn't worry about it.

Rather, I would like to give the `Ledger` only a `BlockStoreReader` handle in case it needs to read the state of the main (adopted) chain, but keep both the `Ledger` and the `BlockStore` as part of the `Protocol`. We don't lose anything by doing so: the `Protocol` is still the only owner of the `BlockStoreWriter`, whereas it feels totally normal to have multiple readers for the chain.

I would be curious to hear what folks think about this.

geigerzaehler: Let’s bring this up during the sync and find a place to discuss this

kim: Sounds like a reasonable concern. It would mean that `stepBlock` is responsible for validation, `materializeBlock`-ing, and storing the block, is that correct?

adinapoli-mndc: If by `stepBlock` you mean `stepProtocol` (sorry, I am starting to become carb-depleted 😁 ) I think this would align with my view that the `Procotol` was this overarching component which would perform chain-selection and make sure our view of the chain would be consistent. Originally I did envision a schema where we would simply "push out" events to consumers (like the `Ledger` for example) in the style of `[AddBlock, RollbackBlock,...]` and that the `Ledger` would have been able to update its state accordingly. However , apparently in our current system we cannot do that directly as a block must be evaluated first before it can be considered valid for inclusion. This means that yes, we do need the `Ledger` in scope in the `Protocol` to perform such step (which is unfortunate, but it is the way it is) but we can potentially still retain full control over the chain and its consistency. One very strong argument would also be that `stepProtocol` runs in a critical section and we need to ensure that:

1. We keep the lock as little as possible;
2. We do not call into slow sub-routines which might delay block applications.

(This is slightly orthogonal to the concern discussed here but I guess it's good to remember while we discuss the design).

cloudhead: > I have discussed this with @geigerzaehler in a call and, in my humble opinion, this is the wrong design move. The `Ledger` is about the computation/application level, whereas the `Protocol` and the `Blockstore` deal exclusively with the settlement layer.
> 
Could this be in any way an argument against naming this "Ledger"? vs. against the chosen abstraction?

adinapoli-mndc: > Could this be in any way an argument against naming this "Ledger"? vs. against the chosen abstraction?

Uhm, maybe (thinking out loud here as always): I guess the sour taste in my mouth comes from the fact that, in an ideal world, the `Ledger` would only deal with materialising the state by being fed with information coming from the settlement/consensus layer (as in: this block has been added, these many blocks have been rolled back etc) and at the moment we cannot achieve such clean split due to this validation/evaluation business, which perhaps calls for another component/different design somehow?

geigerzaehler:
Could you have another look @adinapoli-mndc? The Ledger now does not write to the blockstore anymore

src/Oscoin/Storage/Ledger.hs
```diff
@@ -0,0 +1,257 @@
...
+-- The genesis block is applied to the given initial state
+materialize
+    :: (Monad m, Crypto.Hashable crypto tx)
+    => Ledger crypto seal tx output state m
+    -> state
+    -> m ()
+materialize ledger initialState = do
+    genesisBlock <- BlockStore.getGenesisBlock (blockStoreReader ledger)
+    allFollowingBlocks <- BlockStore.getBlocksByParentHash (blockStoreReader ledger) (blockHash genesisBlock)
+    _ <- foldlM (materializeBlockFromState ledger) initialState (genesisBlock : Chrono.toOldestFirst (Chrono.reverse allFollowingBlocks))
```
adinapoli-mndc: How does that play out with performances and space bounds? We obviously cannot run this function on _all_ the blocks all the time. I guess this is not a concern now, but I suppose what we need to do is to write the ledger state on disk for the adopted, immutable part of the chain, and compute in-memory only the part concerning the immutable part (i.e. the 2016 blocks). Should we track this into a separate ticket?

geigerzaehler: This function is only called when initializing a ledger from a block store. But indeed we should persist the state to not have this run on every boot of the node.

I would be more concerned about `materializeBlock` which is not tail recursive and has running time linear to the number of unevaluated transactions (which might be all transactions).

I think as part of our planning we should elicit what the performance requirements of the ledger are and address them when the need arises.

PR #488, opened 2019-04-12
Sized blockchain generators with reduced length

geigerzaehler:
We change the blockchain generator so that
* the length of the blockchain is arbitrary instead of fixed by the size
* the length of the generated blockchain is limited to 15.

Since generating blockchains is expensive limiting the size speeds up the tests by about 15%.

### DIFF
--- a/test/Oscoin/Test/Crypto/Blockchain/Generators.hs
+++ b/test/Oscoin/Test/Crypto/Blockchain/Generators.hs
@@ -4,6 +4,7 @@ module Oscoin.Test.Crypto.Blockchain.Generators
       -- * Generating generic chains
     , genBlockchain
     , genBlockchainFrom
+    , genBlockchainFrom'
     , genOrphanChainsFrom
 
       -- * Generating Nakamoto chains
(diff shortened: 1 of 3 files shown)


test/Oscoin/Test/Crypto/Blockchain/Generators.hs
```diff
@@ -35,11 +38,14 @@ genBlockchainFrom
        )
     => Block c tx (Sealed c s)
     -> Gen (Blockchain c tx s)
-genBlockchainFrom genBlock = sized $ \n ->
+genBlockchainFrom genBlock = sized $ \n -> do
+    let maxSize = (n * 15) `div` 100
```
kim: How about passing the upper bound explicitly? Having to always specify a `Range` is one of the things that make `hedgehog` nicely less magical.

geigerzaehler: Good point! I added `genBlockchainFrom'` which allows you to specify the range.

geigerzaehler: Can you have another look @kim?

geigerzaehler:
> Mixing and matching hedgehog and QC other than converting between the two `Gen` types wholesale is a bit worrisome, though.

I agree. But converting everything that uses `genBlockchain` to hedgehog is a massive undertaking unfortunately.

kim:
Well, `genBlockchain` could be implemented using hedgehog, and converted for QC properties using `hedgehog-quickcheck`. Or not using hedgehog at all, and converted for hedgehog properties using the same package.

geigerzaehler:
I eliminated the use of `Range` to not mix QC and hedgehog.

PR #491, opened 2019-04-12
Eliminate redundant 'reverse' calls when using 'Chrono'

geigerzaehler:
We change the implementation of `Chrono` so that it becomes easier to use. In particular we eliminate calls to `reverse` by folding it into `toNewestFirst` and `toOldestFirst`.

- [ ] update docs


geigerzaehler:
Since there are some valid concerns about this PR and there is no support for the change I’ll close it.

PR #493, opened 2019-04-15
Oscoin Tx and Core types - Part 2.

cloudhead:
* Tests :memo: for `applyTx` and `TxTransfer`
* Generators
* `Condensed` instances
* `Hedghehog.Extended`

### DIFF
--- a/src/Oscoin/Data/Ledger.hs
+++ b/src/Oscoin/Data/Ledger.hs
@@ -24,8 +24,10 @@ import           Oscoin.Prelude
                  , Word8
                  , div
                  , fromIntegral
+                 , map
                  , maxBound
                  , mod
+                 , sum
                  , ($)
                  , (.)
                  , (<>)
@@ -63,6 +65,8 @@ data StateVal c =
     | MemberVal  (Member c)
     | NatVal     Natural
 
+deriving instance (Eq (PublicKey c)) => Eq (StateVal c)
+
 -- | Like 'Map.adjust', but for 'WorldState'.
 adjust :: (StateVal c -> StateVal c) -> StateKey -> WorldState c -> WorldState c
 adjust f k ws =
@@ -80,6 +84,13 @@ alter f k ws =
     case f (WorldState.lookup k ws) of
         Just v  -> WorldState.insert k v ws
         Nothing -> WorldState.delete k ws
+
+-- | /O(n)/. Returns the total supply of tokens in the ledger.
+balanceTotal :: WorldState c -> Balance
+balanceTotal ws = sum $ map f (WorldState.toList ws)
+  where
+    f (_, AccountVal acc) = accountBalance acc
+    f _                   = 0
 
 -------------------------------------------------------------------------------
 -- Core types
@@ -101,6 +112,8 @@ data Account c = Account
     -- ^ The nonce is equal to the number of transactions
     -- made from this account.
     }
+
+deriving instance (Eq (PublicKey c)) => Eq (Account c)
 
 mkAccount :: Address c -> Account c
 mkAccount addr = Account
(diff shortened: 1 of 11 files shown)


test/Oscoin/Test/Data/OscoinTx.hs
```diff
@@ -0,0 +1,183 @@
...
+    ]
+
+propApplyTx :: forall c. Dict (IsCrypto c) -> Property
+propApplyTx Dict = property $ do
+    (pk, pk') <- publicKeyPair @c
+
+    pk /== pk'
+
+    ws    <- forAll genWorldState
+    tx    <- mkTx <$> forAll (genRegisterProject pk)
```
geigerzaehler: This can be an arbitrary transaction right? So why not use `genTx`?

test/Oscoin/Test/Data/OscoinTx.hs
```diff
@@ -0,0 +1,183 @@
...
+    alice <- forAll (genAddress pk')
+
+    -- Should fail because Alice doesn't have an account
+    applyTx tx alice ws ===  Left (ErrKeyNotFound (addressKey alice))
+
+    -- Give Alice an account
+    ws' <- evalEither $ creditAccount alice 1 ws
+
+    -- Should fail because the transaction nonce is incorrect
+    applyTx (tx { txNonce = 1 }) alice ws' ===  Left (ErrInvalidNonce 1)
```
geigerzaehler: Would it make sense to split these tests up into one property each?

src/Oscoin/Data/Ledger.hs
```diff
@@ -102,6 +113,9 @@ data Account c = Account
     -- made from this account.
     }
 
+instance (Eq (PublicKey c)) => Eq (Account c) where
+    (==) a b = accountAddr a == accountAddr b
```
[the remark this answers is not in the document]
cloudhead: Indeed, I reverted to deriving all instances when possible. The only one I can't derive is `Project` because of the functions in there: 2eb1d515b0a186f9bea7ef657ed44680479c7adb

test/Oscoin/Test/Data/OscoinTx.hs
```diff
@@ -0,0 +1,183 @@
+module Oscoin.Test.Data.OscoinTx where
```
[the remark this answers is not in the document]
cloudhead: 1ad97f0

test/Oscoin/Test/Data/OscoinTx.hs
cloudhead: Not really, or I have to complicate the tests a lot: the reason I picked "register project", is that it cannot fail at the payload level - this way I'm only testing the failure scenarios at the Tx level.

cloudhead: Maybe I should add a comment about this?

test/Oscoin/Test/Data/OscoinTx.hs
cloudhead: I thought about it, but since they all share a common setup I thought it might be more messy ... :thinking:

src/Oscoin/Data/Ledger.hs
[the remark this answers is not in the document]
cloudhead: As the data model gains clarity, I might separate the "Project" that is stored in the state, and the "Project" that is passed to the contract handler, at which point I'll revisit this.

test/Oscoin/Test/Data/OscoinTx.hs
cloudhead: Added comment. f561b6a

PR #494, opened 2019-04-15
Extract protocol equivalence tests

geigerzaehler:
We extract the tests that test the protocol against the `STM` blockstore into a separate module. We also change the test implementation so that there is no need to monkey patch `insertBlock`.

### DIFF
--- a/test/Test/Oscoin/Protocol.hs
+++ b/test/Test/Oscoin/Protocol.hs
@@ -1,0 +1,102 @@
+module Test.Oscoin.Protocol
+    ( tests
+    ) where
+
+import           Oscoin.Prelude
+
+import qualified Oscoin.Consensus.Config as Consensus
+import           Oscoin.Consensus.Nakamoto (blockScore)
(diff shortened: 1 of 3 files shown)


test/Test/Oscoin/Protocol.hs
```diff
@@ -0,0 +1,102 @@
...
+
+import           Test.Oscoin.DummyLedger
+import           Test.QuickCheck.Extended
+import           Test.QuickCheck.Monadic
+import           Test.Tasty
+import           Test.Tasty.QuickCheck hiding ((===))
+
+tests :: Dict (IsCrypto c) -> TestTree
+tests d = testGroup "Test.Oscoin.Protocol"
+    [ testProperty "prop_StmBlockStoreEquivalence"  (prop_StmBlockStoreEquivalence d)
```
adinapoli-mndc: I liked the old name a bit better -- or, at least, I would suggest making this one slightly more descriptive, as it's a bit obscure now.

PR #495, opened 2019-04-15
Add generators for blockchains with correct state hashes

geigerzaehler:
We introduce the `genEvaledBlockchainFrom` generator for blockchains that ensures that the state hash is correctly computed. We base the `genBlockchainFrom` function on the new function and add tests that check the state hash.

 We will need these generators to test the ledger properly.


kim:
Is the test failure in any way related?

geigerzaehler:
> Is the test failure in any way related?

Yup. Seems like an 64 bit Tx is not enough to ensure that generated txs are unique. I was  I bumped it to 80 bits.

In [theory](https://en.wikipedia.org/wiki/Birthday_attack) 64 bits should be enough to ensure hash uniqueness. With 20 billion randomly generated txs the likelyhood of a collision is less than 0.1%. But the issue occured more frequently than that.

geigerzaehler:
As @cloudhead rightfully pointed out, the `arbitrary :: Gen Word8` is biased and this results in hash-collisions. I replace this with `chooseAny`.

test/Oscoin/Test/Crypto/Blockchain/Block/Generators.hs
```diff
@@ -125,6 +144,27 @@ genBlockWith parentBlock txs st = do
...
+       , Serialise tx
+       , Arbitrary s
+       )
+    => Block c tx s
+    -> state
+    -> [tx]
+    -> Evaluator state tx output
+    -> Gen (Block c tx s, state)
+genEvaledBlock parentBlock parentState txs evl = do
+    let (_, newState) = evalTraverse evl txs parentState
```
kim: I am skeptical of ignoring evaluation errors when the `Evaluator` can be passed in, as it's kind of easy to glance over an invariant you're actually assuming in the test. Could we make it so the caller has to explicitly decide to ignore `EvalError`s?

geigerzaehler: In our current implementation of evaluation blocks that contain transactions are valid blocks and no distinction is made between blocks that contain failing transactions and those that don’t. It therefore makes sense to also generate the full set of valid blocks and not restrict ourselves to blocks that only contain non-failing transactions.

As mentioned before, the current approach is not set in stone. I created #500 to track the discussion.

kim: I get that. We are talking about a re-usable generator for tests here, however. For the sake of stating all invariants explicitly, this either needs to disallow passing in an `Evaluator` which can fail, or the property needs to specify that it is explicity not interested in evaluation errors (because "current implementation").

kim: One approach could also be to allow the `Evaluator` to only be chosen from a set of known evaluators, each of which has certain properties.

geigerzaehler: Currently there is no use case to having the generators only include non-failing transactions. Because of YAGNI I will not implement this the more sophisticated generators. If we need them we can implement them. I documented the behavior regarding failed transactions though.

kim: “Currently”

geigerzaehler:
@cloudhead Do you want to weigh in on this?

kim:
Summary of RTC with @geigerzaehler:

The problem, as stated by me, is that passing in _any_ `Evaluator` gives too many degrees of freedom for properties - the choice of concrete `Evaluator` shouldn't matter, unless it is itself under test. Hence, a property should either receive any arbitrary function, or an arbitrary function which satisfies some invariants (eg. never fails, always fails, identity function, etc.).

The counter argument by @geigerzaehler is that the `Evaluator` is _actually_ determined by the type of `tx`, thus the correct fix would be to implement #483, so these generators can obtain the `Evaluator` by applying the type family.

If we agree that this is indeed the case, I am ok with this patch as-is as a temporary measure. 

/cc @adinapoli-mndc @cloudhead

cloudhead:
We just spoke a bit about this with @adinapoli-mndc - I don't think the `Evaluator` type has to be a function of the `tx`. It *could* be, but this doesn't really give us much (afaict), however it does limit us to using different evaluators for the same `tx`.

Not sure this helps though? I'll have a deeper look at the code.

test/Oscoin/Test/Crypto/Blockchain/Block/Generators.hs
cloudhead: I share @kim 's concern here - I guess one option could be to return the errors in the tuple?

cloudhead: To restate the concern: if we want this generator to be general-purpose (which is suggested by its name and location), we should somehow capture the errors somewhere. You could make the same argument for capturing the outputs too in fact :thinking:

adinapoli-mndc:
> The problem, as stated by me, is that passing in any Evaluator gives too many degrees of freedom for properties - the choice of concrete Evaluator shouldn't matter, unless it is itself under test. 

Right, this seems a valid concern -- trying to also factor out the conversation we had with @cloudhead yesterday, here is a straw-man proposal:

* We tackle #483 by _not_ baking the `Evaluator` in the type family (i.e. the current status quo). I think this is sensible, because one can have (in principle) multiple ways to evaluate the state given the same `Tx` type, and the choice of `tx` shouldn't necessarily lock us in place;

* We do address @kim 's concern and we separate tests on the Tx alone (which uses an arbitrary evaulator) vs the tests which test the concrete, production-like evaluator, making sure errors are not swallowed.

Wdyt?

kim:
> I don't think the Evaluator type has to be a function of the tx

*sigh* 

Alright. So, inch-pinching again about the testing strategy: in general, we want all _property_ tests (and consequently the code under test) which use a `Block` or a `Blockchain` to be oblivious of the `Evaluator` used (`forall evaluator.` if you wish). Otherwise, every such test will have to a) pick some concrete `Evaluator`, and b) assert that the evaluation results match the expectations (or at least make the developer *think* about whether or not it matters, where the types don't really guide you). At the end of the day, many permutations of `Evaluator`s will have to be tested to achieve the desired coverage -- which is kind of the purpose of using property testing in the first place, no?

So, I guess the best way forward here is to change the `Evaluator` argument to `Gen Evaluator`, which should allow existing tests to continue passing it in (`pure dummyEval`), but would nudge us into using arbitrary functions in the future.

PR #501, opened 2019-04-16
[WIP] Tentative fix for #460

adinapoli-mndc:
_N.B. this PR is a WIP in the sense that the main idea is there, but we have to discuss how we can further improve it._

This PR tries to unify the impedance mismatch we have seen lately during equivalence tests, where we had to "patch up" the tests in a way that one would run the protocol and the other would not. The impedance mismatch came from the fact that calling `stepProtocol` would lock in place two things:

1. The use of the `Orphanage`, which I have later realised is an implementation detail of the particular "realWorld" `BlockStore` we are using. It's surely geared towards a production use where things like time and space performance matters, but is not necessarily the easiest data structure to work with in testing/average case;

2. The use of a single chain selection (based on the `Orphanage`) which would rule out the possibility of using different chain selection algorithms for equivalence testing.

This led me to introduce the `BlockTree` module, which is conceptually very similar to what @cloudhead had in mind originally, with a twist:

```haskell
data BlockTree c tx s m = BlockTree
    { btFullBlockStore :: BlockStore c tx s m
    -- ^ An opaque 'BlockStore'.
    , insertOrphan     :: Block c tx (Sealed c s) -> m (BlockTree c tx s m)
    -- | Returns 'True' if the input block is novel to the 'BlockTree', i.e it's
    -- neither in the orphan store nor in the block store.
    , isNovelBlock     :: BlockHash c -> m Bool
    }
```

It works a bit as a persistent data structure where calling `insertOrphan` generates a new `BlockTree`, then each implementation can decide how to implement it. We can have a `RealWorld` implementation that uses the SQLite store and the `Orphanage` and performs chain selection every time we call `insertOrphan`:

```haskell
newBlockTree
    :: ( Monad m
       , Ord s
       , Ord (Hash c)
       )
    => Consensus.Config
    -> Validate c tx s
    -> (Block c tx (Sealed c s) -> Score)
    -> BlockStore c tx s m
    -> BlockTree c tx s m
newBlockTree cfg validateFull scoreBlock =
    newBlockTreeInternal (O.emptyOrphanage scoreBlock) cfg validateFull scoreBlock

newBlockTreeInternal
    :: ( Monad m
       , Ord s
       , Ord (Hash c)
       )
    => Orphanage c tx s
    -> Consensus.Config
    -> Validate c tx s
    -> (Block c tx (Sealed c s) -> Score)
    -> BlockStore c tx s m
    -> BlockTree c tx s m
newBlockTreeInternal o cfg validateFull scoreBlock bs =
    let btree = BlockTree
            { btFullBlockStore = bs
            , insertOrphan = \blk -> do
                o' <- chainSelection (O.insertOrphan blk o) cfg validateFull scoreBlock bs
                pure $ newBlockTreeInternal o' cfg validateFull scoreBlock bs
            , isNovelBlock = \h -> do
                  insideBlockStore <- not <$> BlockStore.isNovelBlock (fst bs) h
                  pure (not insideBlockStore && not (O.member o h))
            }
    in btree
```

And a `Reference` implementation that simply piggybacks on the dear old `Pure.hs` store to perform storage + chain selection, which is essentially .. vacuous:

```haskell
newBlockTree
    :: ( Monad m )
    => BlockStore c tx s m
    -> BlockTree c tx s m
newBlockTree bs =
    let btree = BlockTree
            { btFullBlockStore = bs
            -- Inserting an orphan is indistinguishable from inserting a
            -- \"normal\" block, for the pure 'BlockTree'.
            , insertOrphan = \blk -> do
                BlockStore.insertBlock (snd bs) blk
                pure $ btree
            -- The pure 'BlockTree' searches automatically both in the main
            -- chain /and/ in the orphans.
            , isNovelBlock = \h -> not <$> BlockStore.isNovelBlock (fst bs) h
            }
    in btree
```

You will notice a bunch of things:

* Chain selection is tucked away inside the call to `insertOrphan`, so that this is just an implementation detail of each particular `BlockTree`. This is _not quite_ what @cloudhead had in mind as he wanted this function to be called `insertBlock` and to have inside the logic to check whether this block was a tip-extension or an orphan. I eventually refrained from doing that, because doing that would have moved all the complexity of block validation which is now in the `stepProtocol` function inside the `BlockTree`, and would have made the `Ledger` integration harder (my gut feeling).

* The `stepProtocol` function clears up and `runProtocol` now takes a `BlockTree` rather than a `BlockStore`.

This had the nice consequence that planet finally align in the equivalence tests, and now they can be written as follows:

```haskell
testEquivalence
    :: forall c a. (IsCrypto c, Eq a, Show a, Condensed a)
    => EquivalenceTestRunner c a
    -> Property
testEquivalence runScript = monadicIO $ do
    metricsStore <- liftIO $ Telemetry.newMetricsStore Telemetry.noLabels
    let telemetry = Telemetry.newTelemetryStore Telemetry.noLogger metricsStore

    stmValue <- liftIO (runWithStm telemetry)
    sqlValue <- liftIO (runWithSqlite telemetry)
    pure $ sqlValue === stmValue
  where

    initialBlockchain = fromGenesis defaultGenesis
    cfg = Consensus.Config 1024 2016
    noValidation _ _ = Right ()

    runWithStm :: Telemetry.Handle -> IO a
    runWithStm telemetry =
        STM.withBlockStore initialBlockchain blockScore $ \blkStore -> do
            let btree = RealWorld.newBlockTree cfg noValidation blockScore blkStore
            runProtocol noValidation blockScore telemetry btree cfg $ \Handle{dispatchBlockSync} ->
                runScript (fst blkStore) dispatchBlockSync

    runWithSqlite :: Telemetry.Handle -> IO a
    runWithSqlite telemetry = do
        SQLite.withBlockStore ":memory:" defaultGenesis $ \blkStore -> do
            let btree = Reference.newBlockTree blkStore
            runProtocol noValidation blockScore telemetry btree cfg $ \Handle{dispatchBlockSync} ->
                runScript (fst blkStore) dispatchBlockSync
```

What I find satisfactory is that there is now no conceptual difference between the two and, most importantly, we actually test `runProtocol` among the other things (i.e. the abstraction end2end, not just a tiny piece). This makes me hopeful this is now going into the right direction.

**As a consequence of this refactoring, now the Memory.hs blockstore introduced by Thomas some PR ago is not necessary anymore, and I was able to remove it.**

What do you guys think?


cloudhead:
I like where this is going, but I find the interface strange:

1. If the orphanage is hidden inside the block tree, how is an external consumer supposed to know if a block is orphan or not before inserting it?
2. How do you insert a non-orphan block into the block tree?
3. How do you lookup a block?

I'd also negate and rename `isNovelBlock` to `isMember` or something because the concept of novelty doesn't matter here, and it can act as a "container" then.

adinapoli-mndc:
@cloudhead I see what you are saying, and I feel there is a deliberate choice we have to do here: either we do replicate (partially) the interface that the `BlockStore` already exposes to have something like `insertBlock` and `lookupBlock`, or we accept that the `BlockTree` is a data structure that simply wraps a `BlockStore` plus some (hidden) way of storing orphans and a way to do chain selection and we keep using the "low-level" `BlockStore` all the way.

Honestly, this would be my preference: I do understand the need to have a `isNovel` function for gossip purposes, but I would argue that for all other consumers of the chain, they don't need the full `BlockTree`: all they care is to have a read handle to read the main chain, and this is what the `BlockStore` is for.

Let me try restating the same concept in different terms: I think the `BlockTree` should be consumed only by the `Protocol`, but the rest of the app should deal exclusively with the `BlockStore` as this is what encapsulate the single-source-of-truth, i.e., the adopted chain. Passing the `BlockTree` around or having it expose methods to insert & lookup stuff sounds a bit like information leakage to me. 

Wdyt?

> I'd also negate and rename isNovelBlock to isMember or something because the concept of novelty doesn't matter here, and it can act as a "container" then.

Fair point -- Even `member` would work, as that is what other libraries like `containers` expose anyway (modulo the fact their function is pure and our effectful 😁 )

cloudhead:
I guess I don't quite understand what this is meant to abstract/encapsulate then. :thinking:

adinapoli-mndc:
> I guess I don't quite understand what this is meant to abstract/encapsulate then. 🤔

It essentially hides away the orphan store + the chain selection, I guess. Prior to that, we had everything tangled in one single data structure which was doing everything. Now, I suppose, we have a _really_ dumb `BlockStore` that is concerned just with the main adopted chain and this "intermediate" data structure which deals with the concern of storing orphans + performing chain selection. As a consequence, now you had the orphanage leaking inside the `Protocol` and this was preventing you from using it directly in equivalence tests as now you had one store (the `Pure.hs` one) which was doing everything but now the `Protocol` was also trying to store orphans and performing chain selection, and that explained the redundancy we had before.

Is `BlockTree` a good name for it? Maybe yes, maybe not 🤔

geigerzaehler:
Hey @adinapoli-mndc. I have a couple of concerns with this approach.

* We don’t tackle the mismatch between the behavior of the STM block store and the Sqlite block store. Calling `insertBlock` on the STM block store still behaves very differently from calling the same method on the Sqlite block store. And I think this was the goal behind #460.
* The Protocol is now responsible for validation. It does that before inserting a block that extends the tip. But we also need to validate when selecting a chain. This code is now in the BlockTree. Doing validation there would spread the validation logic.
* In fact it seems the protocol is now only responsible for validation and passing things to the block tree. The field `protoBlockScore` is unused and `Consensus.Config` is just used for validation. Considering the point before that validation also needs to happen as part of chain selection it seems that there is no reason for `Procol` to exist now.
* I would like to see how this proposal simplifies the integration of the protocol in the consensus tests.

adinapoli-mndc:
I think these are all valid concerns that needs to be addressed. Let me go point by point:

While working on this PR I have indeed slightly steered away from the original goal of #460, because I have realised that really the `Pure.hs` store has chain selection so engraved within (as it's basically a sort over a score function) that trying to separate things out would have been a fool's errand, so I have tried to ease the impedance mismatch in some other way.

>  We don’t tackle the mismatch between the behavior of the STM block store and the Sqlite block store. Calling insertBlock on the STM block store still behaves very differently from calling the same method on the Sqlite block store. And I think this was the goal behind #460.

This is true and not very satisfactory. However, the `Protocol` is the only actor having access to the write handle of the `BlockStore`, so this alleviate the issue -- the `Protocol` either calls `insertBlock` or `insertOrphan`, which I agree feels.. a bit hacky 🙂 

> The Protocol is now responsible for validation. It does that before inserting a block that extends the tip. But we also need to validate when selecting a chain. This code is now in the BlockTree. Doing validation there would spread the validation logic.

I agree also with this point (and it's one of the main reasons I have marked the PR as WIP) -- the problem stem from the fact we would ideally like chain selection to be atomic (i.e. happening under the hood in one single operation as that triggers a rollback and it leaves the world briefly in an inconsistent state). However, in order to do that, you do need to perform validation, and I have always thought that's what the `Protocol` is all about (conversely if you put too much validation logic inside the `BlockTree` we are back at square zero where this has too many responsibility). I also thought how this would work out if we allow ourselves to do this in 2-phases, i.e. the `BlockTree` would give the `Protocol` its selected best chain, the `Protocol` would validate it and at that point would ask the `BlockTree` to "commit it". The reason why I am hesitant doing that is twofold: on one side @cloudhead didn't like the fact of having a `selectBestChain` functions as part of the visible API (and I can see why) but from another it makes it harder to reason about the program from a concurrency perspective (as you have moving parts you exchange around and the OS scheduler might interrupt you making the state inconsistent). However, I guess we could mitigate partially this if we stick to our rigid rule of running the `stepProtocol` function in a critical section and we never use the `BlockTree` outside the `Protocol`.

> In fact it seems the protocol is now only responsible for validation and passing things to the block tree

Bingo, I think this is what the `Protocol` should do, ideally: validate the blocks and orchestrate the different components which are relative to the settlement layer.

> I would like to see how this proposal simplifies the integration of the protocol in the consensus tests.

I didn't try yet but I suspect this proposal per se won't do much -- I suspect the difficulty of integration might stem from the fact the `Protocol` has to run in `IO` due to the `MVar`, but in our current asset doing so universally is not possible with the test machinery -- we might have to reason about that at a later stage.

adinapoli-mndc:
We discussed flaws and concerns with this design with @cloudhead today, and I am going to write a post-mortem as part of #460 , and (probably) going to write this one to start off with a clean slate.

PR #503, opened 2019-04-17
Fix #473 (Replace radicle)

### LINKED ISSUE #473: Remove the dependency on Radicle
I am opening this issue as it was tucked as an item of #404 but since then I have discussed the issue with @geigerzaehler and he agreed for me to take this over, provided this is really achievable to do in a "one big gulp", as the system seems to be quite tangled with radicle stuff at the moment.

adinapoli-mndc:
This is the spiritual successor of #483 . It's still a WIP but while working on it I have gained a bunch of interesting insights I would like to share with you all:

* After @cloudhead 's ~nudge~ suggestion I have eliminated the need for the `IsTx` class and used separate (closed) type families which makes the change less pervasive;

* I have added a `TxOutput` and a `TxState` type family. The latter allows us to get rid of the `st` type parameter in a bunch of places and I think it cleans up the code a bit;

* The `src` compiles (but genesis needs to be regenerated), but at the moment I am using a dummy `Evaluator` and `TxState` in `Main.hs`, and I am not yet switch to be using the `OscoinTx.Tx` type (but I am keeping the old `Tx` type after Alexis' suggestion);

* There are a bunch of places where previously we would have taken a `Rad.Value` as input -- notable mentions are the [submitTransaction](https://github.com/oscoin/oscoin/compare/adinapoli/issue-473-attempt-2?expand=1#diff-2147fb98f50ec7e02a81eba21f92333fR123) function and the [createValidTx](https://github.com/oscoin/oscoin/compare/adinapoli/issue-473-attempt-2?expand=1#diff-f1fa57c224efa379bd6a6f309f5d736fR177) function, but I think now this doesn't make sense anymore. For now I have banged a `TxOutput` there for the sake of making the code compile but that is not obviously correct. I _think_ what we might need is a `TxPayload` typeclass, after all 😏


src/Oscoin/Data/Tx.hs
```diff
@@ -16,9 +16,20 @@ import           Control.Monad.Fail (fail)
...
+type family TxOutput c tx where
+    TxOutput c (OscoinTx.Tx c) = OscoinTx.TxOutput
+    TxOutput c (Tx c)          = OscoinTx.TxOutput
+
+-- FIXME(adn) These are placeholders for now.
+type family TxState c tx where
+    TxState c (OscoinTx.Tx c) = ()
+    TxState c (Tx c) = ()
+
+type Tx c = Tx' c OscoinTx.TxOutput
```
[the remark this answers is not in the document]
adinapoli-mndc: Forget about that, it was just a placeholder I added as I was lacking the `TxPayload`, which I am now adding -- wait and see, sir 😁

src/Oscoin/Data/Query.hs
```diff
@@ -6,3 +6,10 @@ class Query a where
     type QueryVal a :: *
 
     query :: [Text] -> a -> Maybe (QueryVal a)
+
+--NOTE(adn) We don't seem to have any 'instance Query' within the codebase.
```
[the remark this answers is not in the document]
adinapoli-mndc: Yeah, maybe we can indeed.

adinapoli-mndc:
>  I mean, sure, it is nice to write code generic in those types (f.ex. Oscoin.Storage.Ledger), but we are attempting to keep the entire system generic in the type of transactions. This leads to this parallel reality we create under test/, where not only every effect is mocked, but also datatypes.

Right, this is a valid concern. Or, at least, in principle it's conceivable that _some parts_ of a `tx` could be generic (for example as they are simpler to mock out) but that the bulk of the `tx` should be concrete and not polymorphic. Hopefully this will clear up with the ledger specification and we will be able to reach that sweet spot where you don't have to endure the pain of creating a fully-fledged production `Tx` but at the same time you are _actually_ testing meaningful things, not blocks with `Text` blobs in in, which I agree is much more productive and sound.

adinapoli-mndc:
Ok, this should feel and be a lot cleaner now. It compiles just fine, but the miner is broken. Once I fix that, I think we should be in business (modulo code reviews 😁 ).

src/Oscoin/API/Client.hs
```diff
@@ -6,15 +6,15 @@ import           Oscoin.Prelude
...
 
     -- | Returns an error result if a transaction with the given hash
     -- was not found.
-    getTransaction :: Hashed c (RadicleTx.RadTx c) -> m (Result (TxLookupResponse c))
+    getTransaction :: Hashed c (Tx c) -> m (Result (TxLookupResponse c))
 
     -- | Returns an error result if a value with the given key was not
     -- found.
-    getState :: Proxy c -> [Text] -> m (Result RadicleTx.Value)
+    getState :: Proxy c -> [Text] -> m (Result (TxPayload c (Tx c)))
```
cloudhead: This will be `StateVal` eventually, but for now should probably just be `ByteString`

src/Oscoin/Data/Tx.hs
```diff
@@ -15,10 +15,59 @@ import qualified Codec.Serialise.Encoding as CBOR
...
+    TxState c (OscoinTx.Tx c) = Ledger.WorldState c
+    TxState c (Tx c)          = DummyEnv
+
+-- | A 'TxPayload' represent the \"inner body\" of a transaction, i.e. the
+-- unsigned transaction payload that together with a 'Signature' qualifies
+-- a transaction. Due to a slightly incorrect implementation of how signing
+-- worked in the legacy 'Tx', this alias to 'DummyPayload' for the legacy
+-- 'Tx', but hopefully will go away once we switch to 'OscoinTx'.
+-- N.B. The use of this type family might become moot if we switch to a more
+-- concrete way to represeting transactions without the system, but it serve
```
cloudhead:
```suggestion
-- concrete way to representing transactions without the system, but it serve
```
:mag:

test/Oscoin/Test/HTTP/Helpers.hs
```diff
@@ -171,46 +169,36 @@ liftNode na = Session $ ReaderT $ \h -> liftIO (Node.runNodeT h na)
...
+    txVal <- pick arbitrary
+    createValidTx txVal
 
 createValidTx
     :: forall c m.
        (IsCrypto c, MonadIO m)
-    => Rad.Value
-    -> m (Hashed c (API.RadTx c), API.RadTx c)
-createValidTx radValue = liftIO $ do
+    => TxPayload c (Tx c) -- this is not correct
```
cloudhead: :eyes:

cloudhead:
> This is arguably cleaner.
> 
> However, it is still not clear to me why we must keep _everything_ polymorphic in the type of transactions (and consequently their output and the state). I mean, sure, it is nice to write code generic in those types (f.ex. `Oscoin.Storage.Ledger`), but we are attempting to keep _the entire system_ generic in the type of transactions. This leads to this parallel reality we create under `test/`, where not only every effect is mocked, but also datatypes. It seems to me like the situation is like this because we were (and still are) working without a specification of the ledger - this is approaching finality (heh), and from then on we are to evolve the transaction type in a backwards-compatible manner. In other words: we're not writing a blockchain library, let's not make our lives more complicated than need be.
> 
> I think we can go with this as a temporary measure until we are confident of the ledger design (but still can rip out the radicle dependency now), but we should really aim for a `Node` that just fixes all three types.

Agreed - however I do think there is one benefit of the polymorphism here, and it is that it proves to the reader and type system that the contents of the `tx` will not be used or modified.

I'm fine with fixing it in Node, and keeping it polymorphic in "container" modules such as Mempool, BlockStore and Blockchain, if that simplifies things.

test/Oscoin/Test/HTTP/Helpers.hs
adinapoli-mndc: it used not to be correct but with the advent of `TxPayload` the comment is indeed moot. oops!

src/Oscoin/API/Client.hs
adinapoli-mndc: Agreed, the spirit was to change as little things as possible, as we discussed 😁

PR #505, opened 2019-04-18
Fix #381 (use the multibase package directly, drop in-house BaseN)

### LINKED ISSUE #381: Replace the `Multi.hs` module in favour of the external `multihash` library
We now have a standalone `multihash` library in the `ipfs` repo which is basically exactly what we have in `Multi.hs`, so it would be nice to get rid of our module and switch to the external library.

@kim , if you don't mind, I would be happy to tackle this one, as I think could serve me well for #368, as not having that `Multihash` abstraction is scope would force me to treat it as a blackbox.

adinapoli-mndc:
This commit finally fixed #381 and it's dependent on #503 being merged. This finally removes the need of the `BaseN` module in `oscoin` and removes the overlap we previously had between the two libraries.


test/Crypto/Test/Hash/Multi.hs
```diff
@@ -29,7 +29,7 @@ tests = testGroup "Multihash"
                 example BaseN.Base16 C.SHA1 "multihash"
                     "111488c2f11fb2ce392acb5b2986e640211c4690073e"
             , testCase "Base58" $
-                example BaseN.Base58 C.SHA1 "multihash"
+                example BaseN.Base58btc C.SHA1 "multihash"
```
kim: Not sure we need this test module anymore (as per your FIXME)

adinapoli-mndc: Gonna remove them -- opened https://github.com/oscoin/ipfs/issues/51 to not forget.

adinapoli-mndc:
_Screams with all the air in his chest while looking up into the sky:_ _Weeeeder_ 😂

PR #506, opened 2019-04-18
Expose (prometheus) metrics and health check endpoints

kim:
Neither EKG, prometheus metrics, nor health check endpoints should be exposed on the internet. The EKG endpoint is meant for local diagnostics, and not very useful for server deployments.

Thusly, we separate the host/port options for both EKG and metrics/health, and make them optional. I.e. if and only if a host/port pair is explicitly specified on the cmdline we fork the corresponding server thread.

It is not entirely satisfying that these endpoints are only coupled to the lifecycle of the "main" threads by virtue of `Async` supervision, but since L7 load balancing doesn't make a lot of sense for a blockchain full node there doesn't seem to be any obvious alternative.


kim:
@adinapoli-mndc I have squashed all assertions into one `withOscoinExe` now, because it's kinda pointless otherwise. Kent Beck and The Hordes of eXtrem3 pr0gram0rs are gonna haunt me for that, but who's Kent Beck anyway?

adinapoli-mndc:
> I have squashed all assertions into one withOscoinExe now, because it's kinda pointless otherwise. Kent Beck and The Hordes of eXtrem3 pr0gram0rs are gonna haunt me for that, but who's Kent Beck anyway?

I won't stop you, but thou shalt live with this curse on you until the rest of your days 😬 :trollface: (but in practice these tests are meant to be a tool to catch silly things more than doing proper assertions, so I think that's fine)

PR #507, opened 2019-04-19
Modify the serialisation strategy for AddressPayload

### LINKED ISSUE #504: Modify the serialisation of an AddressPayload
## Problem

If we want to tackle https://github.com/oscoin/oscoin/issues/498 and switch to `Ed25519`, we won't be able to perform `PublicKey` recovery from a transaction unless we either embed the `PublicKey` into the `Tx` itself or we make the `PublicKey` recoverable from an `Address`. In the current design, however, this is not possible as we store inside the `AddressPayload` the (short) hash of a public key. 

_Use CBOR to serialise the `PublicKey` into the `AddressPayload`_. This doesn't create any security hazard (a `PublicKey` is meant to be public, after all) at the cost of a small increase in size for the overall `Address` (to quantify when working on the PR).

The alternative solution would be to bake the `PublicKey` into each `Tx`, but @kim raised some good concerns here [already](https://github.com/oscoin/oscoin/pull/477), advocating against this approach.

cc @cloudhead @geigerzaehler

adinapoli-mndc:
Fix #504. I have also added some rudimentary tests to test migration. Albeit not perfect, they are surely a starting point. Something I have realised while writing them is that we will need to be careful if we switch the implementation for the `PK` (for example using a different curve or crypto scheme), as we would need to `newtype` wrap the legacy `PK` and test the migration.

Something else which might be useful when we are close to mainnet is to dump a "legacy address" into binary format and hardcode it inside the codebase (alas golden testing). This way we can test we can always evolve from the first and foremost `Address` we generated and hopefully accidentally breaking backward-compatibility should be caught by our tests.


test/Test/Oscoin/Crypto/Address.hs
```diff
@@ -119,9 +134,106 @@ addressesRef :: IORef (Word64, Set.Set Text)
 addressesRef = unsafePerformIO $ newIORef (0, mempty)
 {-# NOINLINE addressesRef #-}
 
+{------------------------------------------------------------------------------
+  Testing migrations
+------------------------------------------------------------------------------}
+
+-- Testing that we can migrate from different representation of an 'Address'
+-- in important to future-proof our system. To do so, we 'newtype' wrap our
```
cloudhead:
```suggestion
-- is important to future-proof our system. To do so, we 'newtype' wrap our
```

adinapoli-mndc:
@kim @cloudhead before the chaos of the day starts, I wanted to collect my thoughts on something I have been thinking non-stop since I have issued this PR (but I have refrained to comment over the holidays, as any good internet citizen 😁 ).

I think something is not quite right with our `format_version` byte, i.e. the one we decided to reserve as a point of evolution whether in the future we would like to switch away from CBOR. The truth is, this is not very useful in our current shape, and I have realised this while working on the migration tests, so let me try explain.

In the current shape, this byte indicates only the serialisation format of the _current_ address we are trying to encode or decode, and it's only _minimally_ useful for 3rd parties wanting to decode an oscoin address. All they can do is to parse this byte and acknowledge that the address they are working with is either CBOR (or JSON, say). But that's it. If they were going to parse the `AddressPayload` they would know that what follows is a CBOR blob, but then they will have to dig into the source code to understand which CBOR-tag maps to which `AddressPayload` version.

Conversely, if you are a library author (i.e. one of the oscoin contributors) the situation is even worse, as you have to write to replicate the serialisation-tag in both instances you write. Let's take as an example an hypothetical world where we have both CBOR and JSON, and let's say we migrated to JSON from the "version 3" onwards. Then, you will have something like this:

```haskell
deserializeFutureAddress
    :: Serialise (PublicKey c)
    => ByteString
    -> Either DeserializeError (FutureAddress c)
deserializeFutureAddress blob = flip evalState blob . runExceptT $ do
    crc <- Checksum <$> withExceptT DecodeError (consumeFromEnd 6)
    protectedData <- get

    unless (isValidChecksum protectedData crc) $
        throwError (InvalidChecksum crc)

    legacyPrefix <- deserializeAddressPrefix
    fooByte <- withExceptT DecodeError parseWord8
    barByte <- withExceptT DecodeError parseWord8
    let prefix = FutureAddressPrefix legacyPrefix fooByte barByte

    FutureAddress <$> pure prefix <*> deserializeAddressPayload (addressFormat legacyPrefix)
  where
    deserializeAddressPayload (AddressFormat tag) = do
        binaryBlob <- get
        case tag of
          AddressFormatTag_CBOR ->
              liftEither $
                  first DeserializePayloadError
                        (deserialiseOrFail (toS binaryBlob))
```

adinapoli-mndc:
ops, accidentally closed the PR. Also started to partially-write some morning ramble which I think I have now convinced myself it was just a ramble. If you see it pops in your email feed, ignore it 😁

cloudhead:
> Something else which might be useful when we are close to mainnet is to dump a "legacy address" into binary format and hardcode it inside the codebase (alas golden testing). This way we can test we can always evolve from the first and foremost Address we generated and hopefully accidentally breaking backward-compatibility should be caught by our tests.

:+1: we should do this for other things too, this is how bitcoin tests everything - they have test vectors which compare hard-coded hashes to things.

src/Data/Binary/Parsing.hs
```diff
@@ -0,0 +1,60 @@
+-- | Simple parsing utilities.
```
cloudhead: I feel like I commented on something like this some time ago :thinking: but I think this should be `Binary.Decoding`, as "parsing" is used more in the context of language.

src/Oscoin/Crypto/Address/Serialisation.hs
```diff
@@ -50,42 +51,6 @@ data DeserializeError =
       -- ^ The input bytestring wasn't a valid base32z-encoded one
     deriving (Eq, Show)
 
--- | A 'Decoder' is simply a 'StateT' transformer which carries around the
--- binary blob being decoded (and consumed) and using 'Except' at the bottom
--- of the stack, in order to be able to report precise errors in case the
--- decoding process fails.
-type Decoder a = StateT ByteString (Except DeserializeError) a
```
cloudhead: Ah here it is xD, we did call this decoder previously!

src/Oscoin/Crypto/Address/Serialisation.hs
```diff
@@ -154,43 +122,39 @@ deserializeAddressType = do
...
-instance ( ByteArrayAccess (ShortHash c)
-         , CBOR.Serialise (ShortHash c)
-         ) => CBOR.Serialise (AddressPayload c) where
-    encode (AddressPayload (Tagged p)) =
+instance CBOR.Serialise (PublicKey c) => CBOR.Serialise (AddressPayload c) where
+    encode (AddressPayload_V0 pk) =
            CBOR.encodeListLen 2
         <> CBOR.encode (0 :: Word8)
-        <> CBOR.encodeByteArray (toSliced p)
+        <> CBOR.encode pk
```
cloudhead: I think this is what you were talking about - but we should really ensure that whatever this does is consistent with some hard-coded value, eg. that the tags used by cbor are the same accross all versions, or hand-roll the instance of PK (maybe we do this already)?

kim:
Would it be reasonable to add a length check, at least in the tests? Seems easy to change the implementation of `PublicKey` and suddenly the address length explodes.

src/Oscoin/Crypto/Address/Serialisation.hs
adinapoli-mndc: Right, I have now separated the general-purpose monad to be called a `Parser` as it simply interprets binary blobs, whereas a `Decoder` is something which _interprets_ those bytes into a domain-specific type 😁

adinapoli-mndc:
> Would it be reasonable to add a length check, at least in the tests? Seems easy to change the implementation of PublicKey and suddenly the address length explodes.

Yes, this is something which occurred to me this morning as well -- we should have a length check so that if we change anything in our format the relevant test fails and it forces us to _reason_ about the failure, for the good or the bad 😉

src/Oscoin/Crypto/Address/Serialisation.hs
adinapoli-mndc: I think that trying to be consistent with CBOR tags across all versions is a nightmare and simply not feasible 😅 😁 . Yes, we can (and should) hard-roll the instance of PK, but I think I am hinting  at more general use of a pattern here: whenever we have types that might evolve, we should probably represent them as sum types or GADTs (a bit like I did here for the `AddressPayload` or you guys did for the `Tx`) so that we can match the particular constructor with the particular serialised representation. This is probably obvious for the most parts but I guess we should start thinking which are those types and represent them consistently (and perhaps dump some documentation somewhere on how to write new types in the future).

adinapoli-mndc:
> Would it be reasonable to add a length check, at least in the tests? Seems easy to change the implementation of PublicKey and suddenly the address length explodes.

Done as part of 0254591a (will rebase and merge at the end).

src/Oscoin/Crypto/Address/Serialisation.hs
cloudhead: Hmm I'm not sure how the sum types help here - we still have to have a spec somewhere of what the different tags mean.

adinapoli-mndc: By all means, one doesn't exclude the other -- I was merely pointing out that you do need a sum type (or GADT) if you don't want to have each version bump to cause a cascade effect everywhere else (unless that's the intended side-effect, of course), while being able to map each tag to a different constructor.

Ideally we could also think about using some SOP/TH dark magic to generate such documentation directly from the code.

src/Oscoin/Crypto/Address/Serialisation.hs
cloudhead: :smile: not sure this makes sense - parsing has to do with language/text/syntax, so if it's related to binary, then it's not parsing :thinking: - In both cases you are transforming one representation into another.

adinapoli-mndc: _Handwaving failed, target lost_ -- alternative names are accepted 😁 Should we simply call it `Get` in a similar fashion of what other libraries do (i.e. `cereal` & co)? Of course we do not enforce the same strict semantic (`Get` vs `Put` etc) but it might help conveying the idea behind those utilities.

adinapoli-mndc: oh, and btw, the `cereal` library uses the term "parse" in a bunch of places/functions, so I suppose my original naming was not _that_ crazy 😬 http://hackage.haskell.org/package/cereal-0.5.8.0/docs/Data-Serialize-Get.html#t:Get

cloudhead: `Get` sounds good!

> oh, and btw, the cereal library uses the term "parse/parser" in a bunch of places/functions, so I suppose my original naming was not that crazy

Maybe I'm just pedantic with English :thinking: but I'm pretty sure it's a stretch of the word :smile: :

src/Oscoin/Crypto/Address/Serialisation.hs
cloudhead: Ah ok, I'm more concerned about 3rd party clients being able to decode the format reliably, but yes, the sum-type is also important.

src/Oscoin/Crypto/Address/Serialisation.hs
adinapoli-mndc: 9e45c1b0

src/Data/Binary/Parsing.hs
adinapoli-mndc: 90b0c96

src/Data/Binary/Decoding.hs
```diff
@@ -0,0 +1,59 @@
...
+consume :: Int -> Get ByteString
+consume need = do
+    x <- BS.unfoldrN need BS.uncons <$> lift get
+    case x of
+      (bs, Just rest) | BS.length bs == need -> do
+          lift $ put rest
+          pure bs
+      (bs, _) -> throwError $ NotEnoughInput bs
+
+-- | Parse a 'Word8', or fails otherwise.
```
cloudhead: :point_up:

cloudhead:
```suggestion
-- | Get a 'Word8', or fails otherwise.
```

PR #509, opened 2019-04-19
[WIP] Improved BlockTree, take 2

adinapoli-mndc:
Alright, it's getting late over here so I will have to get back to this PR description to add more bells & whistles , but essentially this PR specs out a `BlockTree` by factoring out all we have been saying in the last few days. I think this is shaping up nicely and all the pieces are starting to fit together.

In the `rm -rf` spirit suggested by @cloudhead and @kim , I have added for now only a brand new `BlockTree` module with two implementations, the classic pure one and the `SQLite` one. I think there are strong indications this is the right design now; in particular, I like how the one and only private method of this API is `insertBlock`, and nothing else.

Furthermore, now `insertBlock` returns a `[NotableEvent]`, which will either be something like:

```haskell
[BlockchainTipExtended (BlockHash deadBeef)]
```

Or, in the case of a rollback:

``` haskell
[RollbackOccurred (Depth 6), BlockchainTipExtended (BlockHash leb0eef), BlockChainTipExtended BlockHash (00abcd)]
```

Which ideally should help with the ledger integration (to go back to @kim 's concern: we could potentially now forward these events in a queue of sort and dispatch them around to other components that needs to track the state. I have for now piggybacked on our `Telemetry` infrastructure, but we can change that.


src/Oscoin/Storage/Block/SQLite.hs
```diff
@@ -65,36 +60,3 @@ withBlockStore path genesisBlock action =
     in bracket (newBlockStore <$> (initialize genesisBlock =<< open path))
                (close . snd)
                (action . fst)
-
-
-lookupBlock
```
adinapoli-mndc: These functions have been moved into the `SQLite.Internal` module, so that they can be reused.

src/Oscoin/Storage/Block/Pure.hs
```diff
@@ -146,12 +148,30 @@ getGenesisBlock Handle{hChains} =
...
+    in case length chainDifference of
+         0 -> -- Nothing happened (orphan added)
+              (handle', mempty)
+         1 -> -- Either a switch-to-fork occurred (of length one) or the chain
+              -- was extended at tip (we can't tell)
+              (handle', OldestFirst [BlockchainTipExtended (blockHash blk)])
+         depth ->
+              (handle', OldestFirst $
+                    RollbackOccurred (fromIntegral depth)
+                  : map BlockchainTipExtended chainDifference
```
adinapoli-mndc: Minimal extension to support this event-driven API (see below).

src/Oscoin/Storage/Block/STM.hs
```diff
@@ -39,7 +40,9 @@ withBlockStore chain score action = do
                 , Abstract.getTip          = withHandle Pure.getTip hdl
                 }
             , Abstract.BlockStoreWriter
-                { Abstract.insertBlock     = modifyHandle . Pure.insert
+                { Abstract.insertBlock     = \blk -> do
+                    _evts <- modifyHandle (swap . Pure.insert blk)
```
adinapoli-mndc: The "legacy" `BlockStore` simply ignores the events.

src/Oscoin/Storage/BlockTree.hs
```diff
@@ -0,0 +1,74 @@
...
+-- 3. Performing chain selection.
+-- 4. Perform full-chain validation.
+
+module Oscoin.Storage.BlockTree
+    ( BlockTree
+    , BlockTreeReader
+    , BlockTreeWriter
+
+    , hoistBlockTreeReader
+    , hoistBlockTreeWriter
```
adinapoli-mndc: I have backported this hoist for simplicity but I hope we can avoid having this one in the first place, as ideally the less degree of freedom we expose when it comes to the writer, the better. We'll see how tests adjustment plays out.

src/Oscoin/Storage/BlockTree.hs
```diff
@@ -0,0 +1,74 @@
...
+-- Useful for testing but discouraged for any serious production use.
+-- NOTE(adn): It might be conceivable to have this function be upgraded as
+-- a proper abstract function of the 'BlockTree', so that different implementations
+-- can decide how to implement it efficiently.
+insertBlocksNaive
+    :: Monad m
+    => BlockTreeWriter c tx s m
+    -> OldestFirst [] (Block c tx (Sealed c s))
+    -> m ()
+insertBlocksNaive bs = traverse_ (insertBlock bs)
```
adinapoli-mndc: Ideally this one should be extended to return all the collected event, in temporal order.

src/Oscoin/Storage/BlockTree/Internal.hs
```diff
@@ -0,0 +1,41 @@
+-- | An opaque 'BlockTree', composed by its public and private API. Its main
+-- responsibilities are:
+--
+-- 1. Storing main blocks.
+-- 2. Storing orphan blocks.
+-- 3. Performing chain selection.
+-- 4. Perform full-chain validation.
```
adinapoli-mndc: Slight documentation duplication, will remove it.

src/Oscoin/Storage/BlockTree/Pure.hs
```diff
@@ -0,0 +1,53 @@
...
+                , Abstract.getBlocksByDepth = \depth ->
+                      withHandle (Chrono.NewestFirst <$> Pure.getBlocks depth) hdl
+                , Abstract.getBlocksByParentHash = \h ->
+                      withHandle (Chrono.NewestFirst <$> Pure.getChainSuffix h) hdl
+                , Abstract.getTip          = withHandle Pure.getTip hdl
+                }
+            , Abstract.BlockTreeWriter
+                { Abstract.insertBlock     = \blk -> do
+                    evts <- modifyHandle (swap . Pure.insert blk)
+                    pure evts
```
adinapoli-mndc: I can avoid a `pure` here.

src/Oscoin/Storage/BlockTree/SQLite.hs
```diff
@@ -0,0 +1,235 @@
...
+           -- the genesis as an orphan would insert it into the orphanage with
+           -- its parent hash, which means that we won't be able to retrieve
+           -- a potential candidate branching off from genesis in our 'selectBestChain'
+           -- function (cfr. QuickCheck seed 554786). To avoid that, we avoid
+           -- inserting the genesis block into the orphanage in the first place;
+           -- After all, the genesis block should never be considered an orphan.
+           else if isGenesisBlock incomingBlock
+               then pure (orphanage, mempty)
+               else do
+                   (o', events) <- selectBestChain mutableChainDepth
```
adinapoli-mndc: Can avoid a `pure` here.

src/Oscoin/Storage/BlockTree/SQLite.hs
```diff
@@ -0,0 +1,235 @@
...
+      release (_, hdl) = SQLite.close hdl
+
+      use :: (MVar (Orphanage c tx s), Handle c tx s)
+          -> IO b
+      use (orphansVar, internalHandle) =
+          let newBlockTree =
+                ( Abstract.BlockTreeReader {
+                      Abstract.getGenesisBlock       = SQLite.getGenesisBlock internalHandle
+                    , Abstract.lookupBlock           =
+                        -- NOTE(adn): This won't lookup orphans (currently). Should we
```
cloudhead: By orphan you mean blocks without a parent, or blocks that aren't on the main chain?

adinapoli-mndc: I was specifically thinking about blocks not on the main chain. In other terms, I quite like the fact the previous `BlockStoreReader` was a sort of "oracle" for the main chain, but if we now lift this property it means our queries are ambiguous, i.e. we wouldn't know if the requested block is on the chain or just an orphan. This in principle can be solved by returning a richer type (i.e `LookupBlockResult` or similar) but I do wonder if we ever need to care about orphans for such queries, and what would happen otherwise. Some food for thoughts:

* The `Ledger` calls `lookupBlock` and this means we would be able to materialise a block even if the parent is an orphan -- is this acceptable behaviour or do we want the `Ledger` to care only about main blocks? Wouldn't be wasteful to do otherwise?

* `lookupBlock` and `lookupTx` are used by the gossip layer to reply to queries (cfr. `wrapLookup`) and while in principle it makes sense to lookup both for orphans & main blocks, it's also true that probably the gossip layer will use this function to exchange main chain blocks during sync, so do we need to also be able to return back orphans (cc @kim ) ?

adinapoli-mndc:
What's left to be done (can be delivered in subsequent PRs):

- [ ] Validate the new API design with the team;
- [ ] Address open questions;
- [ ] Integrate the `BlockTree` into the codebase;
- [ ] Amend the tests to remove the impedance mismatch we had before (read #460 for the initial motivation);
- [ ] Remove the `BlockStore`;

src/Oscoin/Storage/BlockTree/SQLite.hs
cloudhead: I think if possible we should only give access to blocks on the main chain with `lookupBlock`. If later we want to give access to other blocks, we can add a `lookupAnyBlock` or something.

On another note, let's try to be more precise with terminology: an orphan is a block without a parent, while a *stale* block is a block not on the main chain.

kim: Gossip deals with _seen messages_, so it should be _all_ blocks seen so far. As mentioned before, this should be handled by a bloom filter maintained by the gossip layer, not read-through to storage.

adinapoli-mndc: Right, ok, so it looks like the morally correct thing to do is to modify `lookupBlock` and `lookupTx` to deal exclusively with blocks on the main chain, and open an issue to extend the gossip layer to do the bookeeping on its side. Sounds like a plan?

> On another note, let's try to be more precise with terminology: an orphan is a block without a parent, while a stale block is a block not on the main chain.

Roger -- it also implies a subtle property: a stale block is a block which has "lost" against the main chain and that has been discarded, otherwise it would have been named simply "orphan". To say it differently, an orphan block is a block without a parent which at some point _might_ be considered for inclusion into the main chain. If that fails, it transitions from being an orphan to being stale.

kim: What's the significance of distinguishing stale from orphan blocks? `lookupBlock` will see stale blocks until the rollback transaction in the event of switching to a new chain commits.

src/Oscoin/Telemetry.hs
```diff
@@ -82,6 +82,12 @@ emit Handle{..} evt = withLogger $ GHC.withFrozenCallStack $ do
         BlockMinedEvent blockHash ->
             Log.withNamespace "node" $
                 Log.infoM "mined block" fmtBlockHash blockHash
+        BlockchainTipExtended blockHash ->
```
kim: Not necessarily now, but I think we should start nesting related event types, and have fewer `NotableEvent` constructors

src/Oscoin/Storage/BlockTree/Internal.hs
```diff
@@ -0,0 +1,41 @@
...
+    , lookupBlock     :: BlockHash c -> m (Maybe (Block c tx (Sealed c s)))
+    -- ^ Lookups a 'Block' in the tree when given its hash.
+    , lookupTx        :: Hashed c tx -> m (Maybe (TxLookup c tx))
+    -- ^ Lookups a transaction by its hash.
+    , getBlocksByDepth :: Depth -> m (NewestFirst [] (Block c tx (Sealed c s)))
+    -- ^ Returns the last N blocks, given the 'Depth'.
+    , getBlocksByParentHash :: BlockHash c -> m (NewestFirst [] (Block c tx (Sealed c s)))
+    -- ^ Returns the blocks starting from the input parent hash.
+    , getTip          :: m (Block c tx (Sealed c s))
+    -- ^ Returns the tip of the chain.
```
geigerzaehler:
```suggestion
    -- ^ Returns the tip of the best chain in the tree.
```

src/Oscoin/Storage/BlockTree/Internal.hs
```diff
@@ -0,0 +1,41 @@
...
+    { getGenesisBlock :: m (Block c tx (Sealed c s))
+    -- ^ Get the genesis block.
+    , lookupBlock     :: BlockHash c -> m (Maybe (Block c tx (Sealed c s)))
+    -- ^ Lookups a 'Block' in the tree when given its hash.
+    , lookupTx        :: Hashed c tx -> m (Maybe (TxLookup c tx))
+    -- ^ Lookups a transaction by its hash.
+    , getBlocksByDepth :: Depth -> m (NewestFirst [] (Block c tx (Sealed c s)))
+    -- ^ Returns the last N blocks, given the 'Depth'.
+    , getBlocksByParentHash :: BlockHash c -> m (NewestFirst [] (Block c tx (Sealed c s)))
+    -- ^ Returns the blocks starting from the input parent hash.
```
geigerzaehler: This is a bot hard to understand. I think this function is supposed to return all descendants of a given block. How does this play with the tree? Do we return an empty list if the block is not in the best chain? Do we always return the best chain of descendants?

src/Oscoin/Storage/BlockTree/Pure.hs
```diff
@@ -0,0 +1,52 @@
+module Oscoin.Storage.BlockTree.Pure
```
geigerzaehler: Pure is a slightly confusing name since this uses `TVars`. Maybe better to call this `Memory`.

src/Oscoin/Storage/BlockTree/SQLite.hs
geigerzaehler: > a stale block is a block which has "lost" against the main chain and that has been discarded, otherwise it would have been named simply "orphan".

I don’t think it is correct to discard stale blocks: Assume we see two blocks _A_ and _B_ with a common ancestor _O_. Now _B_ scores higher than _A_ so we discard _A_. But the miner of the next block never saw _B_ so the next block _C_ is a parent of _A_ and not of _B_. Now the best chain is _O<-A<-C_. Thus _A_ needs to be re-integrated into the best chain and may not be discarded.

kim:
> But the miner of the next block never saw B so the next block C is a parent of A and not of B.

@geigerzaehler Can you explain this sentence?

src/Oscoin/Storage/BlockTree/SQLite.hs
adinapoli-mndc: > I don’t think it is correct to discard stale blocks: Assume we see two blocks A and B with a common ancestor O. Now B scores higher than A so we discard A. But the miner of the next block never saw B so the next block C is a parent of A and not of B. Now the best chain is O<-A<-C. Thus A needs to be re-integrated into the best chain and may not be discarded.

I guess this is part of a bigger discussion (which we might need to spin up separately) but I think the scenario you are describing shouldn't happen in practice -- at least according to "Mastering Bitcoin" (Chapter 10: Mining and consensus, page 244) it reads:

>Any miners working on extending the chain star-upside-down-triangle (i.e. O<-A) will now stop that work because their candidate block is an "orphan", as its parent "upside-down-triangle" (i.e. A) is no longer on the longest chain.

However I agree that this stuff is trick and easy to get wrong: another use case for a spec to state down clearly our consensus rules and how we plan to address forks and under which rules orphan/stale blocks are governed.

src/Oscoin/Telemetry.hs
adinapoli-mndc: Good idea -- perhaps we could factor them out into "semantic buckets" i.e. `gossip` , `storage`, etc?

src/Oscoin/Storage/BlockTree/Internal.hs
adinapoli-mndc: Right -- we discussed a bit these names with @cloudhead in some previous incantations of this work, and we converged with these 2 names. I suppose the ambiguity would be solved if once again we treat our `BlockTreeReader` as a "main chain oracle" so that you don't have branch in those cases but ideally just a liner sequence of blocks (also known as a `Blockchain` 😁 )

src/Oscoin/Storage/BlockTree/Pure.hs
adinapoli-mndc: Or even possibly simply `BlockTree.STM`. I think the notion of purity here is a bit of a lie (as you pointed out) but I am hesitant in calling this `Memory` as in principle you might have _other_ implementations (for example based on `StateT`) which are also in memory.

Probably `STM` hits the sweet spot as it reveals the inner implementation, a bit in the spirit of `SQLite`.

geigerzaehler:
> > But the miner of the next block never saw B so the next block C is a parent of A and not of B.
> 
> @geigerzaehler Can you explain this sentence?

The node that successfully mines the next block might never see the better block _B_. So it mines a block a block with _A_ as the parent. This makes _O<-A<-C_ a better chain than _O<-B_. I updated the comment to make this clearer hopefully.

src/Oscoin/Storage/BlockTree/SQLite.hs
geigerzaehler: > Any miners working on extending the chain star-upside-down-triangle (i.e. O<-A) will now stop that work because their candidate block is an "orphan", as its parent "upside-down-triangle" (i.e. A) is no longer on the longest chain.

This works if the miner receives _B_ before it mines a block. But we have to consider the case that it does not receive a block until after it has mined _C_.

geigerzaehler: The cited example makes it quite clear. “upside-down triangle” corresponds to _B_ and “triangle” corresponds to _A_. The book tells us that “nodes will set the chain star-triangle-rhombus as the main chain and change the star-upside-down-triangle chain to a secondary chain.” This means that triangle cannot be discarded.

cloudhead: > What's the significance of distinguishing stale from orphan blocks? `lookupBlock` will see stale blocks until the rollback transaction in the event of switching to a new chain commits.

From an external API perspective, none, but internally it makes sense as they have different properties and might be stored and retrieved differently, eg. an orphan block cannot have its state materialized and thus cannot be fully validated, whereas a stale block can.

Issue #502, opened 2019-04-17
Nightly CI


adinapoli-mndc:
## Problem

There are some category of tests which take a long time to be executed but nevertheless they are extremely important in order to build reliable Blockchain software. In our particular case, this could be:

* Running the testsuite against the "realWorld" `Crypto` by using a sufficiently high number of tests repetitions (i.e. `--quickcheck-tests=10000 / --hedgehog-tests=10000` or more);

* Testing node sync in a black box fashion, as in trying to boot up a mini cluster and check that nodes can all sync up to 100% (for example using the Oscoin API to check that the sync progress reaches 100%);

* <put_your_favourite_long_running_task_here>

## Solution/implementation plan

_Setup a Nightly CI_. We could have a bunch of tests which wouldn't be triggered by `stack test` and those could be triggered via a CI task running overnight (scheduled).

In order to test the system using the real world `Crypto` we also have to use singleton types to reify a value-level flag into a type-level term, so that we can run our test executable with one or the other crypto (this might require us to create a custom `tasty` `Ingredient` to act on the command line input.

## (Optional) Alternative solutions considered

The alternative solution would be not to have a nightly CI but to now and again run those long running tests manually on a user machine, making our CPUs work unnecessarily hard.

@kim @geigerzaehler @cloudhead 👍 ? 👎 ?

kim:
:+1:

Implementation-wise, I'd suggest to keep it simple by making our own `main` be responsible for option parsing (instead of tasty).

PR #511, opened 2019-04-23
Allow the crypto under test to be passed via the test CLI

### LINKED ISSUE #510: Allow the testsuite to be passed the crypto implementation
Subtask of #502. In order to run our testsuite at night via the nightly CI, we need a mechanism to convert a value (i.e. a command-line option) into the type so that we can then replace our static `CryptoUnderTest` with a runtime value. That's typically a job for singleton types, for which we can use their simplest form (i.e. no need to pull the `singleton` library into the stack).

adinapoli-mndc:
Fixed #510, which is a precondition for #502.

It turns out I didn't need singletons at all but only a rank-2 function (`withCrypto`). I ended up adding a new ingredient (as I initially anticipated despite @kim advising me not to do it 😢 ) as I did write an initial version using only `getArgs` but I have scrapped it for 2 reasons:

1. It would mess up `tasty`, as now calling `stack test --ta "--crypto mock"` would make `tasty` choke (with an "invalid option" error). The problem is that we can intercept the arguments passed via `getArgs`, but tasty uses internally `optparse-applicative` which would not silently ignore the extra argument, making the overall parser fail;

2. My homemade crappy parser would not handle correctly things like `--crypto=mock` (vs `--crypto mock`) to the point where the only sane option would have been writing a proper `optparse-applicative` `Parser`, which is the same that `Tasty` does under the hood anyway, so the gain wouldn't have been so drastic.

Overall, I don't think that's too heavy of a change.

## Usage

Running `stack test --ta "--crypto mock"` will run the tests using the mock crypto (ditto if you omit the option). Running `stack test --ta "--crypto realworld"` will run them with the full crypto.


test/Main.hs
```diff
@@ -25,28 +27,58 @@ import qualified Test.Oscoin.Protocol
...
+readCryptoUnderTest :: String -> Maybe CryptoUnderTest
+readCryptoUnderTest "realworld" = Just RealWorldCryptoTest
+readCryptoUnderTest "mock"      = Just MockCryptoTest
+readCryptoUnderTest _           = Nothing
+
+instance Tasty.IsOption CryptoUnderTest where
+  defaultValue = MockCryptoTest
+  parseValue = readCryptoUnderTest
+  optionName = pure "crypto"
+  optionHelp = pure "The crypto implementation to use (either 'realworld' or 'crypto')"
```
geigerzaehler:
```suggestion
  optionHelp = pure "The crypto implementation to use (either 'realworld' or 'mock')"
```

adinapoli-mndc: 2082f3c

adinapoli-mndc:
> Nice! Could you also document this in the “Testing” section of the Readme?

Done! 😉

Issue #497, opened 2019-04-15
Add the block height to each block header


adinapoli-mndc:
## Problem

Sooner rather than later, when our devnet will be launched, nodes will need to start communicating between each other not just to exchange blocks, but also to sync themselves up, from genesis to to the tip of the blockchain. Why this is important is quite obvious: all the nodes need to have the same, consistent view of the world in order to operate. Some blockchains might impose a node to be 100% synced for certain operations to be possible, other might simply require things to be up-to-date up to a point. Not only that, but it might be quite handy for apps built on-top of a node, to have access to an API call like `/api/node/sync_progress` to get back the sync percentage, as well as the ETA for a full sync.

In a standard gossip setting, nodes would forward blocks to their closest peers when they create/receive them, but that works well only if all the nodes starts from the same block, and works a bit less well if a node is proactively trying to sync. In such scenario, a node would ideally like to request "all the blocks from genesis up to this block I have just received" or again "please send me all the blocks between X and Y".  However, in our current implementation doing that is not easy, because:

1. One cannot use the `Difficulty` to pinpoint block ranges, as that is not monotonically increasing and is subject to swings (it increases/stays the same/decreases every 2016 blocks);

2. Block hashes are not a very good candidate for that, as it doesn't exist a meaningfully ordering between them. A node cannot simply ask "give me all the nodes between `0befa98` and `0cadd97`). 

## Solution

*I propose to add the block height to the block header*.  This is something we talked about in the past, and that (I think) BTC also does (but they include the height in the coinbase transaction). This has several advantages:

* It allows for efficient queries in a relational database who has an index on the block height;
* It helps resolving https://github.com/oscoin/oscoin/issues/353 (where we had trouble with implementing pagination efficiently otherwise);
* It should fix https://github.com/oscoin/oscoin/issues/331 ;
* It will make node syncing (allegedly) more efficient. For example, each node could cache the "highest block" (that it receives unsolicited) and compute a diff between its current height (i.e. height of the current tip) vs its "best height" and then request via gossip those block(headers).
* Computing a node sync percentage is very easy now, it's essentially `last_known_best_height / local_height * 100`

### Implementation plan

[...]

PR #514, opened 2019-04-24
Add the height to the block header (Part I)

adinapoli-mndc:
This is a stepping stone in fixing #497.

This has to be considered "Part I" as it adds the height to the `BlockHeader` but it doesn't reap the benefits out of it (i.e. we can simplify and speed some of our SQLite queries as we now have an index on the `height` field).

I have decided, for review purposes, to include that in a subsequent PR (branching off either from this or from `master` if this PR ends up being merged quickly).

### DIFF
--- a/src/Oscoin/Crypto/Blockchain/Eval.hs
+++ b/src/Oscoin/Crypto/Blockchain/Eval.hs
@@ -80,13 +80,13 @@ buildBlock
     -> Timestamp
     -> st
     -> [tx]
-    -> BlockHash c
+    -> Block c tx s
     -> (Block c tx Unsealed, st, [Receipt c tx o])
-buildBlock eval tick st txs parentHash =
+buildBlock eval tick st txs parentBlock =
     let initialState = st
         (txOutputs, newState) = evalTraverse eval txs initialState
         validTxs = [tx | (tx, Right _) <- txOutputs]
-        newBlock = mkUnsealedBlock parentHash tick validTxs newState
+        newBlock = mkUnsealedBlock (Just parentBlock) tick validTxs newState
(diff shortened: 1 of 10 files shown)


src/Oscoin/Crypto/Blockchain/Eval.hs
```diff
@@ -179,39 +169,39 @@ evalToState eval tx = state go
...
+    :: (st -> Block c tx Unsealed)
     -> [tx]
-    -> BlockHash c
+    -> Evaluator st tx o
     -> st
     -> Either (tx, EvalError) (Block c tx Unsealed)
-mkEvaledUnsealedBlock eval tick txs parent initState =
-    mkUnsealedBlock parent tick txs <$> foldlM step initState txs
+mkEvaledUnsealedBlock createBlock txs eval initState =
+    createBlock <$> foldlM step initState txs
```
geigerzaehler: How about calling turning this into `evalEither txs initState = foldM step initState txs` and applying `createBlock` explicitly at the call site instead of requiring a callback?

adinapoli-mndc: Works for me (despite I don't necessarily see anything evil with the callback approach given all these functions are internal), as the intended purpose here was simply to alleviate the impedance mismatch between creating a "normal" unsealed block vs genesis, where we don't have a parent block at hand 😉

geigerzaehler: I just found the type signature hard to understand.

adinapoli-mndc: Done! 😉 04fc7743

PR #515, opened 2019-04-24
Modify SQLite.Internal queries to use the height

adinapoli-mndc:
This PR builds on top of #514 . It replaces most if not all the usages of `timestamp` in our SQLite queries with `height`. When not strictly necessary, I suspect it might help making things consistent on the long run (plus, there is an index on the `height`, which might speed up some queries in some circumstances).

### DIFF
--- a/src/Oscoin/Storage/Block/SQLite/Internal.hs
+++ b/src/Oscoin/Storage/Block/SQLite/Internal.hs
@@ -33,6 +33,7 @@ import           Oscoin.Crypto.Blockchain.Block
                  ( Block(..)
                  , BlockHash
                  , BlockHeader(..)
+                 , Height
                  , Sealed
                  , blockHash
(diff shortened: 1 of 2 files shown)


src/Oscoin/Storage/Block/SQLite/Internal.hs
```diff
@@ -171,7 +170,9 @@ storeBlock Handle{..} blk = runTransaction hConn $ do
 
   where
     extendsTip :: Block c tx (Sealed c s) -> Bool
-    extendsTip currentTip = blockHash currentTip == parentHash blk
+    extendsTip currentTip =
+        (blockHash currentTip            == parentHash blk) &&
```
cloudhead:
```suggestion
        (blockHash currentTip              == parentHash blk) &&
```
#ocd :face_with_thermometer:

PR #516, opened 2019-04-24
Add the `validateHeight` validator

### LINKED ISSUE #497: Add the block height to each block header

adinapoli-mndc:
Part 3 of my quest to fix #497. This PR adds the `validateHeight` validator, it uses it for both the Nakamoto "strict" and `lenient` validators and it allowed me to catch an oversight from our generators (i.e. a bug I have left as part of #514 , part I).

Fix #497.


src/Oscoin/Consensus/Types.hs
```diff
@@ -31,9 +31,14 @@ type Validate c tx s =
...
 
+type Expected a = a
+type Actual   a = a
+
 -- | A block validation error.
 data ValidationError c =
-      InvalidParentHash       (Crypto.Hash c)
+      InvalidHeight            (Expected Height) (Actual Height)
+    -- ^ Expected and actual height mismatch
+    |  InvalidParentHash       (Crypto.Hash c)
```
cloudhead:
```suggestion
    | InvalidParentHash       (Crypto.Hash c)
```

src/Oscoin/Consensus/Types.hs
```diff
@@ -31,9 +31,14 @@ type Validate c tx s =
     -> Block c tx (Sealed c s)       -- ^ Block to be validated.
     -> Either (ValidationError c) () -- ^ Either an error or @()@.
 
+type Expected a = a
+type Actual   a = a
+
 -- | A block validation error.
 data ValidationError c =
-      InvalidParentHash       (Crypto.Hash c)
+      InvalidHeight            (Expected Height) (Actual Height)
```
cloudhead:
```suggestion
      InvalidHeight           (Expected Height) (Actual Height)
```

src/Oscoin/Consensus/Validation.hs
```diff
@@ -2,21 +2,36 @@
...
 import           Oscoin.Consensus.Types
-import           Oscoin.Crypto.Blockchain
+import           Oscoin.Crypto.Blockchain hiding (height)
 import           Oscoin.Crypto.Blockchain.Block (Difficulty(..))
 import           Oscoin.Crypto.Hash (Hash)
 import           Oscoin.Time
 
 import           Control.Monad.Except (Except, throwError)
 
+-- | Rejects a block if its parent's height is smaller or equal than the
```
cloudhead: I think it's the other way around?

src/Oscoin/Consensus/Validation.hs
```diff
@@ -2,21 +2,36 @@
...
+-- input one.
+validateHeight
+    :: Block c tx s
+    -> Block c tx s
+    -> Except (ValidationError c) ()
+validateHeight parent blk
+    | height <- blockHeight (blockHeader blk)
+    , height <= parentHeight =
```
cloudhead: This would allow heights to be skipped - should be `height == succ parentHeight`

src/Oscoin/Consensus/Validation.hs
adinapoli-mndc: 🤦‍♂

src/Oscoin/Consensus/Validation.hs
adinapoli-mndc: Oops, yes -- this is what happens when you try to cram to many PRs in one day 😉 Sorry 'bout that!

adinapoli-mndc:
> Looks good, don't forget to fix the comment!

Oh, sorry -- I assumed you did as I did find some commits from you as part of this PR 😁  will do!

Issue #517, opened 2019-04-24
Proposal: Use tagged bools for command-line options


adinapoli-mndc:
## Problem

When working with CLI parsers and other functions, proliferations of boolean flags might lead to blindness if those bools are passed as function arguments (small example):

```haskell
doTopo
    :: UseColors
    -> Bool
    -> Bool
    -> PlanJson
    -> Bool
    -> Bool
    -> IO ()
```

## Solution/implementation plan

In a recent blog post [Oleg Grenrus](http://oleg.fi/gists/posts/2019-03-21-flag.html) proposes a simple "tagged" approach to create a `Flag` wrapper with a phantom type which can be used to cure boolean blindness, while not being too invasive in an existing code base.

We could introduce the same type and have things like:

```haskell

data EnableMDns = EnableMDns

data Options network = Options
    { optNetwork    :: network
    , optSeeds      :: [SeedAddr Crypto]
    , optSDDomains  :: [HostName]
    , optEnableMDns :: Flag EnableMDns
    , optNameserver :: Maybe (HostName, PortNumber) -- only for testing currently
    }
```

Which is arguably more descriptive, not so much in the _definition_ of the `Options` record, but in all those places where the boolean is passed/used without the record (if those places exist).

## (Optional) Alternative solutions considered

The alternative is to not press ahead with this proposal and recognise we don't have enough leverage at the moment to justify the introduction of this extra machinery.

If my grep-fu didn't fail me, here are the places we could clean up:

*  [Oscoin.P2P.Disco.Options](https://github.com/oscoin/oscoin/blob/master/src/Oscoin/P2P/Disco/Options.hs#L38) 
* [Oscoin.P2P.Handshake.Simple.whitelist](https://github.com/oscoin/oscoin/blob/c1f6d474249d2c3224f313ac7eff6da24fdd4394/src/Oscoin/P2P/Handshake/Simple.hs#L101)
* [Oscoin.Crypto.Blockchain.Block.Difficulty.decodeDifficulty](https://github.com/oscoin/oscoin/blob/f1e588220644e93e8a611198eece461c2ba5af19/src/Oscoin/Crypto/Blockchain/Block/Difficulty.hs#L113)

Essentially, every time a `Bool` appears in _negative_ position, we should be able to apply this transformation.

cc @kim @cloudhead @geigerzaehler

kim:
While I like this approach, I don't think we have enough `Bool`s flying around naked to really justify this currently:

[...]

PR #519, opened 2019-04-24
GCE discovery

kim:
Discovery of peers when running on GCE, using `compute.instances.aggregatedList`.

This is part of the grander plan to go with plain (container) instances vs. some ~~security and configuration management nightmare~~ scheduling system. Note that the utility (and necessity) is limited to the "devnet" environment, which will feature a relatively small number of instances, and high churn (continuous deployment, developer ad-hoc deployments). "Testnet" and beyond is expected to have a (mostly) stable set of seed nodes deployed as a separate instance group, so DNS discovery is less of a configuration management concern.

Pro tip: if you install the `btc` GHC plugin, you may be able to earn a block reward for compiling `gogol-compute`.

**Alternatives considered**

* Make instances "register" themselves with DNS on boot

   Requires coordination

* Create DNS RRs post-deployment

    More complex deployment (wait until instances are assigned network addresses), prevents autoscaling deployments

* Run a service which maintains DNS RRs

    Hard to justify for devnet, more complex deployment

* Sidecar init container which performs the lookup using a combination of `bash` and `gcloud` and writes some config or environment file

    Please.

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -48,6 +48,8 @@ library:
     - filepath
     - formatting
     - generics-sop
+    - gogol
+    - gogol-compute
     - gossip
     - hashable
     - hostname
@@ -79,6 +81,8 @@ library:
     - stm >=2.4.5.0
     - text
     - transformers
+    - unliftio-core
+    - unordered-containers
     - vector
     - wai
     - wai-middleware-static
--- a/src/Oscoin/P2P/Disco/GCE.hs
+++ b/src/Oscoin/P2P/Disco/GCE.hs
@@ -1,0 +1,123 @@
+{-# LANGUAGE DataKinds #-}
+
+module Oscoin.P2P.Disco.GCE where
+
+import           Oscoin.Prelude
+
+import           Oscoin.P2P.Types (Network, renderNetwork)
+
+import           Control.Monad.IO.Unlift (MonadUnliftIO)
+import           Data.HashMap.Strict (HashMap)
+import qualified Data.HashMap.Strict as Map
+import           Data.IP (IP)
+import qualified Data.Set as Set
+import qualified Data.Text as T
+import           Data.Text.Read as T
+import           Lens.Micro
+                 ( Getting
+                 , each
+                 , filtered
+                 , set
+                 , to
+                 , toListOf
+                 , traversed
+                 , (^?)
+                 , _Just
+                 )
+import           Lens.Micro.Extras (view)
+import           Network.Google
+                 (AllowScopes, Env, HasScope, runGoogle, runResourceT, send)
+import           Network.Google.Compute
+                 ( Instance
+                 , InstanceAggregatedList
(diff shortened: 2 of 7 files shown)


src/Oscoin/P2P/Disco/GCE.hs
```diff
@@ -0,0 +1,123 @@
...
+import           Network.Google.Compute
+                 ( Instance
+                 , InstanceAggregatedList
+                 , InstancesAggregatedList
+                 , NetworkInterface
+                 , i1Labels
+                 , i1NetworkInterfaces
+                 , ialItems
+                 , ialaFilter
+                 , ialiAddtional
```
adinapoli-mndc: Just for sanity check: is `Addtional` an abbreviation or some editor-smart-completion-propagated typo of `Additional`? 🤔

src/Oscoin/P2P/Disco/GCE.hs
```diff
@@ -0,0 +1,123 @@
...
+    runResourceT . runGoogle r $ do
+        instances <- toListOf inst <$> send (mkRq proj)
+        pure . Set.fromList . flip concatMap instances $ \i ->
+            let
+                iport = firstOf  (port . traversed) i
+                addrs = toListOf addr               i
+             in
+                zip addrs (repeat iport)
+  where
+    -- morally. microlens is just .. too micro
```
adinapoli-mndc: Don't even have proper prisms :trollface:

src/Oscoin/P2P/Disco/Options.hs
```diff
@@ -141,4 +142,8 @@ discoParser = Options
         ( long "enable-mdns"
        <> help "Enable mDNS discovery"
         )
+    <*> switch
+        ( long "enable-gce-sd"
```
adinapoli-mndc: And now we have two flags, maybe the case for #517 is stronger now? 😬

package.yaml
```diff
@@ -78,6 +80,8 @@ library:
     - stm >=2.4.5.0
     - text
     - transformers
+    - unliftio-core
```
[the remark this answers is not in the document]
kim: If you look closely, we already depend on the transitive closure of all of edkmett's and snoyman's libraries :P

src/Oscoin/P2P/Disco/GCE.hs
kim: It's fatal to make a typo in a code generator which spits out like 170 libraries 🤷‍♂

adinapoli-mndc: oh dear, so this comes from the library itself! 😅

src/Oscoin/P2P/Disco/GCE.hs
kim: Yeah what's the point really when you can't talk about a cocartesian coproduct?

PR #520, opened 2019-04-24
Use Ed25519 for digital signatures

### LINKED ISSUE #498: Schnorr signatures
## Problem

Space on-chain is precious, and ECDSA suffers from two main problems:

* ECDSA signatures are [malleable](https://en.wikipedia.org/wiki/Malleability_(cryptography));
* Their size in bytes is variable.

Switch to use [Schnorr signatures](https://github.com/sipa/bips/blob/bip-schnorr/bip-schnorr.mediawiki), that produces constant-sized 64-bytes signatures and are not malleable. On top of that, it allows multiple signatures to be combined together (without changing the final size) and therefore this makes it an excellent candidate for multi-sig schemes like [muSig](https://eprint.iacr.org/2018/068).

I have done a bit of research last week on the topic and from my understanding there are various components at play here:

1. There is the [original '91 paper](https://www.researchgate.net/profile/Claus_Schnorr/publication/227088517_Efficient_signature_generation_by_smart_cards/links/0046353849579ce09c000000.pdf) which is described in fairly generic algebraic groups and hashing functions;

2. The linked BIP proposal that specialises it for ECC (i.e. elliptic-curves-cryptography)

3. zero-knowledge protocols _based_ on Schnorr like [this](https://github.com/adjoint-io/schnorr-nizk) which I don't think they are useful to us.  

## Risks & unknowns

Something that it should be pointed out is that [EdDSA](https://github.com/haskell-crypto/cryptonite/issues/120#issuecomment-266657040) (that `cryptonite` implements) it's _based_ on Schnorr signatures , and it's likely what we are using at the moment. On top of that, if one goes and read the [RFC](https://tools.ietf.org/html/rfc8032) it's quite apparent (by looking at the [test vectors](https://tools.ietf.org/html/rfc8032#section-7)) that the signature size is pretty much the same as produced by the linked BIP (the curve we are interested is Ed25519). Therefore, one of the strong selling point of the BIP (i.e. space efficiency) doesn't seem to be an issue in our current implementation. Furthermore, the BIP seems to lock us into using a message of _fixed_ size (which is probably either the hash of some state or a merkle root hash), which is a bit limiting, frankly. 

What needs to be assessed though it's the malleability + composability properties, which might tip the scale towards implementing the BIP proposal vs keeping the current status quo. 

I haven't explored the domain long enough, but this seems to be what the state-of-the-art seems to be rapidly heading. Question remains around the proposal vs what we are using at the moment.

@kim @cloudhead @geigerzaehler thoughts?

adinapoli-mndc:
Fix #498.

The swapping was quite mechanism and modulo a wrinkle with laziness and deallocations, it touched only one module and it's a nice showcase of the effectiveness of our "swappable crypto" design.

What needs to be assessed now, if this PR effectively scored in terms of security & succinctness, so I suspect more work will need to be done to test effectively we are a) generating smaller (real world) signatures and that b) this is effectively "more secure". Can we think of a test that stresses malleability?


adinapoli-mndc:
Ok, I have now rebased against `master` and checked that thanks to the new scheme, our "upper bound size" went from 238 to 47 bytes, which is not too shabby at all. A production address looks like this, now:

```haskell
> import Oscoin.Crypto
> import Oscoin.Configuration 
> import Oscoin.Crypto.PubKey
> import Oscoin.Crypto.PubKey.RealWorld 
> import Oscoin.Crypto.Address
> (pk, sk :: PrivateKey Crypto) <- generateKeyPair 
> let addr = fromPublicKey Oscoin.Configuration.Mainnet pk
> renderAddress addr
"yryonyyyoeyfoebqrmqcmnwy8kbkaceyheewbqi5w3pa47hae7e1bf37siyioo5hmyqbrragdhbo"
```

Much more reasonable!

src/Oscoin/Crypto/PubKey/RealWorld.hs
```diff
@@ -26,34 +25,39 @@ import           Oscoin.Crypto.PubKey.Internal (PK(..), SK(..))
 
 instance HasDigitalSignature Crypto where
 
+    -- N.b. The uses of 'data' instead of a 'newtype' are essential to avoid
```
kim: I think this warrants some further investigation - or an "authoritative" explanation of why this works, and that it is indeed safe (as opposed to deferring the problem to later).

adinapoli-mndc: Totally -- let me open an issue for that.

adinapoli-mndc: https://github.com/oscoin/oscoin/issues/525

PR #521, opened 2019-04-26
Support multiple messages per transaction

cloudhead:
Allow transactions to contain more than one "message". This saves space on the ledger and allows for atomicity accross state changes.


src/Oscoin/Data/OscoinTx.hs
```diff
@@ -31,15 +31,15 @@ type Tx c = Tx' 1 0 c
 
 -- | A transaction with parametric version and flags.
 data Tx' (version :: Nat) (flags :: Nat) c = Tx'
-    { txPayload :: TxPayload c
+    { txMessages :: [TxMessage c]
```
kim: Should be `Set (TxMessage c)`, I would say.

cloudhead: Hmm you might be right, including the same message twice would more likely be an error than deliberate.

kim: Another issue is that the size of the list should be restricted.

cloudhead: > Another issue is that the size of the list should be restricted.

Thought about this, but I think it's easiest to let this be handled by tx byte-size validation (which is a todo in the code). We could impose a limit on the list length too, but it seems redundant if we have a limit on tx size. We're also going to be adjusting the tx fee based on the list length.

src/Oscoin/Data/OscoinTx.hs
```diff
@@ -70,13 +70,16 @@ data TxPayload c =
...
 -- | Transaction output, included in the receipts tree. (Placeholder)
-data TxOutput = TxOutput
+type TxOutput = [TxMessageOutput]
+
+-- | Transaction message output, included in 'TxOutput'.
+data TxMessageOutput = TxMessageOutput
     deriving (Show, Eq, Generic)
 
-instance CBOR.Serialise TxOutput
+instance CBOR.Serialise TxMessageOutput
```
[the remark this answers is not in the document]
cloudhead: Hehe yes.

cloudhead:
> I feel like the `Set vs List` discussion requires more thought (as demonstrated by our IRC convos), and perhaps it should be tackled outside this PR, which is a good starter.

:+1: - I think if we want to use `Set`, we need to think about how to offer the same flexibility as `[]`, which will not be easy.

Issue #432, opened 2019-03-26
Investigate correctness of Nakamoto.chainDifficulty upon difficulty adjustment


adinapoli-mndc:
This is a follow up from https://github.com/oscoin/oscoin/pull/426#issuecomment-476119411 .

In a nutshell, now that we switched to be using the SQL store for blocks persistence and we approach a difficulty change (at block 2016) the miner suddenly stops mining block.

We should start investigating what happens concretely in `Oscoin.Consensus.Nakamoto.chainDifficulty` at block 2016 and go into the rabbit hole trying to trace back the reason why the miner stops.

PR #527, opened 2019-05-02
Allow instrumentation of pure code (Trace & Probes)

adinapoli-mndc:
This PR delivers a simple instrumentation mechanism for pure code by introducing three types:

- A `Traced` (co)monad, suitable to annotate pure values with "tracing" (i.e. events);
- A `Probe`, which can be used to actually _emit_ the traced events;
- A `Traced` type alias which removes the need for the user to remember to pass `HasCallStack` everywhere.

## Problem description

While working on #432, we had this `chainDifficulty` function:

```haskell
-- | Calculate the difficulty of a blockchain.
chainDifficulty :: [Block c tx s] -> Difficulty
chainDifficulty [] =
    minDifficulty
chainDifficulty (NonEmpty.fromList -> blks)
    | NonEmpty.length blks `mod` fromIntegral difficultyBlocks == 0 =
        encodeDifficulty $ fst (decodeDifficulty currentDifficulty)
                         * fromIntegral targetElapsed
                         `div` toInteger actualElapsed
    | otherwise =
        prevDifficulty
```

This function is responsible to adjust the `Difficulty` of the network and, while it's nice and pure, it's not very telemetry-friendly. In practice, though, it would be extremely useful to be able to log & record "difficulty adjustments" as it would help us, before mainnet, to monitor the health of the consensus algorithm and to spot bugs. What can we do, then?

## Solution

We separate the concern of tracing events with the one of actually logging them, and this is what the `Traced` and `Probe` types do. The code now becomes:

```haskell
chainDifficulty :: [Block c tx s] -> Telemetry.Traced Difficulty
chainDifficulty [] =
    pure minDifficulty
chainDifficulty (NonEmpty.fromList -> blks)
    | NonEmpty.length blks `mod` fromIntegral difficultyBlocks == 0 = do
        let newDifficulty =
                encodeDifficulty $ fst (decodeDifficulty currentDifficulty)
                                 * fromIntegral targetElapsed
                                 `div` toInteger actualElapsed
        Telemetry.traced (DifficultyAdjustedEvent newDifficulty currentDifficulty) newDifficulty
    | otherwise =
        pure prevDifficulty
```

Apart from the monadic nature (which is unavoidable if we want to decorate our pure value), this is essentially unchanged, modulo the addition to this `Telemetry.traced` call, which reminds, in spirit, `Debug.Trace`. The difference though is that **we don't do evil IO** and, in fact, our `Traced` type is completely pure:

```haskell
newtype Traced a =
    Traced (StateT (OldestFirst [] NotableEvent) Identity a)
    deriving (Functor, Applicative, Monad)
```

What we gain now is that we can collect events in this monadic context, and we can later on use a `Probe` to emit the events. But how to do that without:

1. Changing the code everywhere;
2. Actually performing IO?

We can use a `Tracer` function, defined as:

```
type Tracer m = HasCallStack => forall a. Traced a -> m a
```

And simply pass this function to any function that want to probe something deep down the guts of the application. Note how this can live in any monad `m` and therefore we don't need to commit to any concrete one just yet. Finally, at the very top-level of our app, we can create a `Probe` and call `probed` to get a `Tracer` out of it:

```haskell
            let probe   = Telemetry.telemetryProbe telemetry
                        & Telemetry.hoistProbe liftIO
            let consensus =
                    case optEnvironment of
                        Production  ->
                            Consensus.nakamotoConsensus (Telemetry.probed probe)
                        Development ->
                            Consensus.nakamotoConsensusLenient (Telemetry.probed probe) optBlockTimeLower
```

When this code is run, it would produce something like:

```
monadic/oscoin (adinapoli/trace-and-probes) > stack exec oscoin -- --genesis data/genesis.yaml --blockstore blockstore.db
info  | node starting env=development logical_network=devnet physical_network=xysnx4b7s-zfocl-ky genesis=2DrjgbCd1rxCgmw24kS8YxJeR9esBPCQdDZ3i29gWRwUt7Vkbw loc="Oscoin.Node:81:13"
Spock is running on port 8477
info  | bootstrapping peer_addr=127.0.0.1:6942 peer_nodeid=2DrjgbLeNjsmAYKS4Lj96mqiRktsqgYnPpwGQyqxGmJqnBJMvU ns="gossip" loc="Oscoin.P2P:134:10"
info  | bootstrapped peer_addr=127.0.0.1:6942 peer_nodeid=2DrjgbLeNjsmAYKS4Lj96mqiRktsqgYnPpwGQyqxGmJqnBJMvU ns="gossip" loc="Oscoin.P2P:134:10"
info  | difficulty adjustment new=Difficulty {fromDifficulty = 539799905} old=Difficulty {fromDifficulty = 553713663} ns="consensus" loc="Oscoin.Consensus.Nakamoto:140:14"
```

## Limitations

Unfortunately we cannot get precise-to-the-events `SrcLoc`, but rather we would get the `SrcLoc` associated to the last `probed` call within the codebase. In practice this should be just fine, but if we want to do better, I feel like we would need to overhaul a bit our logging/telemetry system: the reason is that we can have multiple events we trace (cfr. `NotableEvent`) but at the end of the day we probably will have fewer `emit/probed` call within the code (for efficiency/composability reasons) and therefore the `CallStack` will be fixed by the location where `emit/probed` is called. To fix this properly we would probably need to decorate our `NotableEvent` GADT with a `HasCallStack` constraint for each type constructor, and then modify `emit/withLogger` accordingly. This is a lot of work, so that's not a priority for now.

## Other concerns

If we find ourself using `Traced` in anger, we should keep an eye on the memory efficiency -- i.e., making sure this doesn't leak memory. I have tried my best to use strict functions where it matters, but a sanity check down the line won't hurt.

### DIFF
--- a/src/Oscoin/Consensus/Nakamoto.hs
+++ b/src/Oscoin/Consensus/Nakamoto.hs
@@ -67,7 +67,7 @@ nakamotoConsensus
        , Monad m
        , Serialise tx
        )
-    => (forall a. Telemetry.Traced a -> m a)
+    => Telemetry.Tracer m
     -> Consensus c tx PoW m
 nakamotoConsensus probed = Consensus
     { cScore = comparing chainScore
@@ -132,7 +132,7 @@ mineNakamoto
        , Hashable c (BlockHeader c (Sealed c PoW))
        , Hashable c (BlockHeader c Unsealed)
        )
-    => (forall a. Telemetry.Traced a -> m a)
+    => Telemetry.Tracer m
     -> (forall tx. [Block c tx (Sealed c PoW)] -> Telemetry.Traced Difficulty)
     -> Miner c PoW m
 mineNakamoto probed difiFn getBlocks unsealedBlock = do
--- a/src/Oscoin/Consensus/Nakamoto.hs
+++ b/src/Oscoin/Consensus/Nakamoto.hs
@@ -23,6 +23,8 @@ import           Oscoin.Consensus.Validation
 import           Oscoin.Crypto.Blockchain
 import           Oscoin.Crypto.Blockchain.Block (Difficulty(..))
 import           Oscoin.Crypto.Hash (Hash, Hashable)
+import qualified Oscoin.Telemetry as Telemetry
+import           Oscoin.Telemetry.Events
 import           Oscoin.Time
 
 import           Codec.Serialise (Serialise)
@@ -65,10 +67,11 @@ nakamotoConsensus
        , Monad m
        , Serialise tx
        )
-    => Consensus c tx PoW m
-nakamotoConsensus = Consensus
+    => (forall a. Telemetry.Traced a -> m a)
+    -> Consensus c tx PoW m
+nakamotoConsensus probed = Consensus
     { cScore = comparing chainScore
-    , cMiner = mineNakamoto chainDifficulty
+    , cMiner = mineNakamoto probed chainDifficulty
     , cValidate = validateFull
     }
 
@@ -85,7 +88,7 @@ validateFull [] blk =
 validateFull prefix@(parent:_) blk = runExcept $ do
     validateHeight     parent blk
     validateParentHash parent blk
-    validateDifficulty chainDifficulty prefix blk
+    validateDifficulty (Telemetry.tracing_ . chainDifficulty) prefix blk
     validateTimestamp  parent blk
     validateBlockAge
     liftEither (validateBasic blk)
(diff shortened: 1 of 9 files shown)


src/Oscoin/Consensus/Nakamoto.hs
```diff
@@ -182,16 +187,18 @@ difficultyBlocks :: Depth
...
-    | NonEmpty.length blks `mod` fromIntegral difficultyBlocks == 0 =
-        encodeDifficulty $ fst (decodeDifficulty currentDifficulty)
-                         * fromIntegral targetElapsed
-                         `div` toInteger actualElapsed
+    | NonEmpty.length blks `mod` fromIntegral difficultyBlocks == 0 = do
+        let newDifficulty =
+                encodeDifficulty $ fst (decodeDifficulty currentDifficulty)
+                                 * fromIntegral targetElapsed
+                                 `div` toInteger actualElapsed
+        Telemetry.traced (DifficultyAdjustedEvent newDifficulty currentDifficulty) newDifficulty
```
cloudhead: I think this changes the meaning of the function, and should be moved out to a different function: `chainDifficulty` is meant to return the target chain difficulty - it isn't concerned about adjustment events, as it is meant to be used simply to calculate a value. Adding telemetry to it confuses its purpose. Furthermore, since difficulty is adjusted every 2016 blocks regardless of whether this function is called or not, the event should be fired every 2016 blocks from whichever place is receiving or producing blocks, not from a function which calculates difficulty of an input chain (which may not be the full chain). Thoughts? @kim @adinapoli-mndc

kim: > I think this changes the meaning of the function

Huh?

cloudhead: Example: if I call this function to calculate the difficulty of a fork, I get a `DifficultyAdjustedEvent` - which would be wrong.

cloudhead: > > I think this changes the meaning of the function
> 
> Huh?

This function now assumes a certain usage - which is that if the returned difficulty is a multiple of 2016, it should trigger an event that should be observable from our logs/metrics. This is only useful if used on every block and only on the main chain..

kim: We can argue about the meaning of the event, but I fail to see any wrongness nor alteration of the meaning of the function.

PR #528, opened 2019-05-02
Fix chain difficulty adjustment 

adinapoli-mndc:
Tentative fix for #432. There were 2 bugs that caused the miner to stop mining:

* The `mine` function within `mineNakamoto` was wrong (as a certain Italian programmer broke it during a refactoring 😅 ). In particular, we were never "twiddling" the input `BlockHeader` with the computed `PoW` nonce, so after the 2016th block we were simply spinning this loop ad-infinitum without never sealing the block;

* The `validateDifficulty` function needed to prepend the input block to the blocks passed to the `chainDifficulty` function, as the expected difficulty needed to take into account the incoming block which was triggering the difficulty adjustment.

I have run the miner and I can now see that the difficulty changes, also thanks to the support of the "trace & probes" machinery (see #527):

```
monadic/oscoin (adinapoli/issue-432●) > stack exec oscoin -- --genesis data/genesis.yaml --blockstore blockstore.db 2>&1 | grep difficulty
info  | difficulty adjustment new=4d840f2c old=0151052b direction=increasing ns="consensus" loc="Oscoin.Consensus.Nakamoto:138:14"
info  | difficulty adjustment new=5d285b2d old=4d840f2c direction=increasing ns="consensus" loc="Oscoin.Consensus.Nakamoto:138:14"
info  | difficulty adjustment new=396f3c2f old=5d285b2d direction=increasing ns="consensus" loc="Oscoin.Consensus.Nakamoto:138:14"
info  | difficulty adjustment new=fffc2731 old=396f3c2f direction=increasing ns="consensus" loc="Oscoin.Consensus.Nakamoto:138:14"
info  | difficulty adjustment new=50571a33 old=fffc2731 direction=increasing ns="consensus" loc="Oscoin.Consensus.Nakamoto:138:14"
info  | difficulty adjustment new=68d01135 old=50571a33 direction=increasing ns="consensus" loc="Oscoin.Consensus.Nakamoto:138:14"
info  | difficulty adjustment new=19140c37 old=68d01135 direction=increasing ns="consensus" loc="Oscoin.Consensus.Nakamoto:138:14"
info  | difficulty adjustment new=7bf40739 old=19140c37 direction=increasing ns="consensus" loc="Oscoin.Consensus.Nakamoto:138:14"
info  | difficulty adjustment new=6bc9043b old=7bf40739 direction=increasing ns="consensus" loc="Oscoin.Consensus.Nakamoto:138:14"
info  | difficulty adjustment new=6bcb023d old=6bc9043b direction=increasing ns="consensus" loc="Oscoin.Consensus.Nakamoto:138:14"
```

I still think there is something fishy going on in this area as the seal for blocks from `2016 .. 4032` seems to be > 0, but then the difficulty swings back to a high value and we keep getting a seal of 0 with a lots of cheap blocks being generated. Probably we might need to review the implementation of `chainDifficulty` again in the future, but at least as far as #432 is concerned, we should be in business now.


src/Oscoin/Consensus/Nakamoto.hs
```diff
@@ -200,6 +202,16 @@ chainDifficulty (NonEmpty.fromList -> blks)
     | otherwise =
         pure prevDifficulty
   where
+
+    -- The difficulty is adjusted every 'difficultyBlocks' blocks, and only if we do have
+    -- 2016 blocks at hand to compute the new one. As the 'height' of genesis
+    -- is 0, we need to add 1 to the value returned by 'blockHeight', to be
+    -- sure to trigger an adjustment every 'difficultyBlocks' blocks.
+    adjustmentRequired =
+        (bHeight (NonEmpty.head blks) + 1 ) `mod` fromIntegral difficultyBlocks == 0
```
cloudhead:
```suggestion
        (bHeight (NonEmpty.head blks) + 1) `mod` fromIntegral difficultyBlocks == 0
```

src/Oscoin/Consensus/Nakamoto.hs
```diff
@@ -200,6 +202,16 @@ chainDifficulty (NonEmpty.fromList -> blks)
...
+        && NonEmpty.length blks == fromIntegral difficultyBlocks
```
cloudhead: Looks good, I think this is correct. Another way to formulate this would be "block height > 0 && block height % 2016 == 0`

src/Oscoin/Consensus/Nakamoto.hs
```diff
@@ -147,9 +145,13 @@ mineNakamoto probed difiFn getBlocks unsealedBlock = do
   where
     mine :: BlockHeader c Unsealed -> PoW -> Maybe PoW
     mine hdr (PoW nonce)
-        | hasPoW hdr        = Just (PoW nonce)
-        | nonce < maxBound  = mine hdr (PoW (nonce + 1))
-        | otherwise         = Nothing
+        | hasPoW (sealHeader (PoW nonce) hdr) = Just (PoW nonce)
```
cloudhead: can we use `pow@(PoW nonce)` to not repeat this here?

src/Oscoin/Consensus/Nakamoto.hs
adinapoli-mndc: Yeah, possibly -- in my formulation I wanted to stress the fact that non only we need to trigger this every 2016 blocks, but that we likely want also _exactly_ (at least?) `difficultyBlocks` blocks as input, in order to make an informative decision (otherwise we cannot compute a proper `rangeEnd` and `rangeStart`).

src/Oscoin/Consensus/Nakamoto.hs
adinapoli-mndc: maybe, I can try 😉

src/Oscoin/Consensus/Nakamoto.hs
adinapoli-mndc: 386578cc

src/Oscoin/Consensus/Nakamoto.hs
adinapoli-mndc: 386578cc

PR #529, opened 2019-05-02
Display a more useful error when the keypair cannot be deserialised from disk 

### LINKED ISSUE #526: Extremely poor error when keys cannot be read from file
## Problem

If the oscoin keypair gets compromised for whichever reason and the `oscoin` client fails to deserialise it from disk, the user is greeted with an extremely poor and confusing error (essentially the raw CBOR failure):

```
monadic/oscoin (adinapoli/issue-432) > stack exec oscoin -- --genesis data/genesis.yaml --blockstore blockstore.db    
oscoin: DeserialiseFailure 0 "end of input"
```

It took me a while to realise what was going on!

Prettify the error, which can span from simply embellishing the deserialisation failure with a cue of what's going on to some human-friendly error, with steps to solve the problem.

adinapoli-mndc:
Fixes #526.

There are gazillion different ways to achieve the same goal, but I have settled on the cheapest one: we simply rewrap the CBOR error in a richer type, and we override the way the exception is displayed. Theoretically we can avoid the `displayException = show` line as this is the default anyway, but whatever.

We could get fancy with pretty errors etc, but at least this now gives a very tangible cue (especially within the codebase) of what's going on:

```
monadic/oscoin (adinapoli/issue-526●) > stack exec oscoin -- --genesis data/genesis.yaml --blockstore blockstore.db   
oscoin: Loading the oscoin's private key from the configuration folder failed. The reason was: DeserialiseFailure 0 "end of input"
```

Variation on the theme might include:

* Turn `DeserialiseFailure 0 "end of input"` into something like "Your file seems to be empty, have you tried regenerating the keypair?"

* Pick the shed's paint colour you like 😏


src/Oscoin/CLI/KeyStore.hs
```diff
@@ -61,6 +63,21 @@ ensureConfigDir mbUserPath = do
     configDir <- getConfigPath mbUserPath
     liftIO $ createDirectoryIfMissing True configDir
 
+data KeyStoreError =
+    PrivateKeyDeserialisationFailure DeserialiseFailure
+
+instance Show KeyStoreError where
```
kim: nit: In the spirit of `read . show = id`, shouldn't this be just a derived `Show`, and `displayException` showing the descriptive error?

src/Oscoin/CLI/KeyStore.hs
```diff
@@ -61,6 +63,21 @@ ensureConfigDir mbUserPath = do
...
+    show = \case
+      PrivateKeyDeserialisationFailure cborError ->
+          mconcat [
+                    "Loading the oscoin's private key from the configuration "
```
kim:
```suggestion
                    "Loading the private key from the configuration "
```

kim: Actually, it might be useful to stick the path to the "configuration folder" in the `PrivateKeyDeserialisationFailure` ctor

src/Oscoin/CLI/KeyStore.hs
adinapoli-mndc: That would be my preference as well. However, there were some turbo-shenanigans when I've tried that (i.e. it seemed like the exception was being shown via `show` rather than `displayException`). But this deserves a more accurate try this morning, with a semi-working brain 😁

src/Oscoin/CLI/KeyStore.hs
adinapoli-mndc: True, let me do that!

src/Oscoin/CLI/KeyStore.hs
adinapoli-mndc: Alright, unless my reptilian brain is deceiving me, I indeed tried again the above, and given this code:

```haskell
data KeyStoreError =
    PrivateKeyDeserialisationFailure DeserialiseFailure

instance Show KeyStoreError where
    show = \case
      PrivateKeyDeserialisationFailure cborError ->
          mconcat [
                    "Loading the oscoin's private key from the configuration "
                  , "folder failed. The reason was: "
                  , show cborError
                  ]

instance Exception KeyStoreError where
    displayException _ = "TODO"
```

It doesn't work as expected -- the exception is shown via `show` it seems, rather than `displayException`. (i.e. I don't see "TODO"). If I can speculate, As `displayException` has been added fairly recently (I think from base 4.8 onwards if I am not mistaken) the exception handler installed in the main thread doesn't display the exception via `displayException` but rather via `show`. I can surely adjust things to behave as we want, if we are OK with the slightly more invasive patch.

adinapoli-mndc:
@kim Seized the chance for unifying the error to be generic (i.e. doesn't distinguish between private or public keys) and to include the proper configuration path. I think this should be cleaner now (modulo the `displayException` shenanigans).

src/Oscoin/CLI/KeyStore.hs
kim: meh, dissapointing

PR #535, opened 2019-05-06
Restore chainDifficulty purity

adinapoli-mndc:
This PR removes the `Traced` annotation from `chainDifficulty`, while leaving the `probed` call inside `mineNakamoto`. This means that we get slightly imprecise `CallStack`, but we do not compromise on purity for important functions like `chainDifficulty`.


src/Oscoin/Consensus/Nakamoto.hs
```diff
@@ -131,11 +131,18 @@ mineNakamoto
...
 mineNakamoto probed difiFn getBlocks unsealedBlock = do
     blks  <- getBlocks difficultyBlocks
-    diffy <- probed (difiFn blks)
+    let prevDifficulty = blockTargetDifficulty
+                       . blockHeader
+                       . NonEmpty.head
+                       . NonEmpty.fromList
+                       $ blks
+    let diffy = difiFn blks
+    when (prevDifficulty /= diffy) $
```
cloudhead: This is new actually, and makes sense - since we weren't checking that the difficulty changed.

src/Oscoin/Consensus/Nakamoto.hs
```diff
@@ -131,11 +131,18 @@ mineNakamoto
        , Hashable c (BlockHeader c (Sealed c PoW))
        )
     => Telemetry.Tracer m
-    -> (forall tx. [Block c tx (Sealed c PoW)] -> Telemetry.Traced Difficulty)
+    -> (forall tx. [Block c tx (Sealed c PoW)] -> Difficulty)
     -> Miner c PoW m
 mineNakamoto probed difiFn getBlocks unsealedBlock = do
     blks  <- getBlocks difficultyBlocks
-    diffy <- probed (difiFn blks)
+    let prevDifficulty = blockTargetDifficulty
```
cloudhead: This suggests the difficulty is about to change - maybe `currentDifficulty` makes more sense

adinapoli-mndc: Agreed.

src/Oscoin/Consensus/Nakamoto.hs
```diff
@@ -131,11 +131,18 @@ mineNakamoto
...
+                       . blockHeader
+                       . NonEmpty.head
+                       . NonEmpty.fromList
```
cloudhead: Do we actually need to convert to NonEmpty?

adinapoli-mndc: one way or the the other we had to perform an unsafe operation, so I did stick to exactly what we do for the `prevDifficulty` inside `chainDifficulty`. Perhaps a better design would be to pass to the miner directly a NonEmpty?

src/Oscoin/Consensus/Nakamoto.hs
adinapoli-mndc: 3bd77183

src/Oscoin/Consensus/Nakamoto.hs
adinapoli-mndc: I mean, modulo turning this into a `Maybe` and then doing some unwrapping down the line, but if we have a bug here (i.e. an empty list), this will fail exactly one line below as well, as part of the call to `difiFn` -- imho the correct solution is to pass a `NonEmpty` in the first place, but this is outside the scope of this PR, methinks 😉

PR #537, opened 2019-05-07
Extend the Gossip handshake with more information

### LINKED ISSUE #536: Extend the gossip protocol to advertise its HTTP API port during handshake
## Problem

In order to implement node syncing (see #531) we need a way to get our active peers and use the HTTP API to fetch the information we need. We already have a `getPeers` and `getPeers'` functions available, but they return (at most) the `SockAddr` the gossip protocol is communicating on, which is hardly suitable for HTTP traffic (modulo clunky and buggy multiplexing). 

We should extend the handshake mechanism between peers to also announce the port the HTTP API is listening to. This way, we can try reusing the existing `SockAddr` the `Peer` data structure provide to kickoff HTTP traffic.

## Blockers

(cc @kim) Currently a `SockAddr` has these constructors:

```haskell
SockAddrInet !PortNumber!HostAddress |  
-- | --
SockAddrInet6 !PortNumber!FlowInfo !HostAddress6!ScopeID |  
SockAddrUnix String
``` 

This means that if we pattern match on the already-available `SockAddr` and we get a `SockAddrUnix`, we are out of luck. I do wonder if it would be better to extend our `Peer` type with something like:

```haskell
data Peer n = Peer
    { peerNodeId :: n
    , peerAddr   :: SockAddr
    , peerHttpAPI :: SockAddr
   } deriving (Eq, Show)
```

So that we don't have to do any guessing.

adinapoli-mndc:
~Nothing too exciting to see yet, needs moar tests~ (I have added the tests), but this is ready for an early review to have a sanity check whether this approach is sensible or not.

Fix #536 .

### DIFF
--- a/exe/node/Main.hs
+++ b/exe/node/Main.hs
@@ -156,6 +156,7 @@ main = do
             let myHost     = P2P.numericHost optHost
             let myApiInfo  = P2P.mkAddr myHost optApiPort
             let myNodeInfo = P2P.mkNodeInfo myApiInfo nid
+            let mySK       = snd keys
             gossip <- managed $
                 withGossip telemetry
                            P2P.BootstrapInfo
@@ -166,7 +167,7 @@ main = do
                            (Set.map (Nothing,) seeds)
                            (storage node)
                            (Handshake.secureHandshake
-                               keys
+                               mySK
                                myApiInfo
                                (P2P.Disco.optNetwork optDiscovery'))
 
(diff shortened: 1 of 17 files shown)


adinapoli-mndc:
@kim apart the unification of `infoExchange` and `keyExchange`, which proved to be trickier than expected due to the use of the inner guard, the rest should hopefully get your holy mark of approval. I didn't change the name of the `Proxy` back to `crypto` for the reasons I have explained, but I don't care too much -- if you have a strong preference, can amend that back 😉

adinapoli-mndc:
@kim Ok, your code merged and roundtrip tests added -- let me know if you are happy with the current shape 😉

PR #538, opened 2019-05-07
Make (node) options invertible

kim:
i.e. render CLI options from a Options value

It is desirable for certain deployment configurations to be able to safely override the configuration of the deployed executable. For this, we need to be able to render the set of CLI options which fully constitute the configuration from the configuration value itself.

Ideally, the (partial) isomorphisms would be built up field-wise  à la [invertible syntax](http://www.informatik.uni-marburg.de/~rendel/unparse/). I have, however, not yet found a satisfying formulation. Of course, we could just define suitable functions `a -> [Opt Text]`, yet this is quite error prone. Reading the latest in [~~hipster~~higher kinded data](https://chrispenner.ca/posts/hkd-options) experiments, it occured to me that would be quite nice syntactically and mnemonically to stick functorial "unparsers" in a record's fields, even though not actually "safe". Doing the same with the parsers is also quite cute, although not strictly required for this patch.

Unfortunately, the `barbies` (!!1) library chokes on Sandy's type-family-of-fame, which eliminates the boring `Identity` functor, and I can't at the moment be bothered to get this to work with `generics-sop` (because a boggling of the mind). Since options usually get scrutinised early, it seems bearable.

### DIFF
--- a/exe/node/Main.hs
+++ b/exe/node/Main.hs
@@ -15,7 +15,6 @@ import           Oscoin.Data.Tx
 import           Oscoin.Node (runNodeT, withNode)
 import qualified Oscoin.Node as Node
 import qualified Oscoin.Node.Mempool as Mempool
-import           Oscoin.Node.Options (Options(..))
 import qualified Oscoin.Node.Options as Node
 import           Oscoin.P2P (mkNodeId, runGossipT, withGossip)
 import qualified Oscoin.P2P as P2P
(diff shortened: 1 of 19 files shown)


exe/node/Main.hs
```diff
@@ -45,35 +44,57 @@ import           Options.Applicative
 type GenesisBlock =
     Block Crypto (Tx Crypto) (Sealed Crypto Nakamoto.PoW)
 
+
+unI :: Identity a -> a
+unI = runIdentity
```
adinapoli-mndc: 👻 😁

src/Oscoin/P2P/Disco.hs
```diff
@@ -102,18 +102,18 @@ withDisco tracer opt defaultGossipPort !advertise k = do
         DNS.makeResolvSeed
             . maybe identity
                     (\ns -> \rc -> rc { DNS.resolvInfo = uncurry DNS.RCHostPort ns })
-                    (optNameserver opt)
+                    (runIdentity (optNameserver opt))
```
adinapoli-mndc: worthwhile moving `unI` inside `Oscoin.Prelude` so we can reuse it here as well?

test/Test/Oscoin/Configuration.hs
```diff
@@ -43,3 +59,16 @@ prop_roundtripNetworkShowRead :: Property
 prop_roundtripNetworkShowRead = property $ do
     net <- forAll Gen.enumBounded
     tripping net (show @Network) readEither
+
+prop_roundtripPathsCLI :: Property
+prop_roundtripPathsCLI = property $ do
```
adinapoli-mndc: It might be nice a short description (in prose, perhaps with some inlined code) of what this test is supposed to return, even just to have a "wow factor" for this _tour de force_ 😉

test/Test/Oscoin/Node/Options.hs
```diff
@@ -0,0 +1,45 @@
...
+    [ testProperty "prop_roundtripOptions" (prop_roundtripOptions d)
+    ]
+
+props :: Dict (IsCrypto c) -> IO Bool
+props d = checkParallel $ Group "Test.Oscoin.Node.Options"
+    [ ("prop_roundtripOptions", prop_roundtripOptions d)
+    ]
+
+prop_roundtripOptions :: forall c. Dict (IsCrypto c) -> Property
+prop_roundtripOptions Dict = property $ do
```
adinapoli-mndc: I guess same comment about inline-documentation as above 😁

src/Oscoin/Configuration.hs
```diff
@@ -23,64 +30,93 @@ module Oscoin.Configuration
...
     -- Default: $oscoin_datadir/data/genesis.yaml
+    } deriving ( Generic
+               , FunctorB
+               , TraversableB
+               , ProductB
+               , ConstraintsB
+               , ProductBC
+               )
+
+deriving instance AllBF Show f Paths => Show (Paths f)
```
adinapoli-mndc: I would imagine that writing such parsers via `barbies` (!!) is going to be (in practice) a mechanical process, but having a few documentation lines about how one should approach writing a new CLI `Parser` wouldn't harm, here.

test/Test/Oscoin/P2P/Gen.hs
```diff
@@ -75,19 +77,22 @@ genHostname = do
...
-genIP :: Gen IP
+genIP :: MonadGen m => m IP
 genIP = Gen.choice [ ipv4, ipv6 ]
   where
     ipv4 = IPv4 . toIPv4 <$> replicateM 4 (Gen.int (Range.constant 0 255))
     ipv6 = IPv6 . toIPv6 <$> replicateM 8 (Gen.int (Range.constant 0 65535))
 
-genMsg :: IsCrypto c => Gen (Msg c Text (Sealed c Text))
+genPortNumber :: MonadGen m => m PortNumber
+genPortNumber = fromIntegral <$> Gen.word16 Range.constantBounded
```
adinapoli-mndc: Stealing this temporarily while working on #536 😍

src/Oscoin/P2P/Disco.hs
kim: No. Should we move to `generics-sop`, the types won't match.

kim:
> document
> document
> prose
> documentation

Mm kay. As this is all very obvious to me, and it's not actually very sophisticated (most of the work was to get the test suite going for you testing-obsessed folks): how about _you_ propose some documentation where you feel it is missing?

adinapoli-mndc:
> how about you propose some documentation where you feel it is missing?

Yup, can do -- no reason to keep this PR hanging just for that though, we can do this in a separate strand of work.

kim:
me cago en la leche, but I need to control ordering for dumber option parsers than `optparse-applicative`, so I can't reuse this approach elsewhere. Thus reverting to just functions for simplicity's sake (the pattern matches + rountrip tests give the same level of confidence).

adinapoli-mndc:
por favor, señor! (will review in a bit, 😁 )

PR #540, opened 2019-05-08
Oscoin Tx and Core types - Part 3.

cloudhead:
A small PR with a few updates:
* Fix and test project registration/unregistration
* Handle txBurn
* Handle txNonce


test/Test/Oscoin/Data/OscoinTx.hs
```diff
@@ -64,7 +70,10 @@ propApplyTx Dict = property $ do
     fee <- forAll $ genFee tx bal
 
     -- Should succeed because Alice can pay the minimum fee
-    assertRight $ applyTx (tx { txFee = fee }) alice ws'
+    (ws'', _) <- evalEither $ applyTx (tx { txFee = fee }) alice ws'
+
+    -- Alice's nonce is now incremented to `1`.
+    lookupNonce alice ws'' === Right 1
```
geigerzaehler: I think it would be nice if we had a generative test for the nonces. I.e. that after applying a tx to an account with an arbitrary nonce the nonce is incremented.

cloudhead: Was a bit hard to do in this PR so I added it to #544: https://github.com/oscoin/oscoin/pull/544/commits/c0b51eadfc8a88dc0e32a018f456089237922c90

PR #544, opened 2019-05-09
Split 'Tx' into 'Tx' and 'TxPayload' / Test nonce

cloudhead:
* Rename Tx to TxPayload
 * Create new Tx type which includes signature
 * Move txNetwork to new Tx type
 * Move version to new Tx type
 * Remove flags


src/Oscoin/Data/OscoinTx.hs
```diff
@@ -364,7 +367,7 @@ lookupEpoch ws =
         _               -> Left (ErrTypeMismatch "epoch")
 
 -- | The minimum transaction fee that is valid for the given transaction.
-minimumTxFee :: Tx c -> Balance
+minimumTxFee :: TxPayload c -> Balance
 minimumTxFee = const 1
```
adinapoli-mndc: This is something for the future, but it would be nice to think _if_ we want to name our unit something different from oscoin (cfr. 1 Satoshi vs 1 BTC). This just because as it stand is very ambiguous -- one might say this is "1 oscoin", but that would probably be wrong 😉

cloudhead: Yeah, been thinking about this but I don't know :D ... micro-oscoin?

adinapoli-mndc: Naming is hard 😉 I was thinking we could do some brainstorming related to open-source contribution, something which would encapsulate the notion of "smallest open source contribution possible", like a "diff" or a "patch", but trying to avoid ambiguous, English names -- perhaps we can play with those and come up with a unique name 😀

PR #546, opened 2019-05-10
Add the getTip HTTP endpoint.

### LINKED ISSUE #545: Add getTip as an HTTP endpoint
## Problem

In order to implement #531 , it's useful to have an HTTP endpoint to request the tip of the blockchain for a node. While potentially this could be implemented user-side in terms of `getBestChain`, it's rather tedious as the latter returns a list of blocks, and as such calling `Data.List.head` is a partial operation, and calling `headMay` introduce a needless layer of error checking, as a blockchain is defined as a non-empty list of blocks, and thus fetching the tip should always succeed.

Implements a new HTTP endpoint (e.g. GET `/blockchain/tip`) which returns the tip of the chain.

See the problem section for a nonoptimal alternative solution.

adinapoli-mndc:
Fixes #545.

This PR introduces a new HTTP endpoint `/blockchain/tip` which can be used to retrieve the current tip of the queried node.


test/Oscoin/Test/API/HTTP.hs
```diff
@@ -38,6 +38,9 @@ tests Dict =
     , testGroup "GET /blockchain/best"
         [ test "No depth specified" (getBestChain @c)
         ]
+    , testGroup "GET /blockchain/tip"
+        [ test "it works" (getTip @c)
```
geigerzaehler: I’d prefer a more expressive name, like “returns blockchain tip”.

adinapoli-mndc: cc7f836e

PR #547, opened 2019-05-13
Add beneficiary to blocks

cloudhead:
Add the *beneficiary* to the block. That is, the account that should be receiving block rewards.

This just adds the ability to have a block beneficiary, but does not dispense any rewards. That will be a separate PR.

A new flag is introduced: `--beneficiary`, which takes an `Address` that is parsed and converted into a `PublicKey` and stored in the node `Config`. It is then passed to the miner for inclusion in blocks.

A new data type is introduced: `BlockData`, which encapsulates beneficiary and transactions.
The beneficiary (a `PublicKey`) is hashed alongside the transactions, for tamper-evidence. This means that now, `blockDataHash = merklehash(beneficiary || txs)`.


src/Oscoin/P2P/Class.hs
```diff
@@ -7,5 +7,5 @@ import           Oscoin.P2P.Types (Msg)
 
 import           Codec.Serialise (Serialise)
 
-class Monad m => MonadBroadcast c m where
+class Monad m => MonadBroadcast c tx m where
     broadcast :: (Serialise s, Serialise tx, Hashable c tx) => Msg c tx s -> m ()
```
cloudhead: I couldn't get things compiling without adding this type parameter unfortunately.

test/Oscoin/Test/Crypto/Blockchain/Block/Helper.hs
```diff
@@ -0,0 +1,15 @@
...
+import           Oscoin.Crypto.PubKey (HasDigitalSignature, PublicKey)
+
+import           System.IO.Unsafe (unsafePerformIO)
+import           Test.QuickCheck
+
+-- | Generates an arbitrary beneficiary.
+defaultBeneficiary
+    :: (Arbitrary (PublicKey c), HasDigitalSignature c, HasHashing c) => PublicKey c
+defaultBeneficiary = unsafePerformIO $ generate arbitrary
+{-# NOINLINE defaultBeneficiary #-}
```
cloudhead: This seemed like the least intrusive way to add the beneficiary to the tests for now. When we implement a "burn" address, we can switch it to that, and remove the unsafe code.

data/blockstore.sql
```diff
@@ -12,9 +12,8 @@ CREATE TABLE IF NOT EXISTS "blocks" (
   "difficulty"    integer       NOT NULL,
   "seal"          blob          NOT NULL,
 
-  -- NOTE: The "timestamp" and "parenthash" fields carry an invariant that is
-  -- not checked here: the timestamp of a row must always be greater than
-  -- the timestamp of the row referenced by the "parenthash" field.
```
cloudhead: This invariant is no longer necessary, now that we use *height* to get blocks. It's also not true in practice.

src/Oscoin/P2P/Class.hs
geigerzaehler: You can fix this by adding `Serialise (PubKey c)` to the constraints of `broadcast`

test/Oscoin/Test/Crypto/Blockchain/Block/Helpers.hs
```diff
@@ -0,0 +1,14 @@
...
+import           Oscoin.Crypto.PubKey (PublicKey)
+
+import           System.IO.Unsafe (unsafePerformIO)
+import           Test.QuickCheck
+
+-- | Generates an arbitrary beneficiary.
+defaultBeneficiary
+    :: Arbitrary (PublicKey c) => PublicKey c
+defaultBeneficiary = unsafePerformIO $ generate arbitrary
+{-# NOINLINE defaultBeneficiary #-}
```
geigerzaehler: I have some concerns about this. It introduces randomness everywhere in the tests where we use this. The randomness is not controlled by the quickcheck seed which means we can never replay this.

I see two solutions to this problem
* Lift every function that uses `defaultBeneficiary` to `Gen` and generate the beneficiary
* Choose a fixed public key that is the same for every run. (I guess this would require another method on the crypto type class.)

test/Test/Oscoin/Crypto/Address/Gen.hs
```diff
@@ -0,0 +1,22 @@
...
+import           Oscoin.Test.Crypto.PubKey.Arbitrary
+import           Test.Oscoin.Crypto.Address (genAddress)
+
+import           Hedgehog
+import qualified Hedgehog.Gen as Gen
+import qualified Hedgehog.Range as Range
+
+import qualified System.Random.SplitMix as SplitMix
+
+address :: MonadGen m => (HasHashing c, HasDigitalSignature c) => m (Address c)
```
geigerzaehler:
```suggestion
address :: (MonadGen m, HasHashing c, HasDigitalSignature c) => m (Address c)
```

geigerzaehler: Shouldn’t we use `genAddress` by convention? Shouldn’t `Test.Oscoin.Cyrpto.Address.genAddress` also live in this module?

src/Oscoin/Crypto/Blockchain/Block.hs
```diff
@@ -225,15 +230,67 @@ type StateHash crypto = Crypto.Hash crypto
...
+    , blockDataTxs         :: Seq tx
+    } deriving Generic
+
+blockTxs :: Block c tx s -> Seq tx
+blockTxs = blockDataTxs . blockData
+
+blockBeneficiary :: Block c tx s -> Crypto.PublicKey c
+blockBeneficiary = blockDataBeneficiary . blockData
+
+instance (Serialise tx, Serialise (Crypto.PublicKey c)) => Serialise (BlockData c tx) where
```
geigerzaehler: Could you add roundtrip tests for this?

test/Test/Oscoin/Crypto/Address/Gen.hs
cloudhead: Looking at some of the other `*.Gen` modules, I think you're right - there's only one module that doesn't do this - `Config.Gen` :thinking:

test/Oscoin/Test/Crypto/Blockchain/Block/Helpers.hs
cloudhead: The second solution is what I was going to do with the *burn address*. The problem is currently we have no way of generating a fixed public key, unless we start from a fixed address. So the plan was to work on that in a separate PR.

You're right about the uncontrolled randomness, but since we aren't actually inspecting this value anywhere yet, or using it, it has no effect on tests at the moment. As soon as we have the burn address working, I'll remove this.

test/Test/Oscoin/Crypto/Address/Gen.hs
cloudhead: 0b172de

src/Oscoin/Crypto/Blockchain/Block.hs
cloudhead: 72e5d64

src/Oscoin/P2P/Class.hs
cloudhead: 3fbe6cf

PR #551, opened 2019-05-14
Always publish artifacts

kim:
The same SHA1 doesn't get rebuilt, so fast-forward merges to master won't actually publish any artifacts. Our latest Doge image is from Apr 11.

It is anyways generally useful to be able to spin up a non-master container image, and _potentially_ to look at API docs (because Haddock markup is hard). I would, however, also be fine to not generate/publish Haddocks at all, as nobody is looking at them.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -84,19 +84,16 @@ steps:
     - |
       set -euo pipefail
 
-      if [ "$BRANCH_NAME" == "master" ]; then
-        stack build \
-          --test \
-          --flag=oscoin:static \
-          --no-terminal \
-          --pedantic \
-          --copy-bins \
-          --local-bin-path=/workspace/bin \
-          --haddock \
-          --exec="mv $(stack path --local-doc-root) api-docs"
-      else
-        stack build --test --no-terminal --pedantic
-      fi
+      stack build \
+        --test \
+        --flag=oscoin:static \
+        --no-terminal \
+        --pedantic \
+        --copy-bins \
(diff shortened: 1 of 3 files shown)


kim:
Alright, not sure if you noticed, but the `master` build was mostly broken for at least the past month. The issue is basically due to haddocks: by default, `stack` builds the docs for all dependencies, too, so all links work. This means that it would rebuild all dependencies, regardless of the cached ones, and the build simply times out.

I have now disabled dependency haddocks, which makes them less useful, but since I already caught some invalid markup it is possibly worth to keep building docs in case we want to publish them at some point.

Takeaway: never do anything only on master. Nobody's even looking.

geigerzaehler:
I’d be ok with removing the documentation build
* Nobody is looking at the published version
* Without links to documentation in packages the usability of the docs is limited
* It takes longer to build and the build logs are cluttered with haddock warnings.

kim:
@geigerzaehler I've heard "the build takes too long" from both you and @cloudhead a few times, but I don't think this is an excuse for doing less - the slowness is mainly a consequence of our build and CI systems. It is possible to keep the build between 2-4min with some effort.

The issue with disabling docs entirely is that we then basically give up on them forever - if the build doesn't fail on invalid markup, and we don't publish the documentation anywhere, then why bother writing it in the first place?

I do think we should put some effort into our build system, as we're now approaching the stage of every non-trivial Haskell project, where simply running `stack build` or `cabal new-build` doesn't cut it anymore.

PR #555, opened 2019-05-15
Move transaction validation to Mempool

### LINKED ISSUE #553: Move transaction validaiton to the Mempool
## Problem

Transaction validation is the process of checking that a transaction is properly signed. Invalid transactions should be rejected by the Ledger API and not make their way into the Mempool. Currently the logic responsible for that lives in `Oscoin.API.HTTP.Handlers.submitTransaction`. This blocks #552 and does not guarantee that the Mempool does not contain invalid transaction.

I propose to do transaction validation as part of the insertion into the Mempool. `Mempool.addTxs :: [tx] -> m ()` will be replaced with
~~~haskell
class Mempool m c tx where
  addTx :: tx -> m (Either (TxValidationError c tx) ())
~~~

`TxValidationError c tx` is a type family similar to `TxOutput`. The Mempool will be instantiated with the transaction validation.

geigerzaehler:
Before, transaction validation only happened in the API Handlers when a transaction was submitted. We now validate a transaction when it is inserted into the Mempool.

We replace the `Mempool.addTxs` with `Mempool.addTx`. Inserting multiple txs at once was never properly used.

Now, the API does not know how a transaction is validated. It only gets the result of that validation.

We also start rejecting invalid transactions from gossip.

Fixes #553


src/Oscoin/API/HTTP/Handlers.hs
```diff
@@ -69,26 +69,19 @@ submitTransaction
...
-  where
-    getVerifiedTxBody = do
-        tx <- getBody
-        if verifyTx tx
-        then pure tx
-        else respond badRequest400 $ errBody "Invalid transaction signature"
+    tx <- getBody
+    addResult <- node $ Mempool.addTx tx
+    case addResult of
+        Left _ -> respond badRequest400 $ errBody "Invalid transaction signature"
```
cloudhead: Would be better if this message just came from the Mempool, since we can't know what the error was.

src/Oscoin/API/HTTP/Handlers.hs
```diff
@@ -69,26 +69,19 @@ submitTransaction
...
+        Right _ -> do
+            telemetryStore <- node $ view telemetryStoreL
+            forM_ [ TxsAddedToMempoolEvent [Crypto.fromHashed $ Crypto.hash @c tx]
+                  , TxSubmittedEvent (Crypto.hash @c tx)] $
```
cloudhead: I wonder if we should trigger a `TxSubmittedEvent` on failed validation too? @adinapoli-mndc

src/Oscoin/Node/Mempool.hs
```diff
@@ -31,39 +31,36 @@ import qualified Data.Foldable as Fold
...
-    -> STM ()
-insertMany Handle{hMempool, hBroadcast} txs = do
-    STM.modifyTVar' hMempool (Internal.insertMany txs)
-    STM.writeTChan hBroadcast (Insert (Fold.toList txs))
+    -> tx
+    -> STM (Either (TxValidationError c tx) ())
+insert Handle{hMempool, hBroadcast, hTxValidate} tx = runExceptT $ do
+    ExceptT $ pure $ hTxValidate tx
+    lift $ STM.modifyTVar' hMempool (Internal.insert tx)
+    lift $ STM.writeTChan hBroadcast (Insert [tx])
```
cloudhead: Since only `hTxValidate` returns an Either here, wouldn't it be better to just `case` on that, and remove the `runExceptT`` machinery and lifts?

src/Oscoin/Node/Mempool/Class.hs
```diff
@@ -8,13 +8,15 @@ module Oscoin.Node.Mempool.Class
...
+import           Oscoin.Data.Tx (TxValidationError)
 import           Oscoin.Node.Mempool.Event (Channel, Event(..))
 
 -- MonadMempool ---------------------------------------------------------------
 
 class Monad m => MonadMempool c tx m | m -> tx, m -> c where
-    -- | Add transactions to the mempool, notifying all subscribers.
-    addTxs :: Foldable t => t tx -> m ()
+    -- | Add a transaction to the mempool if it is valid.
+    -- Notifies all subscribers.
```
cloudhead:
```suggestion
    -- If valid, notifies all subscribers.
```

src/Oscoin/API/HTTP/Handlers.hs
adinapoli-mndc: My inclination would be to say 'yes', alongside with the relevant "validation failed" event. We can later tune down the verbosity of telemetry if that turns out to generate too much traffic 😉

src/Oscoin/API/HTTP/Handlers.hs
adinapoli-mndc: Also, out of loud thought: when we will overhaul our API to be RESTful, we should look at ways to make errors consumable by machines, rather than spitting out an unstructured and unparsable string (but for now, whatever). In the past I have used the [jsend](https://github.com/omniti-labs/jsend) format, but if you know of any better semi-de-facto formats to encode errors, let me know.

src/Oscoin/Storage.hs
```diff
@@ -94,7 +94,11 @@ applyTx
 applyTx bs tx = do
     let txHash = Crypto.hash tx
     novel <- isNovelTx bs txHash
-    if | novel     -> Mempool.addTxs [tx] $> Applied [TxAppliedEvent txHash]
+    if | novel     -> do
+            addResult <- Mempool.addTx tx
+            case addResult of
+                Left _  -> pure $ Error []
+                Right _ -> pure $ Applied [TxAppliedEvent txHash]
```
adinapoli-mndc: It's a fixation, but I would be in favour of writing `Right ()`, so that the minute we will (ever) change the result of the `TxValidator` from `()` to something more structured, the compiler will bark at us.

src/Oscoin/Storage.hs
```diff
@@ -94,7 +94,11 @@ applyTx
 applyTx bs tx = do
     let txHash = Crypto.hash tx
     novel <- isNovelTx bs txHash
-    if | novel     -> Mempool.addTxs [tx] $> Applied [TxAppliedEvent txHash]
+    if | novel     -> do
+            addResult <- Mempool.addTx tx
+            case addResult of
+                Left _  -> pure $ Error []
```
adinapoli-mndc: It's kind of a shame we are throwing a way the error and not spitting out any telemetry event -- can we add one?

src/Oscoin/Node/Mempool/Class.hs
```diff
@@ -8,13 +8,15 @@ module Oscoin.Node.Mempool.Class
...
+    addTx :: tx -> m (Either (TxValidationError c tx) ())
```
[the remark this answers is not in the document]
geigerzaehler: But this would prevent us from defining generic instances of `Mempool` that work with any `tx`. For example we would need to provide two instances for `Tx` and `OscoinTx` which would have the exact same method definitions.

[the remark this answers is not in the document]
adinapoli-mndc: Oh, right, sorry, still half awake. Yeah, the fundeps here are not really doing us a favour, as they are locking in place the `tx` given the concrete monad 😏

src/Oscoin/API/HTTP/Handlers.hs
geigerzaehler: @cloudhead I agree this is not ideal. I would follow @adinapoli-mndc’s suggestion and think about API errors more holistically at another point. I created an issue for that https://github.com/oscoin/oscoin/issues/558

geigerzaehler:
Please take another look, @adinapoli-mndc and @cloudhead. Please also consider [@kim’s alternative suggestion](https://github.com/oscoin/oscoin/issues/553#issuecomment-492603757)

src/Oscoin/Storage.hs
geigerzaehler: https://github.com/oscoin/oscoin/pull/555/commits/03cd7f8d81783a06a3a24bf3b90ffea3fb75fcb7

src/Oscoin/Node/Mempool.hs
geigerzaehler: https://github.com/oscoin/oscoin/pull/555/commits/03cd7f8d81783a06a3a24bf3b90ffea3fb75fcb7

src/Oscoin/API/HTTP/Handlers.hs
geigerzaehler: https://github.com/oscoin/oscoin/pull/555/commits/03cd7f8d81783a06a3a24bf3b90ffea3fb75fcb7

PR #562, opened 2019-05-17
Keep Discoing

kim:
If no peers where found during startup, keep trying.

Additionally, `gogol` has mined me enough object files to be able to retire on the Caymans, bye! :grimacing:


kim:
Depends on oscoin/gossip#26

package.yaml
```diff
@@ -48,8 +48,6 @@ library:
     - filepath
     - formatting
     - generics-sop
-    - gogol
-    - gogol-compute
```
cloudhead: :pray:

src/Oscoin/P2P.hs
```diff
@@ -1,5 +1,8 @@
+{-# LANGUAGE NumericUnderscores   #-}
```
adinapoli-mndc: 😎 Best GHC 8.6 extension! 😉

PR #563, opened 2019-05-20
Reward transaction fees to beneficiary

cloudhead:
Transaction fees are deducted from the author and transfered to the beneficiary.

### DIFF
--- a/src/Oscoin/Data/OscoinTx.hs
+++ b/src/Oscoin/Data/OscoinTx.hs
@@ -145,10 +145,12 @@ applyTxPayload
        )
     => TxPayload c
     -- ^ The transaction to apply.
+    -> AccountId c
+    -- ^ The account of the beneficiary, where fees should be sent.
     -> WorldState c
     -- ^ The input world state.
     -> Either (TxError c) (WorldState c, TxOutput)
-applyTxPayload tx@TxPayload{..} ws = do
+applyTxPayload tx@TxPayload{..} beneficiary ws = do
     Account{..} <- lookupAccount txAuthor ws
 
     when (txNonce /= accountNonce) $
@@ -160,13 +162,13 @@ applyTxPayload tx@TxPayload{..} ws = do
     when (txFee + txBurn > accountBalance) $
         Left (ErrInsufficientBalance accountBalance)
 
-    -- TODO(cloudhead): Transfer fee
+    ws' <- transferBalance txAuthor beneficiary txFee ws
+
     -- TODO(cloudhead): Verify transaction size
 
     applyTxMessages txMessages txAuthor
         =<< debitAccount txAuthor txBurn
-            (adjustAccount incrementNonce txAuthor ws)
-
+            (adjustAccount incrementNonce txAuthor ws')
 
 -- FIXME(adn) This is currently a stub.
 verifyTx :: Tx c -> Bool
@@ -267,8 +269,7 @@ applyTxMessage (TxTransfer receiver _) sender _
     | sender == receiver =
         Left (ErrNotAuthorized sender)
 applyTxMessage (TxTransfer receiver bal) sender ws = do
-    ws'  <- debitAccount  sender   bal ws
-        >>= creditAccount receiver bal
+    ws' <- transferBalance sender receiver bal ws
 
     pure (ws', [])
 
@@ -282,6 +283,21 @@ insertAccount acc@Account{..} = WorldState.insert (accountKey accountId) (Accoun
 -- Nb. The caller must ensure that the value under the account id is indeed an 'Account'.
 deleteAccount :: CBOR.Serialise (PublicKey c) => AccountId c -> WorldState c -> WorldState c
 deleteAccount id = WorldState.delete (accountKey id)
+
+-- | Transfer balance between two accounts. Uses 'creditAccount' and 'debitAccount'.
+transferBalance
+    :: (CBOR.Serialise (PublicKey c))
+    => AccountId c
+    -- ^ Transfer sender ("from").
+    -> AccountId c
+    -- ^ Transfer receiver ("to").
+    -> Balance
+    -- ^ Transfer amount.
+    -> WorldState c
+    -- ^ Input state.
+    -> Either (TxError c) (WorldState c)
+transferBalance a b bal ws =
+    debitAccount a bal ws >>= creditAccount b bal
 
 -- | Credit an account with coins.
 --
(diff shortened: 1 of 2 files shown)


src/Oscoin/Data/OscoinTx.hs
```diff
@@ -283,6 +284,17 @@ insertAccount acc@Account{..} = WorldState.insert (accountKey accountId) (Accoun
 deleteAccount :: CBOR.Serialise (PublicKey c) => AccountId c -> WorldState c -> WorldState c
 deleteAccount id = WorldState.delete (accountKey id)
 
+-- | Transfer balance between two accounts. Uses 'creditAccount' and 'debitAccount'.
+transferBalance
+    :: (CBOR.Serialise (PublicKey c))
+    => AccountId c
+    -> AccountId c
```
adinapoli-mndc: Apart from using a `newtype`, which would make working with these heavier, can you think of any other way we can prevent these two `AccountId` from being swapped? I can totally foresee how that can happen, and when our code will grow complex we will probably soon forget "which one was which" if we don't recall the type signature.

test/Test/Oscoin/Data/OscoinTx.hs
```diff
@@ -113,30 +114,36 @@ propMultiTransfer Dict = property $ do
...
 
     lookupBalance bob ws' === Right 4
 
 propTxFeeBurn :: forall c. Dict (IsCrypto c) -> Property
 propTxFeeBurn Dict = property $ do
-    alice <- publicKey @c
-    tx    <- pure (mkTxPayload [] alice)
-    bal   <- forAll genBalancePositive
-    ws    <- evalEither $ creditAccount alice bal WorldState.empty
+    (alice, eve)
```
geigerzaehler: Having Eve, the evil eavesdropper, roaming our tests makes me slightly uncomfortable ;)

test/Test/Oscoin/Data/OscoinTx.hs
```diff
@@ -103,7 +103,8 @@ propTransferBalance Dict = property $ do
 
 propMultiTransfer :: forall c. Dict (IsCrypto c) -> Property
 propMultiTransfer Dict = property $ do
-    (alice, bob)  <- publicKeyPair @c
+    (alice, bob, eve) <- publicKeyTriple @c
```
adinapoli-mndc: Out-loud comment: it's quite a bummer in recent version of GHCs we cannot use (easily) the (unsafe) partial pattern match without having the monad we are in satisfy `MonadFail`. In particular, creating all these `publicKeyPairXX` functions feels wasteful, when one could have had:

```haskell
publicKeys :: forall c. (IsCrypto c) => PropertyT IO [PublicKey c]
```

and later:

```haskell
[alice,bob,eve] <- take 3 <$> publicKeys
```

test/Test/Oscoin/Data/OscoinTx.hs
```diff
@@ -113,30 +114,36 @@ propMultiTransfer Dict = property $ do
...
     lookupBalance bob ws' === Right 4
 
 propTxFeeBurn :: forall c. Dict (IsCrypto c) -> Property
 propTxFeeBurn Dict = property $ do
-    alice <- publicKey @c
-    tx    <- pure (mkTxPayload [] alice)
-    bal   <- forAll genBalancePositive
-    ws    <- evalEither $ creditAccount alice bal WorldState.empty
+    (alice, eve)
+          <- publicKeyPair @c
```
adinapoli-mndc: Do you need the line break for some strange OCD reason? 🤔

test/Test/Oscoin/Data/OscoinTx.hs
```diff
@@ -113,30 +114,36 @@ propMultiTransfer Dict = property $ do
...
+
+    annotate . condensedS $ (accountKey alice, accountKey eve)
+
+    tx   <- pure (mkTxPayload [] alice)
```
adinapoli-mndc: A `let` suffices? 🤔

src/Oscoin/Data/OscoinTx.hs
geigerzaehler: Could you document which account is the sender and which is recipient to avoid misuse?

test/Test/Oscoin/Data/OscoinTx.hs
cloudhead: Hahaha yeah I was thinking "what comes after bob?"

test/Test/Oscoin/Data/OscoinTx.hs
cloudhead: :joy: yes actually - to align the `<-`s

test/Test/Oscoin/Data/OscoinTx.hs
cloudhead: I'm kind of disturbed by `let` without an `in` - so I usually use bind syntax in do-blocks. :thinking:

test/Test/Oscoin/Data/OscoinTx.hs
cloudhead: Actually @geigerzaehler added `MonadFail` to hedghehog's `Property`, so we should be able to do that pretty soon! (well, when Stackage updates, which might be a while)

src/Oscoin/Data/OscoinTx.hs
cloudhead: That's part of the :game_die: :trollface:

cloudhead: Hmm - can't think of any way, though the current order encodes the directionality of the transfer, ie. `From -> To`. I expect anyone using this function to triple-check :smile:

cloudhead: 8edca46

cloudhead: Added some doc to the params, should help.

test/Test/Oscoin/Data/OscoinTx.hs
geigerzaehler: Maybe Charlie? Or Ben the beneficiary?

adinapoli-mndc: Yeah, maybe I'll go with Charlie -- using `eve` it might even be misleading, like suggesting we are testing against adversarial behaviour 😮

cloudhead: Haha fair enough.

cloudhead: b5384c2

Issue #552, opened 2019-05-15
Make HTTP API polymorphic in TX


geigerzaehler:
## Problem

Currently the HTTP API (`Oscoin.API.HTTP.Handlers`) only works with the fixed `Tx` from `Oscoin.Data.Tx`. For the proper Ledger API (#549) we need the HTTP API to work with `OscoinTx`.

For testing it is beneficial to run the HTTP API not with the Oscoin ledger but with a more simple ledger with less complex state transitions.

## Solution/implementation plan

The code in `Oscoin.API.HTTP.Handlers` (and related modules) will be made polymorphic in the transaction. This should be easy to accomplish.

The main blocker is currently `verifyTx` used in `Oscoin.API.HTTP.Handlers.submitTransaction` which is not polymorphic in `tx`. In fact, this function does not exist for `OscoinTx`. As a solution I propose to move the transaction verification to the `Mempool`.

## Alternative solutions considered

Instead of making the API polymorphic we could just switch to the new transaction type. This would require us to rewrite all the tests using this type and make the tests more complicated in the future.

## Dependencies

- [x] Move transaction validation to the Mempool #553

PR #564, opened 2019-05-20
Make API polymorphic in 'tx'

geigerzaehler:
The code in `Oscoin.API.HTTP.Handlers` and related modules is now 
 polymorphic in the transaction type. See #552.

 As part of this change we also eliminate the `getAllTransactions` endpoint.
 It’s implementation did not use a serialisation of a transaction list but
 rather a serialisation of the Mempool instance. This coupled the API to the
 concrete mempool instance used by the node.

### DIFF
--- a/src/Oscoin/Crypto/Blockchain/Block.hs
+++ b/src/Oscoin/Crypto/Blockchain/Block.hs
@@ -5,6 +5,7 @@
 module Oscoin.Crypto.Blockchain.Block
     ( -- * Types
       Block -- opaque to disallow construction of sealed blocks.
+    , SealedBlock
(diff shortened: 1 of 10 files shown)


src/Oscoin/Crypto/Blockchain/Block.hs
```diff
@@ -237,6 +238,8 @@ data Block c tx s = Block
     , blockData   :: BlockData c tx
     } deriving Generic
 
+type SealedBlock c tx s = Block c tx (Sealed c s)
```
adinapoli-mndc: Nice, I guess it saves some typing 😉

Issue #550, opened 2019-05-14
Add lookupHeight and lookupHeightRange to both the Blockstore reader and HTTP APIs


adinapoli-mndc:
## Problem

In order to implement #531 , we do need a way to fetch blocks from peers, given a block `Height` or a range. This corresponds to the `lookupByHeight` and `lookupByHeightRange` from the spec, but we don't have these functions yet, neither in the `BlockStoreReader` nor in the HTTP API.

## Solution/implementation plan

Even though our internal interface might change later (i.e. by introducing a `BlockTree`), we do want to get syncing implemented relatively soon, which means the most pragmatic approach is to implement these two functions directly on the existing `BlockStore`. As written in the specs, we can even emulate `lookupByHeightRange` using the minimal API and change things later.

We do also want to expose the relevant endpoints in our HTTP API, so that clients can consume it.

### A note on the HTTP API

If we were to do things properly (ala RESTful way) we would ideally expose a `GET /blocks` endpoint which would return the first 50 (or so, upper limit to be decided) blocks and then implement these two queries in terms of filters, as in:

```
GET /blocks?height=GTE[10]&height=LTE[20]
```

(Note how the above would also allow us to scrap `/blockchain/best` in favour of a more resource-centric API).

However, doing this properly is a non-trivial amount of work and I am not sure we have time for it now, but if we disregard the complication of a fully-fledged query system for the filtering, we could _at least_ implement an ad-hoc filter like `height[10,20]` and perhaps when we have a bit more capacity sit at the table and implement this properly.

@cloudhead @geigerzaehler @kim thoughts?

geigerzaehler:
👍 on exposing `lookupByHeight` and `lookupByHeightRange` on the API.

Regarding the HTTP API I would forgo the a generic RESTful implementation. Since we only have this one way of querying blocks it would lead to bikeshedding about the parameter encoding and lots of code for parameter encoding and decoding. I’d suggest and endpoint of the form `/blocks/by-height-range?minHeight=10&maxHeight100` or whatever you like to call it.

cloudhead:
:+1: 

I wouldn't go crazy on the REST, but an adhoc filter like you suggest might be a nice middle ground.

adinapoli-mndc:
[...]

Issue #531, opened 2019-05-03
Node syncing


adinapoli-mndc:
_This is a work in progress ticket not ready for review yet. I am merely dumping some ideas around to avoid losing work due to some funky UI misbehaviour 😁_.

## Problem

When a node "joins late to the party" (highly technical term 😎 ), it needs to catch up with the blocks it has missed. This comes in two flavours (one is a specific case of a more general one):

1. A node it's joining the network for the first time (i.e. a new user) and it needs to download all the blocks from genesis up to the current tip;

2. A node it's joining the network after being offline for a while, so it needs to catch up on all the blocks he missed, as well as taking into accounts forks, if any (i.e. a fork might have occurred while it was offline, and now it needs to rollback first and start tracking the new chain).

## Existing solutions

There are multiple ways to perform syncing, and this is very much an interesting problem, as it's possibly one of the first usability challenge we have to face.

### The BTC way

We could start by understanding what BTC does, and reason how we can adapt this to our gossip network. The algorithm is described fairly succinctly in _Mastering Bitcoin_, Chapter 8, [Exchanging Inventory](https://github.com/bitcoinbook/bitcoinbook/blob/develop/ch08.asciidoc#exchanging-inventory), page 181. In a nutshell, when a new node `NewPeer` joins the network, it sends out a [version](https://en.bitcoin.it/wiki/Protocol_documentation#Message_types) message and it gets back the version of all the nodes it will communicate back with (in theory, this could be the root of our epidemic tree). The `version` contains the `BestHeight` for that node, so that `NewPeer` will know which is the current tip of the chain (this info might or might not be 100% accurate, but at least offers as starting point). At this point though, unless I am misreading the book, it gets fairly stateful, in particular, a passage says:

> The peer that has the longer blockchain has more blocks than the other node and can identify which blocks the other node needs in order to "catch up."

This seems to establish a "coupling" between nodes and making the syncing process fairly stateful, and I am not sure how well this would translate into our gossip scheme.

## Ethereum

I haven't researched the actual algorithm, but it seems like the bottleneck is not on the block download but on state materialisation (see [1]). 

## Ethereum (Geth)

Things seems to have improved with their "fast syncing" algorithm, for which there is [a PR](https://github.com/ethereum/go-ethereum/pull/1889) with a fantastic write-up. However, it comes with its own tradeoffs, which we will need to understand and explore.

## Ethereum (Parity)

[...]

PR #567, opened 2019-05-21
Node syncing - Part 1 (Types, abstractions, sim monad and basic tests)

adinapoli-mndc:
This PR ships the first part of #531 , and follows the spec outlined [here](https://hackmd.io/2cPkrWTjTIWo2EmHZpPIQw?both). 

It's unfortunately another almost-1k-big PR, but the commit should be hopefully be reviewable in isolation and I am happy to pair-review this with people. As the PR is already quite substantial but definitely more stuff needs to be implemented, I think it's beneficial if we start with the basics. I wouldn't antagonise too much on the performance side of things as I am aware this is pretty much just a reference implementation and I didn't even try to be efficient here 😬 .

## What's in here

- [x] Basic types, abstractions and data structures;
- [x] Mostly monad-agnostic syncing algorithm;
- [x] Two concrete implementations, one for a `Sim` monad and another one for `IO/GossipT`;
- [x] Tests for the syncing over `Sim`.

## What is (obviously) missing

- [ ] Any form of wiring to the `Protocol`, in order to actually kick off syncing at node startup.

## What is (not-so-obviously) missing

- [ ] A real implementation for the IO data fetcher, which requires #550 , but subsequent PRs can fake this using an `IORef` for now;
- [ ] Some kind of support for streaming blocks efficiently from the data fetcher to the guts of the syncing. I have to think about this a bit more but we have tests to catch any regression in this area;
 - [ ] Performance optimisations when it comes to not only parallelism but also the algorithm to handle missing blocks;
- [ ] Tests for the `IO` implementation -- I think I can generalise a bit what I have now to reuse pretty much the same tests without all the boilerplate (but boilerplate might do for now);
- [ ] Telemetry and logging -- for now we don't log/emit anything substantial;
- [ ] Use the `OscoinTx` rather than the `Tx` in the `IO` `DataFetcher`.

### DIFF
--- a/src/Oscoin/Prelude.hs
+++ b/src/Oscoin/Prelude.hs
@@ -44,6 +44,8 @@ module Oscoin.Prelude
     , module X
     , module Control.Exception.Safe
 
+    , Actual
+    , Expected
     , String
     , LText
     , LByteString
@@ -461,3 +463,8 @@ notImplemented = withFrozenCallStack (error "Not implemented")
 {-# WARNING undefined "'undefined' remains in code" #-}
 undefined :: (HasCallStack) => a
 undefined = withFrozenCallStack (error "Prelude.undefined")
+
+-- Simple type aliases that makes easier for the reader to distinguish, in
+-- a type signature, which is the expected vs actual value.
+type Expected a = a
+type Actual   a = a
--- a/src/Oscoin/Protocol/Sync.hs
+++ b/src/Oscoin/Protocol/Sync.hs
@@ -48,6 +48,7 @@ import           Oscoin.Time.Chrono
 
 import           Data.HashSet (HashSet)
 import qualified Data.HashSet as HS
+import           Data.List ((\\))
 import           Data.List.Split (chunksOf)
 import qualified Data.Map.Strict as M
 import qualified Data.Set as Set
@@ -69,6 +70,11 @@ deriving instance ( Eq (Hash c)
                   , Eq s
                   , Eq (PublicKey c)
                   ) => Eq (SyncEvent c tx s)
+deriving instance ( Ord (Hash c)
+                  , Ord tx
+                  , Ord s
+                  , Ord (PublicKey c)
+                  ) => Ord (SyncEvent c tx s)
 deriving instance ( Show (Hash c)
                   , Show tx
                   , Show s
@@ -301,6 +307,7 @@ getRemoteHeight
     => Sync c tx s m Height
 getRemoteHeight = map height getRemoteTip
 
+-- | Given a valid 'Range', syncs the blocks within the range.
 syncBlocks
     :: ( ProtocolResponse c tx s 'GetBlocks ~ OldestFirst [] (Block c tx (Sealed c s))
        , Monad m
@@ -316,27 +323,61 @@ syncBlocks rng = do
     let fetcher = scDataFetcher ctx
     withActivePeers $ \ activePeers -> do
 
-      let workDistribution =
-              zip (cycle (HS.toList activePeers))
-                  (chunksOf maxBlocksPerPeer [start rng .. end rng])
+        let blocksPerPeer =
+                min maxBlocksPerPeer
+                    (ceiling @Double @Int (fromIntegral (end rng - start rng) / fromIntegral (length activePeers)))
 
-      -- FIXME(adn) Deal with missing blocks.
-      (requestedBlocks, _missingBlocks) <- lift $
-          foldM (\(requested, missing) (activePeer, heightRange) -> do
-                   let mbRange = Range <$> head heightRange <*> pure (last heightRange)
-                   case mbRange of
-                     Nothing -> pure (requested, missing)
-                     Just subRange -> do
-                         res <- fetch fetcher activePeer SGetBlocks subRange
-                         case res of
-                           Left _timeout -> pure (requested, missing) -- TODO(adn) Telemetry
-                           Right blks    -> pure $ (foldl' (flip Set.insert) requested (toOldestFirst blks), missing)
-                )
-                (mempty, [])
-                workDistribution
+        -- Distribute work across peers, which is equally split, where
+        -- each peer can be given @at most@ 'maxBlocksPerPeer' at the time,
+        -- in order to limit the amount of blocks in transit each time.
+        let workDistribution =
+                zip (cycle (HS.toList activePeers))
+                    (chunksOf blocksPerPeer [start rng .. end rng])
 
-      -- FIXME(adn) streaming?
-      forM_ requestedBlocks $ \b ->
-          forM_ (scUpstreamConsumers ctx) $ \dispatch ->
-              lift $ dispatch (SyncBlock b)
+        -- FIXME(adn) streaming?
+        (requestedBlocks, !missingBlocks) <- lift $
+            foldM (\(requested, missing) (activePeer, !heightRange) -> do
+                     case Range <$> head heightRange <*> pure (last heightRange) of
+                         Nothing -> pure (requested, missing)
+                         Just subRange -> do
+                             res <- fetch fetcher activePeer SGetBlocks subRange
+                             case res of
+                               Left _timeout ->
+                                   pure (requested, missing <> heightRange) -- TODO(adn) Telemetry
+                               Right blks    -> do
+                                   -- If the peer didn't have the requested range,
+                                   -- the list of missing blocks increases.
+                                   -- FIXME(adn) not even trying to be efficient
+                                   -- in this first version.
+                                   let missingHeights = heightRange \\ map height (toOldestFirst blks)
+                                   pure $ ( foldl' (flip Set.insert) requested (toOldestFirst blks)
+                                          , missing <> missingHeights
+                                          )
+                  )
+                  (mempty, [])
+                  workDistribution
 
+        forM_ requestedBlocks $ \b ->
+            forM_ (scUpstreamConsumers ctx) $ \dispatch ->
+                lift $ dispatch (SyncBlock b)
+
+        -- Handle missing blocks
+        -- NOTE(adn) Not terribly convinced about this optimization.
+        case sort missingBlocks of
+          [] -> pure ()
+          stillMissing -> forM_ stillMissing $ \missing -> do
+              -- Tries to download each individual block from each and every
+              -- peer. We don't fail in case the block cannot be downloaded,
+              -- as we hope that eventually the peers will be able to catch up.
+              results <- forM (HS.toList activePeers) $ \peer -> do
+                           res <- lift $ fetch fetcher peer SGetBlocks (Range missing missing)
+                           case res of
+                             Left _err -> pure Nothing
+                             Right oldest -> case toOldestFirst oldest of
+                                               [b] -> pure (Just b)
+                                               _   -> pure Nothing
+              case catMaybes results of
+                []  -> pure () -- TODO(adn) telemetry?
+                b:_ ->
+                    forM_ (scUpstreamConsumers ctx) $ \dispatch ->
+                        lift $ dispatch (SyncBlock b)
--- a/src/Oscoin/Protocol/Sync.hs
+++ b/src/Oscoin/Protocol/Sync.hs
@@ -31,14 +31,20 @@ module Oscoin.Protocol.Sync
     ) where
 
 import           Oscoin.Crypto.Hash (Hash)
-import           Oscoin.Crypto.PubKey (PublicKey)
 import           Oscoin.Prelude
 import           Prelude (last)
 
 import           Oscoin.Consensus.Nakamoto (PoW)
 
 import           Oscoin.Crypto.Blockchain.Block
-                 (Block, BlockHeader, Height, Sealed, blockHeader, blockHeight)
+                 ( Beneficiary
+                 , Block
+                 , BlockHeader
+                 , Height
+                 , Sealed
+                 , blockHeader
+                 , blockHeight
+                 )
 import           Oscoin.Data.Tx
 import           Oscoin.P2P as P2P
 import           Oscoin.Storage.Block.Abstract (BlockStoreReader)
@@ -68,17 +74,17 @@ data SyncEvent c tx s =
 deriving instance ( Eq (Hash c)
                   , Eq tx
                   , Eq s
-                  , Eq (PublicKey c)
+                  , Eq (Beneficiary c)
                   ) => Eq (SyncEvent c tx s)
 deriving instance ( Ord (Hash c)
                   , Ord tx
                   , Ord s
-                  , Ord (PublicKey c)
+                  , Ord (Beneficiary c)
                   ) => Ord (SyncEvent c tx s)
 deriving instance ( Show (Hash c)
                   , Show tx
                   , Show s
-                  , Show (PublicKey c)
+                  , Show (Beneficiary c)
                   ) => Show (SyncEvent c tx s)
 
 data SyncError =
@@ -100,7 +106,7 @@ data Timeout =
     -- ^ The operation exceeded the expected timeout, in nanoseconds.
     deriving (Show, Eq)
 
--- | A closed range [lo,hi].
+-- | A closed (inclusive) range [lo,hi].
 data Range = Range
     { start :: Height
     , end   :: Height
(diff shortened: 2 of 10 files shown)


src/Oscoin/Protocol/Sync/RealWorld.hs
```diff
@@ -0,0 +1,169 @@
...
+    => Consensus.Config
+    -> BlockStoreReader c tx s IO
+    -> [SyncEvent c tx s -> IO ()]
+    -> Probe IO
+    -> GossipT c IO (SyncContext c tx s IO)
+newSyncContextIO config chainReader upstreamConsumers probe = do
+    env <- ask
+    mgr <- liftIO (HTTP.newManager HTTP.defaultManagerSettings)
+    pure $ SyncContext
+        { scNu          = Consensus.mutableChainDepth config
```
adinapoli-mndc: Oh, this is wrong and a clear leftover from a previous version of the spec. I can fix it once I finalise the implementation for the `RealWorld` implementation.

src/Oscoin/Protocol/Sync.hs
```diff
@@ -0,0 +1,393 @@
...
+                  , Show s
+                  , Show (PublicKey c)
+                  ) => Show (SyncEvent c tx s)
+
+data SyncError =
+      RequestTimeout ProtocolRequest Addr Timeout
+      -- ^ The given peer exceeded the request timeout when serving this
+      -- 'SyncProtocolRequest'.
+    | NoActivePeers
+      -- ^ There are not active peers to talk to.
```
cloudhead:
```suggestion
      -- ^ There are no active peers to talk to.
```

src/Oscoin/Protocol/Sync.hs
```diff
@@ -0,0 +1,393 @@
...
+{------------------------------------------------------------------------------
+  Types
+------------------------------------------------------------------------------}
+
+type ActivePeer c  = NodeInfo c
+type ActivePeers c = HashSet (ActivePeer c)
+
+data SyncEvent c tx s =
+      SyncBlock (Block c tx (Sealed c s))
+    | SyncBlockHeader (BlockHeader c s)
```
cloudhead: Would be good to have docs for the type and constructors

src/Oscoin/Protocol/Sync.hs
```diff
@@ -0,0 +1,393 @@
...
+    | AllPeersTimeoutError
+      -- ^ All the queried peers didn't respond in the allocated timeout.
+    deriving (Show, Eq)
+
+instance Exception SyncError
+
+data Timeout =
+    MaxTimeoutExceeded (Expected Duration)
+    -- ^ The operation exceeded the expected timeout, in nanoseconds.
+    deriving (Show, Eq)
```
cloudhead: Not clear why this isn't a `SyncError`?

src/Oscoin/Protocol/Sync.hs
```diff
@@ -0,0 +1,393 @@
...
+    , Timeout(..)
+    , ActivePeer
+    , ActivePeers
+    , SyncEvent(..)
+    , SyncContext(..)
+    , DataFetcher(..)
+    , ProtocolRequest(..)
+    , SProtocolRequest(..)
+    , ProtocolResponse
+    , Range(..)
```
cloudhead: Does this have to be exported? If so it would be good to prefix the accessors, like `rangeStart`, `rangeEnd`

cloudhead: Perhaps also naming it `HeightRange` :man_shrugging:

src/Oscoin/Protocol/Sync.hs
```diff
@@ -0,0 +1,393 @@
...
+type family ProtocolRequestArgs (c :: ProtocolRequest) :: * where
+    ProtocolRequestArgs 'GetTip          = ()
+    ProtocolRequestArgs 'GetBlocks       = Range
+    ProtocolRequestArgs 'GetBlockHeaders = Range
+
+-- | A protocol response, indexed by the corresponding 'ProtocolRequest, so
+-- that matching the wrong type will result in a type error.
+type family ProtocolResponse c tx s (r :: ProtocolRequest) :: *
+
+type instance ProtocolResponse c (Tx c) PoW 'GetTip =
```
cloudhead: Is there a reason the sync code can't be polymorphic in `tx`?

cloudhead: And the seal?

src/Oscoin/Protocol/Sync.hs
```diff
@@ -0,0 +1,393 @@
...
+      fetch :: forall r. ActivePeer c
+            -> SProtocolRequest r
+            -> ProtocolRequestArgs r
+            -> m (Either Timeout (ProtocolResponse c tx s r))
+                }
+
+-- | An opaque reference to a record of functions which can be used to query
+-- the network and the other parts of the system for information.
+data SyncContext c tx s m = SyncContext
+    { scNu               :: Nu
```
cloudhead: I'd prefer if this was just a `Natural` - we already know it's `nu` based on the name, but now the type tells me nothing because I don't know what `Nu` could be..

src/Oscoin/Protocol/Sync.hs
```diff
@@ -0,0 +1,393 @@
...
+    -- ^ A tracer for interesting events that happens during syncing.
+    , scConcurrently      :: forall a b t.  Traversable t => t a -> (a -> m b) -> m (t b)
+    -- ^ An action that can be used to fetch some resources in parallel.
+    -- 'IO'-based implementations will probably use 'forConcurrently' or some
+    -- variation of it.
+    , scLocalChainReader  :: BlockStoreReader c tx s m
+    -- ^ A read-only handle to the local blockchain.
+    , scUpstreamConsumers :: [SyncEvent c tx s -> m ()]
+    -- ^ A list of \"upstream consumers\", represented as effectful actions from
+    -- a 'SyncEvent' to 'm ()'.
```
cloudhead: Based on the type, this is more a list of "event handlers", which is much more powerful.

src/Oscoin/Protocol/Sync.hs
```diff
@@ -0,0 +1,393 @@
...
+
+-- | A sync monad over @m@, that can throw 'SyncError's.
+newtype Sync c tx s m a =
+    Sync { runSync :: ExceptT SyncError (ReaderT (SyncContext c tx s m) m) a }
+    deriving ( Functor
+             , Applicative
+             , Monad
+             , MonadError SyncError
+             , MonadReader (SyncContext c tx s m)
+             )
```
cloudhead: Should we call this `SyncT`? :man_shrugging:

src/Oscoin/Protocol/Sync.hs
```diff
@@ -0,0 +1,393 @@
...
+------------------------------------------------------------------------------}
+
+height :: Block c tx s -> Height
+height = blockHeight . blockHeader
+
+-- | The nu parameter, from the spec.
+type Nu = Natural
+
+maxBlocksPerPeer :: Int
+maxBlocksPerPeer = 50
```
cloudhead: Docs would be good - I guess it's the max number of blocks to request from a peer?

src/Oscoin/Protocol/Sync.hs
```diff
@@ -0,0 +1,393 @@
...
+       , Eq s
+       , Eq (Hash c)
+       , Eq (PublicKey c)
+       )
+    => Nu
+    -> Block c tx s
+    -> Block c tx s
+    -> Bool
+isDone (fromIntegral -> nu) localTip remoteTip =
+    remoteTip == localTip || height remoteTip - height localTip < nu
```
cloudhead: Might be a good idea to handle the case where `localTip` is ahead of `remoteTip`, or mark an invariant. Even if `Height` is currently an `Integer` and can go in the negative, if that ever changes, this could silently break.

src/Oscoin/Protocol/Sync.hs
```diff
@@ -0,0 +1,393 @@
...
+-- | Returns the range between two blocks.
+range
+    :: Block c tx s
+    -> Block c tx s
+    -> Maybe Range
+range localTip remoteTip
+  | outOfRange = Nothing
+  | otherwise  = Just (Range start end)
+  where
+    start      = succ (height localTip)
```
cloudhead: Hmmm, I thought `Range` had the starting height be inclusive?

src/Oscoin/Protocol/Sync.hs
```diff
@@ -0,0 +1,393 @@
...
+    -> Bool
+isDone (fromIntegral -> nu) localTip remoteTip =
+    remoteTip == localTip || height remoteTip - height localTip < nu
+
+-- | Returns the range between two blocks.
+range
+    :: Block c tx s
+    -> Block c tx s
+    -> Maybe Range
+range localTip remoteTip
```
cloudhead: This is baking in the assumption that the localTip is always behind the remoteTip

src/Oscoin/Protocol/Sync.hs
```diff
@@ -0,0 +1,393 @@
...
+      -- ^ There are not active peers to talk to.
+    | NoRemoteTipFound
+      -- ^ It was not possible to fetch a valid tip from our peers.
+    | AllPeersTimeoutError
+      -- ^ All the queried peers didn't respond in the allocated timeout.
+    deriving (Show, Eq)
+
+instance Exception SyncError
+
+data Timeout =
```
kim: Are timeouts ever considered? If not, then why have a "checked" error at all?

src/Oscoin/Protocol/Sync.hs
```diff
@@ -0,0 +1,393 @@
...
+    start      = succ (height localTip)
+    end        = height remoteTip
+    outOfRange = end < start
+
+{------------------------------------------------------------------------------
+  Monadic operations
+------------------------------------------------------------------------------}
+
+-- | The 'remoteTip' operation from the spec. It tries to fetch the tips
+-- concurrently, picking the tip returned by /most/ of the peers.
```
cloudhead: If I understand, there's the possibility that each peer returns a different tip. If that's the case, what happens?

src/Oscoin/Protocol/Sync.hs
```diff
@@ -0,0 +1,393 @@
...
+                  (mempty, [])
+                  workDistribution
+
+        forM_ requestedBlocks $ \b ->
+            forM_ (scUpstreamConsumers ctx) $ \dispatch ->
+                lift $ dispatch (SyncBlock b)
+
+        -- Handle missing blocks
+        -- NOTE(adn) Not terribly convinced about this optimization.
+        case sort missingBlocks of
```
kim: Move to top level?

src/Oscoin/Protocol/Sync/RealWorld.hs
```diff
@@ -0,0 +1,169 @@
...
+    :: forall c tx s.
+    ( Serialise (ProtocolResponse c tx s 'GetTip) )
+    => HTTP.Manager
+    -> DataFetcher c tx s IO
+newDataFetcherIO httpManager = DataFetcher $ \peer -> \case
+    -- This implementation is basically 'fetchTip' function from the spec.
+    SGetTip -> \() -> timed maxAllowedNetworkLatency $ do
+        let hn = addrHost . nodeHttpApiAddr $ peer
+        -- FIXME(adn) Support for https if needed.
+        rq0 <- HTTP.parseUrlThrow ("http://" <> toS (renderHost hn) <> "/blockchain/tip")
```
cloudhead: We should make sure this only returns the header, since it can otherwise result in several megabites of download just to get the best height.

src/Oscoin/Protocol/Sync.hs
```diff
@@ -0,0 +1,393 @@
...
+    ctx <- ask
+    let fetcher = scDataFetcher ctx
+    withActivePeers $ \ activePeers -> do
+
+        let blocksPerPeer =
+                min maxBlocksPerPeer
+                    (ceiling @Double @Int (fromIntegral (end rng - start rng) / fromIntegral (length activePeers)))
+
+        -- Distribute work across peers, which is equally split, where
+        -- each peer can be given @at most@ 'maxBlocksPerPeer' at the time,
```
cloudhead:
```suggestion
        -- each peer can be given /at most/ 'maxBlocksPerPeer' at the time,
```

src/Oscoin/Protocol/Sync.hs
adinapoli-mndc: Historical reasons: originally I wanted to have `timed` be a standalone (read: reusable) combinator (perhaps to add as part of our Prelude) but eventually (read: in the latest revision of the code) it actually returns a `SyncError` (I think?), so we might just as well decide to remove `Timeout` and tuck it inside the `SyncError`.

src/Oscoin/Protocol/Sync.hs
adinapoli-mndc: It is useful to have it exported for testing, but sure enough we can prefix the accessors.

src/Oscoin/Protocol/Sync.hs
adinapoli-mndc: I was originally afraid that if I wasn't monomorphic in the `tx` and `s`, the compiler wouldn't have been able to specialise the typefam and actually apply. However, in the last version of my code `tx` is completely generic, so I might perhaps take another stab with `PoW` and see how far I get.

src/Oscoin/Protocol/Sync.hs
adinapoli-mndc: Agreed.

src/Oscoin/Protocol/Sync.hs
adinapoli-mndc: Yes, it loosely corresponds to the "in-flight" cap we have during inventory exchange in BitCoin. Will document 👍

src/Oscoin/Protocol/Sync.hs
adinapoli-mndc: I think you're right -- not just for the silent breakage, but also because if we are ahead, it might mean there was a fork, so we _do want_ to sync again (i.e. `isDone` should yield `False` in that case).

src/Oscoin/Protocol/Sync.hs
adinapoli-mndc: It does, but in order to compute it, you pass this function the current tip you are in, so that the range can request _the next block_.

src/Oscoin/Protocol/Sync.hs
adinapoli-mndc: Good point, I will ponder on this a little bit more.

src/Oscoin/Protocol/Sync.hs
adinapoli-mndc: We pick the longest, but I guess even "pick randomly" could have worked:

```
      comparingTips (b1,f1) (b2,f2) =
          case f2 `compare` f1 of
            EQ -> height b2 `compare` height b1
            x  -> x
```

The advantage of picking randomly is that you are (allegedly) less subject to attacks where a dishonest player would trick you into following his chain by forging a tip with a very high .. height (sorry for the repetition 😉 )

src/Oscoin/Prelude.hs
```diff
@@ -461,3 +463,8 @@ notImplemented = withFrozenCallStack (error "Not implemented")
 {-# WARNING undefined "'undefined' remains in code" #-}
 undefined :: (HasCallStack) => a
 undefined = withFrozenCallStack (error "Prelude.undefined")
+
+-- Simple type aliases that makes easier for the reader to distinguish, in
+-- a type signature, which is the expected vs actual value.
+type Expected a = a
```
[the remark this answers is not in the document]
adinapoli-mndc: Yeah, fair point. For now I will revert this change and open another PR (if necessary). Still not convinced these type aliases are worth the inclusion into the Prelude :)

src/Oscoin/Protocol/Sync.hs
```diff
@@ -0,0 +1,393 @@
...
+    => Range
+    -> Sync c tx s m ()
+syncBlocks rng = do
+    ctx <- ask
+    let fetcher = scDataFetcher ctx
+    withActivePeers $ \ activePeers -> do
+
+        let blocksPerPeer =
+                min maxBlocksPerPeer
+                    (ceiling @Double @Int (fromIntegral (end rng - start rng) / fromIntegral (length activePeers)))
```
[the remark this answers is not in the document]
adinapoli-mndc: sorry about that, did my best to keep everything under 80 cols. The good news is that the next PR moves this into a `where` clause!

src/Oscoin/Protocol/Sync.hs
```diff
@@ -0,0 +1,393 @@
...
+
+        -- Distribute work across peers, which is equally split, where
+        -- each peer can be given @at most@ 'maxBlocksPerPeer' at the time,
+        -- in order to limit the amount of blocks in transit each time.
+        let workDistribution =
```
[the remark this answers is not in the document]
adinapoli-mndc: Yup, I think this is already in the second PR.

src/Oscoin/Protocol/Sync.hs
```diff
@@ -0,0 +1,393 @@
...
+                zip (cycle (HS.toList activePeers))
+                    (chunksOf blocksPerPeer [start rng .. end rng])
+
+        -- FIXME(adn) streaming?
+        (requestedBlocks, !missingBlocks) <- lift $
+            foldM (\(requested, missing) (activePeer, !heightRange) ->
```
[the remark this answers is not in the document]
adinapoli-mndc: Yup, PR number 2 should feel much more linear.

src/Oscoin/Protocol/Sync.hs
adinapoli-mndc: I originally was thinking to have `timed` being a standalone combinator, but looks like we might get away by simply merging `Timeout` into `RequestTimeout`.

src/Oscoin/Protocol/Sync/RealWorld.hs
adinapoli-mndc: At the moment we can't , as we don't have an equivalent `getBlockHeadersByHeight`, and I am OK for now to keep the dumb version up until we don't need to start supporting light clients.

See also https://github.com/oscoin/oscoin/issues/550#issuecomment-494331328 (TL;DR; Thomas and I accepted the compromise to delay this a little longer).

adinapoli-mndc: (Deleted my previous comment as it wasn't very meaningful). At the moment we can't, as we don't have a `getTipHeader` HTTP API. I feel the case of "give me the header" vs "give me the full block" is common enough that perhaps we should be thinking having our HTTP API accept a parameter (for each request) where you can specify which "view" you want over a block (i.e. either `light` or `full`). This way at least we wouldn't duplicate all the endpoints.

PR #573, opened 2019-05-21
Experimental streaming support for node syncing

adinapoli-mndc:
This is an incomplete, experimental support for streaming blocks during syncing. It relies on `conduit` to stream blocks from the `DataFetcher` down to the upstream consumers (cfr. `syncBlocks`). At the moment the elephant in the room is not addressed, i.e. how to efficiently write a `Sink` capable of streaming the HTTP API results. That will come as an additional commit and should be possible by leveraging `cbor` 's incremental decoders. (Actually, I think I can use either `Data.Conduit.Serialise` from our own project or some minor variation on it).


src/Oscoin/Protocol/Sync/RealWorld.hs
```diff
@@ -1,84 +1,136 @@
...
+           -- Actual type: IO (ProtocolResponse 'GetBlockHeaders)
+
+            liftIO $ HTTP.withResponse rq httpManager $ \res -> do
+                cborBlob <- HTTP.brRead (HTTP.responseBody res)
+                pure $ CBOR.deserialise . toS $ cborBlob
+
+        -- This implementation is basically 'fetchBlocks' from the spec.
+        -- N.b. It assumes oscoin#550 is implemented.
+        SGetBlocks -> \Range{..} -> timed maxAllowedNetworkLatency $ do
+            let url = stext
```
kim: Do we really want to operate with strings here, and parse every time?

adinapoli-mndc: Uhm, probably not -- perhaps we should pass a function from `Peer -> Request` ? We wouldn't know which port/ip we want to call (that's intrinsic of the `Peer`) but surely the `Request` can be computed beforehand, so that we can avoid the parsing every time.

PR #574, opened 2019-05-22
Expose lookupBlocksByHeight and lookupBlockByHeight (Blockstore & HTTP API)

### LINKED ISSUE #550: Add lookupHeight and lookupHeightRange to both the Blockstore reader and HTTP APIs

adinapoli-mndc:
Fixes #550. This PR adds two new functions, `lookupBlocksByHeight` and `lookupBlockByHeight` to our various block stores & exposes them as part of the HTTP API. 

It's worth pointing out that `lookupBlocksByHeight` has a slightly different semantic from the one described in the spec, as it takes the range _inclusively_. I have done this for two reasons:

* The syncing code needs to work with inclusive ranges, so it's pretty much for conveniency reasons;
* We can recuperate the original spec semantic if we want by having a `lookupBlocksByHeight'` that simply asks for `(succ start, end)`.

Let me know what you think!


src/Oscoin/Storage/Block/Abstract.hs
```diff
@@ -55,6 +55,10 @@ data BlockStoreReader c tx s m = BlockStoreReader
     -- ^ Get the genesis block.
     , lookupBlock     :: BlockHash c -> m (Maybe (Block c tx (Sealed c s)))
     -- ^ Lookups a 'Block' from the store when given its hash.
+    , lookupBlockByHeight :: Height -> m (Maybe (Block c tx (Sealed c s)))
+    -- ^ Lookup a block, given its 'Height'.
+    , lookupBlocksByHeight :: (Height,Height) -> m (OldestFirst [] (Block c tx (Sealed c s)))
+    -- ^ Returns the blocks (if any) between the height range.
```
geigerzaehler: I think it would be useful to document whether the range is inclusive or exclusive and that we might get fewer blocks than we expect (if the upper bound is too big).

src/Oscoin/API/HTTP/Handlers.hs
```diff
@@ -97,6 +97,27 @@ getBlock h = do
...
+        Nothing ->
+            respond notFound404 $ errBody "Block not found"
+
+lookupBlocksByHeight
+    :: (Serialise (SealedBlock c tx s))
+    => ApiAction c tx s i a
+lookupBlocksByHeight = do
+    start <- param' "start"
+    end   <- param' "end"
+    blocks <- liftNode $ Node.lookupBlocksByHeight (start, end)
```
geigerzaehler: I think we might want to unwrap `OldestFirst` and just send a list over the wire. Because the encoding of `OldestFirst`, `NewestFirst` and the wrapped list are the same we don’t get anything out of this. We would also not need the serialise instance for the chrono constructors.

src/Oscoin/Storage/Block/Memory.hs
```diff
@@ -57,6 +57,13 @@ mkBlockStore getBlocks modifyBlocks = (blockStoreReader, blockStoreWriter)
                 _ Seq.:|> gen  -> pure gen
         , lookupBlock = \h ->
             find (\b -> blockHash b == h) <$> getBlocks
+        , lookupBlockByHeight = \h ->
+            find (\b -> (blockHeight . blockHeader $ b) == h) <$> getBlocks
```
geigerzaehler: For a blockchain to be valid its blocks must have height equaling their position in the chain. If we are sure that `Block.Memory` satisfies this invariant we could implement this by just requesting the block in the list at the given height. If not we should add a comment so that people can understand why we didn’t chose the simple version.

adinapoli-mndc: I agree, but I think that here the `Seq` doesn't give us such guarantee (if `getBlocks` were using a `BlockChain c tx s` type, that would have been easier to enforce). As the memory implementation was always meant as a simplified store where performance is not the primary concern, I would be inclined to go with the simpler version. Will add a comment.

src/Oscoin/API/HTTP/Handlers.hs
adinapoli-mndc: d146be09

src/Oscoin/Storage/Block/Abstract.hs
adinapoli-mndc: d146be09

src/Oscoin/Storage/Block/Memory.hs
adinapoli-mndc: d146be09

PR #575, opened 2019-05-22
Dict API for Client

geigerzaehler:
We replace the MTL-style client interface with a dictionary based client interface. We also eliminate the test client for the CLI tests because it is unused. This will make it easier to make changes to and extend the client.

Preparation for #576.


src/Oscoin/API/Client/HTTP.hs
```diff
@@ -51,45 +57,27 @@ data Response = Response
...
+-- requests.
+makeHttpClient
+    :: ( Buildable (Hash c)
+       , Serialise (BlockHash c)
+       , Serialise (PublicKey c)
+       , Serialise (Signature c)
+       , Monad m
+       )
+    => (Request -> m Response) -> Client c m
+makeHttpClient httpRequest = Client{..}
```
cloudhead: Perhaps we can call this `httpClientFromRequest`, to differentiate it from the other function, as they sound like they do the same thing currently.

test/Oscoin/Test/API/HTTP/TestClient.hs
```diff
@@ -1,34 +1,32 @@
--- | Provide an instance of 'MonadClient' that can be run in a WAI Test
--- 'Session'.
+-- | Provide a 'Client' object for in a WAI Test 'Session'.
```
cloudhead:
```suggestion
-- | Provide a 'Client' object for a WAI Test 'Session'.
```

src/Oscoin/API/Client/HTTP.hs
geigerzaehler: Done

Issue #549, opened 2019-05-13
Devnet Ledger API


geigerzaehler:
## Problem

The full node needs an API that the CLI in #548 can use.

## Requirements

The HTTP API provides a `submitTx` and `lookupState` endpoint.

#### `submitTx`

A Full Node may accept or reject transactions.

If a transaction is accepted, the node will gossip it to the network and try to include it in blocks it produces. Conversely, transactions that are rejected by the API will not be gossiped.

A transaction is accepted if it is correctly signed by the author contained in the transaction.

#### `lookupState`

Given a state key returns the value in the state of the tip of the best chain. Returns nothing if value is not in the state.

PR #576, opened 2019-05-22
Prepare Client.getState for OscoinTx

geigerzaehler:
As part of implementing the Ledger API for the devnet (#549) we prepare the `Client.getState`. We change the interface of `Client.getState` so that it matches the expected interface defined in #549 while retaining the legacy tx under the hood.

* Eliminate `getPath` as it is unused
* Rename `getPathLatest` to `getStateValue`
* Eliminate the `QueryVal` associated type. State values are (for now) only `ByteString`
* Change the `key` argument type from `[Text]` to `ByteString`. This requires us to encode the keys with base58 when sending it over HTTP.

### DIFF
--- a/src/Oscoin/API/HTTP/Handlers.hs
+++ b/src/Oscoin/API/HTTP/Handlers.hs
@@ -2,7 +2,6 @@ module Oscoin.API.HTTP.Handlers where
 
 import           Oscoin.Prelude
 
-import           Lens.Micro.Mtl (view)
 import           Oscoin.API.HTTP.Internal
 import           Oscoin.API.Types
 import           Oscoin.Crypto.Blockchain (Height, TxLookup(..))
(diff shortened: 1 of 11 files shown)


src/Oscoin/API/HTTP/Handlers.hs
```diff
@@ -119,15 +120,16 @@ lookupBlocksByHeight = do
...
-getStatePath _chain = do
-    path <- listParam "q"
-    result' <- liftNode $ Node.getPathLatest path
-    case result' of
-        Just val ->
-            respond ok200 (Ok val)
-        Nothing ->
-            respond notFound404 $ errBody "Value not found"
+getStateValue encodedKey =
+    case decodeBase58btc (toS encodedKey) of
```
cloudhead: This is a strange API to me... The previous one had the benefit of being human readable. Can't we keep the old API and just encodeUtf8 it to convert it to a bytestring?

cloudhead: If we must use some kind of encoding, then I would opt for a simple hex/base16 encoding - there's no reason to trade space for implementation complexity here.

geigerzaehler: The API needs to be deal with arbitrary bytestrings. The OscoinTx WorldState uses ByteString as its key and this ByteString is most likely not valid `utf8`. Spock does not give us a way to extract ByteString values from parameters so we need to use our own encoding. I will switch it to hex, though.

PR #577, opened 2019-05-22
Hash Account IDs

cloudhead:
* Update AccountId: `type AccountId c = ShortHash c`
* Update `Address` to contain the new `AccountId`.
  * This mainly reverts to the original `Address` format which had `ripemd160(hash(pk))`.
* Make `txAuthor` a `PublicKey`
  * Required for signature verification.
* Remove unsafe code in `defaultBeneficiary`
  * We use a burn address now (all zeros)
* Beneficiary is now a hash (`AccountId`), and therefore `--beneficiary` flag is updated
  * Now works with `--beneficiary 0x0000000000000000000000000000000000000000` (40 zeros)


src/Oscoin/Crypto/Hash/RealWorld.hs
```diff
@@ -150,6 +170,9 @@ instance Show.Show (Hash Crypto) where
                   . convert
                   $ d
 
+instance Show.Show (ShortHash Crypto) where
+    show = T.unpack . F.sformat ("0x" F.% formatShortHash)
```
adinapoli-mndc: Apparently there is a `hex` formatter inside `Formatting.Formatters` that should allegedly do the same thing.

src/Oscoin/Crypto/Hash/RealWorld.hs
```diff
@@ -111,6 +121,16 @@ instance FromJSON (Hash Crypto) where
...
             . second Hash . decodeAtBase BaseN.Base58btc . encodeUtf8
 
+instance ToJSON (ShortHash Crypto) where
+    toJSON     = toJSON . T.pack . show
+    toEncoding = toEncoding . T.pack . show
+
+instance FromJSON (ShortHash Crypto) where
+    parseJSON = withText "ShortHash" $ \t ->
+        case parseShortHash t of
+            Nothing -> fail "invalid short hash"
```
adinapoli-mndc: Shouldn't we fail by also showing the input text, or do you feel that's redundant? I can imagine how this might be useful if this failure is triggered deep down the bowels of the app, and gets logged without any context.

test/Main.hs
```diff
@@ -79,7 +80,7 @@ main = do
                 , Test.Oscoin.Configuration.tests
                 , Test.Oscoin.Crypto.Blockchain.Eval.tests crypto
                 , Test.Oscoin.Crypto.Blockchain.GeneratorsTest.tests crypto
-                , Test.Oscoin.Crypto.Hash.tests crypto
+                , Test.Oscoin.Crypto.Hash.tests realCrypto
```
adinapoli-mndc: Why we need this? If you want to exercise the real world crypto imho the correct solution is to leave the crypto generic but to pass `--crypto=realWorld` to your test arguments (see https://github.com/oscoin/oscoin/pull/511). You want this to be run all the time? Then https://github.com/oscoin/oscoin/issues/502 is the answer.

Passing the real crypto here will create a singularity in the tests and we will soon be start adding all this special case around, at which point this would defeat the purpose of having introduced a mock crypto in the first place 😉

test/Test/Oscoin/Crypto/Hash.hs
```diff
@@ -16,18 +17,39 @@ import           Test.Tasty.QuickCheck hiding ((===))
...
+prop_textRoundtripShortHash Dict h =
+    (Crypto.parseShortHash . show) h === Just h
+
+prop_serialiseRoundtripHash :: forall c. Dict (IsCrypto c) -> Crypto.Hash c -> Property
+prop_serialiseRoundtripHash Dict h =
+    (CBOR.deserialise . CBOR.serialise $ h) === h
+
+prop_serialiseRoundtripShortHash :: forall c. Dict (IsCrypto c) -> Crypto.ShortHash c -> Property
+prop_serialiseRoundtripShortHash Dict h =
+    (CBOR.deserialise . CBOR.serialise $ h) === h
```
adinapoli-mndc: Good lad for adding roundtrips 😁 👏

src/Oscoin/Crypto/Hash.hs
```diff
@@ -108,10 +112,16 @@ class
...
 
+    -- | The zero short hash. Also the minimum short hash value.
+    zeroShortHash  :: ShortHash crypto
+
     -- | A short hash, in the similar spirit of the BTC one.
     -- eg. ripemd160(blake2b(x))
     toShortHash :: Hash crypto -> ShortHash crypto
 
+    -- | Parse a 'ShortHash' from 'Text'.
+    parseShortHash :: Text -> Maybe (ShortHash crypto)
```
adinapoli-mndc: It's slightly strange that we have this `parseShortHash` but not the equivalent for the "normal" one. I do agree though that this interface grew up a little bit out of control. For now let's stick with what you have done here and perhaps in the next iteration we will think at how we can make the API a bit more consistent.

src/Oscoin/Crypto/Address/Internal.hs
```diff
@@ -80,14 +80,14 @@ data AddressFormatTag =
     AddressFormatTag_CBOR
     deriving (Eq, Ord, Show)
 
--- | The 'AddressPayload' is (currently) simply a 'PublicKey'.
+-- | The 'AddressPayload' is a 'ShortHash' of a 'PublicKey'.
 --
 data AddressPayload c where
-    AddressPayload_V0 :: PublicKey c -> AddressPayload c
+    AddressPayload_V0 :: ShortHash c -> AddressPayload c
```
adinapoli-mndc: In my very first iteration of the address format, I wrote [this](https://github.com/oscoin/oscoin/pull/431/files#diff-0a67e40a5b386d32bba504d153d37813R92). How did you avoid this problem? I guess you did use the `parseXXX` machinery to do that, but I do wonder if it might be easier simply to have a `ByteArray` there, if that clean up things. Not proposing to actually go with this design, but just flagging up that might be a route worth exploring 😉

test/Main.hs
cloudhead: The thinking was that testing the mock hashing functions is not that useful vs. testing the real hashing functions. Ideally we would test both mock and realworld here. #502 is not really the issue here, because the tests are actually fast to run, and if you're working with the hashing, you actually want to test the realworld crypto.

src/Oscoin/Crypto/Hash.hs
cloudhead: Yeah - I didn't want to add `parseHash` if we weren't actually using it, but I agree that we should look at this at some point (and actually the whole module).

src/Oscoin/Crypto/Address/Internal.hs
cloudhead: I also wondered - I don't quite understand what the stated problem was to be honest, due to my limited haskell experience :confused: 

Could it be that I used `ShortHash` instead of `ShortHashed`?

test/Main.hs
adinapoli-mndc: Yeah, I know where you are coming from -- I am just worried that if we start going down this route there is no coming back, we will keep adding a galore of this "special cases" as now it's the hash, in the future it might be anything else crypto-dependent. If we can, I would prefer to keep (one and only) dictionary at the beginning of the test run and use the `--crypto` CLI or, if you are strongly opposed to it, at least a comment describing while are we (deliberately) doing that 😉

src/Oscoin/Crypto/Address/Internal.hs
adinapoli-mndc: (Going by memory here): Back in the days I think the problem was that I wanted to use a `ShortHashed` to be able to "remember" which was the thing being hashed (perhaps I thought it was useful for migration purposes, who knows). But that wasn't possible (for the reasons I have explained in the comment, essentially there was no inverse function during decoding to go from a `ShortHashed` back to the `PublicKey`) so the only way to encode the fact that Hash was actually hashing a `PublicKey` was using `Tagged`. Perhaps it's useful to do the same here?

src/Oscoin/Crypto/Hash/RealWorld.hs
cloudhead: Looks like it's not what we want (expects an integral) :( - I moved the prefix to `formatShortHash` to clean things up instead.

src/Oscoin/Crypto/Hash/RealWorld.hs
cloudhead: 6949e75

test/Main.hs
cloudhead: Yeah, I see the concern - but I think for this suite to be useful, we have to have some kind of compromise, because in this case, what we're testing is actually the `crypto` itself, not something that uses the crypto. Another example would be the tests for the `PubKey` module. So basically what I'm proposing is that for those special cases, we "override" the argument, and test *both* dictionaries. I'll commit something and let me know what you think.

adinapoli-mndc: I think I can live with such compromise, provided we document those exceptions properly 😉

cloudhead: Here it is: c2f04cf

src/Oscoin/Crypto/Hash/RealWorld.hs
adinapoli-mndc: Looks good 😉

src/Oscoin/Crypto/Address/Internal.hs
cloudhead: Right, that's what I was thinking. I decided against using `Hashed` for exactly this reason - since we don't always have the original `PublicKey`, and would have to "cheat" it. I think this is the case almost always, and so I think `Hashed` is actually not that useful and should potentially be removed.

adinapoli-mndc:
The `Tagged` bit of the `AddressPayload` is still in the flux, but I think it's more important to get this merged that to antagonise over the details :)

Issue #568, opened 2019-05-21
Run node with OscoinTx


geigerzaehler:
## Problem

At the moment the node runs with `Oscoin.Data.Tx.Tx` as the transaction type. For the devnet we need to run it with `Oscoin.Data.OscoinTx.Tx` as the transaction type.

## Solution/implementation plan

We need to create issues and spec out the following

- [x] Implement Sqlite persistence for OscoinTx (#590)
- [x] Provide a `Serialise` instance for OscoinTx (#590)
- [x] CLI can generate valid genesis block for OscionTx (#589)
- [x] Replace `Evaluator` that can only evaluate txs with block data evaluator. The oscoin ledger exposes `applyTxMessages` which applies all txs in a block in bulk and needs to know about the beneficiary. (#591)

There is probably something missing here.

cloudhead:
Closed by #607

PR #579, opened 2019-05-22
cli: Prepare genesis creation for OscoinTx

geigerzaehler:
To run the node with OscoinTx (#568) we need to create a genesis block for the Oscoin ledger (#589). We prepare for this by removing the inclusion of transactions in the genesis block. While we might want to have that feature later the current implementation will not work with the Oscoin ledger. The reason is that currently transactions are signed with the node key, but we need to sign it with a fixed account owner key.


geigerzaehler:
> Based on previous conversation, if I understand correctly this is an interim piece of work to eventually be able to _do_ include transactions in the genesis block (unless you guys concluded to not go there, eventually).

That is correct. You can find the details in the #589

Issue #572, opened 2019-05-21
Log output on cloud instances


kim:
Logs aggregated by GCP's Stackdriver look like this:

```
2019-05-21 16:51:15.899 CEST
[96minfo [0m | mined block block_hash=2DrjgbKpP1aNbLnz1kQzBubeuSekWGTrffHTT88SCmmCrezMr8 ns="node" loc="Oscoin.Node:121:22"
 
Expand all | Collapse all {
 insertId:  "1qpq9l3g2ir1tuh"  
 jsonPayload: {
  container_id:  "c4ae3397b1a9abfca264c539dcb737a5aa1851a39bf5f4207029f93aa84fbce6"   
  message:  "[96minfo [0m | mined block block_hash=2DrjgbKpP1aNbLnz1kQzBubeuSekWGTrffHTT88SCmmCrezMr8 ns="node" loc="Oscoin.Node:121:22"
"   
  stream:  "stderr"   
 }
 labels: {
  compute.googleapis.com/resource_name:  "0b79a73e1518"   
 }
 logName:  "projects/opensourcecoin/logs/cos_containers"  
 receiveTimestamp:  "2019-05-21T14:51:18.955495345Z"  
 resource: {
  labels: {
   instance_id:  "1629246337835447054"    
   project_id:  "opensourcecoin"    
   zone:  "europe-west1-b"    
  }
  type:  "gce_instance"   
 }
 timestamp:  "2019-05-21T14:51:15.899358941Z"  
}
```

This is obviously not ideal. While it is possible to tweak the log aggregation configuration until it yields something more useful, this would be highly specific to the deployment platform, and thus anyone trying to deploy to their environment would have to put in this effort, too. We should instead try to make the log output more suitable to server deployments. Improvements include:

* [ ] Allow turning off colored output, even when running in development mode
* [ ] Extend the namespacing functionality to add more tags regardless of logging context. This should probably include: bound IP address, node id

adinapoli-mndc:
Definitely not opposed to kill the colored output -- it was a bit of a gimmick to begin with 😁 . I think that apart from the colours what is also bothersome for logging aggregation is that we format in a slightly different format (in a similar fashion of what this [post](https://brandur.org/logfmt) describes, the "human" section).

This makes me think wether or not it would actually be useful to separate the "local development" environment from the "clown development environment", as I can see how the human readable format might be slightly more useful when staring at logs on our own machines, but un-optimal for cloud deployments.

kim:
[...]

Issue #518, opened 2019-04-24
Proposal: add unit of measures to Duration


adinapoli-mndc:
## Problem

In the current codebase, our `Duration` type is a flat `Int64` which default resolution is nanoseconds. This leads to the situation where engineers have to add [comments](https://github.com/oscoin/oscoin/blob/master/src/Oscoin/P2P/Disco/MDns.hsc#L295) to specify which is the resolution and use [conversion functions](https://github.com/oscoin/oscoin/blob/master/src/Oscoin/P2P/Disco/MDns.hsc#L302) appropriately to get the resolution they want (see also [here](https://github.com/oscoin/oscoin/blob/c75ef71ab2f24c6622d1af77b08bbce7a4918be8/src/Oscoin/Telemetry/Metrics.hs#L251) for example.  

## Solution/implementation plan

It might be nice to experiment in decorating the `Duration` type like this (for example):

```haskell
data MeasuredIn =
    Nanoseconds
  | Microseconds
  | Milliseconds
  | Seconds

newtype Duration (m :: MeasuredIn) = Duration Int64
```

And then having a bunch of functions to convert between these representations. We could still use the current machinery we have (the zoo of `seconds`, `nanoseconds` etc functions we have) but the obvious advantage is that now `Duration 'Seconds` will be a different type than `Duration 'Milliseconds`, which will force us to reason about the code.

## (Optional) Alternative solutions considered

The alternative is to stick with our current status quo if we think this proposal will "bulkify" too much our codebase, making the task of working with duration & time quantities heavier.

cc @cloudhead @kim @geigerzaehler

geigerzaehler:
I’d be interested in how other people tackle this issue.

An alternative solution would be to not expose the `Duration` constructor and not provide a `Num` instance (to avoid `fromInteger`). Instead we would have `durationInMilliseconds :: Duration -> Int64` and `durationFromMilliseconds :: Int64 -> Duration`, etc. functions. This would avoid making types with different units incompatible.

kim:
+1

The whole point of this proposal is to prevent passing incompatible units. One would, however, still need to convert all over the place. An alternative could be to also provide shims for some commonly used things which expect a duration (of some type), e.g. `threadDelay`, which would accept a polymorphic `Duration a`, and perform the conversion under the hood. A simple `newtype` wouldn't suffice for that, though.

PR #580, opened 2019-05-22
Continuous Deployment (aka "devnet")

kim:
# What

Every build will execute a validation of the deployment configuration for a oscoin full node. The buid for the `master` branch will deploy that configuration to Google Compute Engine (GCE).

# How

A machine configuration consists of the CLI options passed to the `oscoin` executable, a systemd unit, and a [cloud-init](https://cloud-init.io) configuration which installs the systemd unit, and does other things which are specific to the cloud provider and VM image used. Currently the only supported configuration is to run `oscoin` as a Docker container on GCE using Google's [Container Optimized OS](https://cloud.google.com/container-optimized-os/docs) (COS).

The cloud (GCE) configuration is kept as minimalistic as possible: a managed instance group is created, for which the instance template (versioned by the commit SHA1) is updated on every deploy. The instance group is then instructed to perform a rolling upgrade, which spins up the new nodes before demoting the old ones. Once that succeeded, the DNS SRV records are updated to point to the new nodes. This way, the new nodes will be able to join the network through the old nodes (if there are no old nodes, #562 applies, ie. nodes will be isolated until the SRV records are set).

#  Known Limitations

* Currently, no stable storage is attached to VMs. This means that after a deployment, new nodes will have to perform a full sync with the old ones. Since we do not currently have a way to determine if a node has finished syncing, the rolling update just waits two minutes until it considers a new instance up.

  Interestingly, stable storage would just reduce the time it takes to catch up, but not eliminate the need for syncing. Thus, we should introduce a health check endpoint which indicates syncing status.

* No attempt is made at (secure) management / provisioning of key material, instead `--allow-ephemeral-keys` is used, which means no node identity will survive a deployment nor an instance restart.

* Log output is rather ugly and useless (cf. #572). It is also not written to the instance's local `journal`, which is somewhat inconvenient.

* There is no documentation (yet)

# Design Decisions

The basic rationale is that the executable's configuration is (in part) specific to the environment - yet the only way to ensure that the configuration is actually valid from the executable's point of view is to run it through it's parsing / validation routines. However, different deployment environments require to _adjust_ certain settings according to it's quirks, which in turn requires _some_ confidence that the result is stil a valid executable configuration.

Based on this, the choice was made that it is the responsibility of the application (Haskell) code to produce valid _machine_-level configuration. The environment-specific deployment code calls an executable to produce this configuration (and may specify additional executable options), and, if it succeeds, uses the resulting artifacts for any further deployment steps. Provisioning of any further infrastructure is up to the deployment code.

# Implementation Choices

* Docker
  There is actually no reason for us to use Docker containers (and it is, in fact, more of a hassle than a benefit), except that this is the only "installable" artifact we currently produce.
* COS
  COS is Google's interpretation of CoreOS, and has a very similar feature set (including automatic updates), tailored to running Docker containers. It seemed like a natural choice, as the future of CoreOS is unknown.
* cloud-init
  While its capabilities are somewhat limited, we also don't (and shouldn't!) need anything fancy which would justify are more heavyweight provisioning tool. [CoreOS's Ignition](https://github.com/coreos/ignition) was conceived as a successor to cloud-init, but there is again uncertainty about it's future, whereas cloud-init remains widely supported.

## Why not Kubernetes?

The promise of Kubernetes is to abstract over all the details of machine provisioning this PR actually deals with. That's nice, but the reality is that it is non-trivial to operate a production Kubernetes cluster in a secure way, worse if all the services are actually internet-facing.

That being said, it should be trivial to extend the code to produce a (sample) Kubernetes manifest people can use if they want to run `oscoin` nodes on K8s.

## Why not terraform?

Simple: `gcloud` is enough for our purposes. Also, terraform is not a deployment tool, but a cloud provisioning system - as such, it requires much broader privileges than absolutely necessary to deploy a single application, which makes it harder to run it securely as part of a CI/CD pipeline. It's configuration "language" also doesn't provide for much modularity and reuse, and the only way to test a typical terraform configuration is by running `terraform plan`, and eyeballing the output.

That being said, it should be trivial to integrate the devised separation of concerns into an existing terraform configuration - one can `local-exec` `oscoin-deploy`, and use the resulting `cloud-config` to provision instances.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -125,7 +125,7 @@ steps:
     - STACK_ROOT=/workspace/.stack # weeder uses stack under the hood
     args: ["stack", "exec", "--", "weeder", "."]
 
-  - id: "Build images"
+  - id: "Publish images"
     waitFor:
     - Lint (weeder)
     name: gcr.io/cloud-builders/docker
@@ -144,7 +144,40 @@ steps:
         /workspace/bin/oscoin-cli
       )
 
-      echo "$${bins[*]}" | xargs -P8 -n1 ./scripts/docker/build-image.sh
+      echo "$${bins[*]}" | xargs -P8 -n1 ./scripts/docker/publish-image.sh
+
+  - id: Deploy
+    waitFor:
+    - Publish images
+    name: "gcr.io/cloud-builders/gcloud"
+    entrypoint: bash
+    env:
+    - BRANCH_NAME=$BRANCH_NAME
+    - OSCOIN_VERSION=$COMMIT_SHA
+    - OSCOIN_NETWORK=devnet
+    - OSCOIN_BENEFICIARY=0x0000000000000000000000000000000000000000
+    - OSCOIN_REPLICAS=3
+    - OSCOIN_GOSSIP_PORT=6942
+    - OSCOIN_API_PORT=8080
+    - OSCOIN_METRICS_PORT=8081
+    - OSCOIN_DEPLOY=bin/oscoin-deploy
+    - GCE_PROJECT_ID=$PROJECT_ID
+    - GCE_REGION=europe-west1
+    - GCE_VPC=devnet
+    - GCE_SERVICEACCOUNT=oscoin-node@opensourcecoin.iam.gserviceaccount.com
+    - GCE_MACHINE_TYPE=n1-standard-1
+    - GCE_DNS_ZONE=oscoin-internal
+    - GCE_DNS_DOMAIN=oscoin.internal
+    args:
+    - "-c"
+    - |
+      set -euo pipefail
+
+      if [[ "$BRANCH_NAME" == "master" ]]; then
+        ./scripts/deploy/gce/node
+      else
+        DRY_RUN=yes ./scripts/deploy/gce/node
+      fi
 
   - id: "Publish Haddocks"
     waitFor:
@@ -161,6 +194,4 @@ steps:
 
       gsutil -m rsync -d -r api-docs gs://oscoin-apidocs/${COMMIT_SHA}
 images:
-- "eu.gcr.io/$PROJECT_ID/oscoin"
-- "eu.gcr.io/$PROJECT_ID/oscoin-cli"
 - "eu.gcr.io/$PROJECT_ID/haskell-scratch:integer-gmp"
--- a/package.yaml
+++ b/package.yaml
@@ -47,6 +47,7 @@ library:
     - fast-logger
     - filepath
     - formatting
+    - generic-lens
     - generics-sop
     - gossip
     - hashable
@@ -64,6 +65,7 @@ library:
     - multihash-cryptonite
     - multihash-serialise
     - network
+    - network-uri
     - optparse-applicative
     - prettyprinter
     - pretty-terminal
@@ -79,6 +81,7 @@ library:
     - stm >=2.4.5.0
     - text
     - transformers
+    - unix
     - unordered-containers
     - vector
     - wai
@@ -114,6 +117,17 @@ executables:
       ld-options: -static
     dependencies:
       - oscoin
+
+  oscoin-deploy:
+    main: exe/deploy/Main.hs
+    when:
+    - condition: "flag(static)"
+      ld-options: -static
+    dependencies:
+      - optparse-applicative
+      - oscoin
+    ghc-options:
+      - -with-rtsopts=-I0
 
 tests:
   documentation:
--- a/scripts/deploy/gce/node
+++ b/scripts/deploy/gce/node
@@ -1,0 +1,136 @@
+#!/usr/bin/env bash
+
+set -eoux pipefail
+
+#
+# Deploy a full node on GCE
+#
+# See ./bootstrap for assumptions about the environment.
+#
+# Note: the 'gcloud' beta commands have to be installed (`gcloud components
+# install beta`).
+#
+
+DRY_RUN=${DRY_RUN:-""}
+
+OSCOIN_VERSION=${OSCOIN_VERSION:-$(git rev-parse HEAD)}
+OSCOIN_NETWORK=${OSCOIN_NETWORK:?Network not set}
+OSCOIN_REPLICAS=${OSCOIN_REPLICAS:-3}
+OSCOIN_GOSSIP_PORT=${OSCOIN_GOSSIP_PORT:-6942}
+OSCOIN_API_PORT=${OSCOIN_API_PORT:-8080}
+OSCOIN_METRICS_PORT=${OSCOIN_METRICS_PORT:-8081}
+OSCOIN_BENEFICIARY=${OSCOIN_BENEFICIARY:?Beneficiary address not set}
+
+GCE_PROJECT_ID=${GCE_PROJECT_ID:-opensourcecoin}
+GCE_REGION=${GCE_REGION:-europe-west1}
+GCE_VPC=${GCE_VPC:-$OSCOIN_NETWORK}
+GCE_SERVICEACCOUNT=${GCE_SERVICEACCOUNT:-oscoin-node@opensourcecoin.iam.gserviceaccount.com}
+GCE_MACHINE_TYPE=${GCE_MACHINE_TYPE:-n1-standard-1}
+GCE_DNS_ZONE=${GCE_DNS_ZONE:-oscoin-internal}
+GCE_DNS_DOMAIN=${GCE_DNS_DOMAIN:-oscoin.internal}
+GCE_NUM_SSDS=${GCE_NUM_SSDS:-1};
+GCE_NUM_SSDS=$((GCE_NUM_SSDS > 8 ? 8 : (GCE_NUM_SSDS < 1 ? 1 : GCE_NUM_SSDS)))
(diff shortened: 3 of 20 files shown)


cloudbuild.yaml
```diff
@@ -144,7 +145,40 @@ steps:
...
+  - id: Deploy
+    waitFor:
+    - Publish images
+    name: "gcr.io/cloud-builders/gcloud"
+    entrypoint: bash
+    env:
+    - BRANCH_NAME=$BRANCH_NAME
+    - OSCOIN_VERSION=$COMMIT_SHA
+    - OSCOIN_NETWORK=devnet
+    - OSCOIN_BENEFICIARY=yrbonyyyoeyfoebik1y6pmradnkyjakqj9s14ckf4kgwoetoskd9apqwan7wwgrfmweoydesdero
```
adinapoli-mndc: Funny how this address has the word `desde` at the end, which apparently it's a valid Spanish word 😀

package.yaml
```diff
@@ -47,6 +47,7 @@ library:
     - fast-logger
     - filepath
     - formatting
+    - generic-lens
```
adinapoli-mndc: yay for `generic-lens` 😉

scripts/deploy/gce/node
```diff
@@ -0,0 +1,136 @@
...
+
+OSCOIN_DEPLOY=${OSCOIN_DEPLOY:-cabal v2-run -- oscoin-deploy}
+
+# create machine config via cloud-init
+_clowninit=$(mktemp "/tmp/node.cloudinit.yaml.XXXXX")
+$OSCOIN_DEPLOY node \
+    --format=cloud-init \
+    --version="$OSCOIN_VERSION" \
+    --network="$OSCOIN_NETWORK" \
+    --host="0.0.0.0" \
```
adinapoli-mndc: I think that if we want to execute on our plan of syncing via the HTTP API and by using the (handshaked) `NodeInfo`, we would need the (real) external IP of the machine here, otherwise all the peers would advertise something like `Addr { host = "0.0.0.0", port = 8080 }`. If you think this is not feasible, perhaps we should extend the handshake routine and _not_ use the IP address passed via the CLI but read it (effectfully) when the application starts?

src/Oscoin/Deployment/Internal/CloudInit.hs
```diff
@@ -0,0 +1,241 @@
...
+    { name   :: Text
+    , uid    :: Maybe Word16
+    , groups :: Set Group
+    } deriving (Eq, Ord, Generic)
+
+instance ToJSON User where
+    toJSON u = Aeson.object $ catMaybes
+        [ Just $ "name" .= view (the @"name") u
+        , ("uid" .=) . toJSON <$> uid u
+        , if Set.null (view (the @"groups") u) then
```
adinapoli-mndc: I know that this is an internal tool and it's pointless to antagonise on it (plus, I do understand that you were having some good fun with `generic-lens`) but I find interesting how you went for `view (the @"groups") u` rather than the more succinct `groups u` here 😉

src/Oscoin/Deployment/Internal/CloudInit.hs
```diff
@@ -0,0 +1,241 @@
...
+import           Lens.Micro.Extras (view)
+import           Numeric (showIntAtBase)
+import qualified System.Posix as Posix
+
+data CloudConfig = CloudConfig
+    { users       :: Vector User
+    , groups      :: Vector Group
+    , write_files :: Vector WriteFiles
+    , runcmd      :: Vector Text
+    , bootcmd     :: Vector Text
```
adinapoli-mndc: Genuine question -- what's the advantage of using vectors here?

test/Integration/Test/Executable.hs
```diff
@@ -95,7 +95,8 @@ testSmoke :: Assertion
 testSmoke = do
     ports@Ports { apiPort, ekgPort, metricsPort } <- randomPorts
     withOscoinExe ports $ \stdoutHdl _ -> do
-        out <- C8.hGet stdoutHdl 2048
+        threadDelay 2000000
```
adinapoli-mndc: Any way we could avoid the artificial delay here? I can see how it might cause the tests to become slightly less deterministic. If it was inserted with the intent of account for boot time, shouldn't `hGet` block up until we have read the requested bytes or that is only the max buffer size and it might read actually less (too lazy to RTFM 😛 ). Alternatively, I guess we could try to loop over reads to `hGet` up until we don't read the requested number of bytes (which would require extending a little bit the tests to keep around a buffer of what we have read so far).

scripts/deploy/gce/node
kim: The "self" address is [resolved during startup](https://github.com/oscoin/oscoin/blob/master/src/Oscoin/P2P.hs#L133). Am I missing something?

src/Oscoin/Deployment/Internal/CloudInit.hs
kim: I am using `DuplicateRecordFields` everywhere due to lack of imagination. That doesn't always work well, but inference didn't fail me so far when using lenses.

src/Oscoin/Deployment/Internal/CloudInit.hs
kim: [Fewer allocations](https://hackage.haskell.org/package/aeson-1.4.3.0/docs/Data-Aeson.html#t:Array)

test/Integration/Test/Executable.hs
kim: Yeah idk. This started to fail at some point, but I couldn't be arsed to investigate. One would think that `hGet` blocks, but apparently it doesn't.

package.yaml
kim: Isn't `view (the @"foo")` cute?

(Between the two of us: it does break all module export restrictions (_sic_), though :no_mouth:)

src/Oscoin/Node/Options.hs
```diff
@@ -81,7 +81,7 @@ nodeOptionsParser cps = Options
                \the development environment, and is useful to avoid busy looping \
                \in an idle network."
        <> metavar "SECONDS"
-       <> value 1
+       <> value (1 * seconds)
```
geigerzaehler: This now shows `(default: 1000000000)` in the help text. I guess we need to revert this and `map (* seconds)` over the whole option.

src/Oscoin/Deployment/Node/Options.hs
```diff
@@ -0,0 +1,80 @@
...
+    <$> option str
+        ( long    "version"
+       <> help    "Version to deploy. Should be commit SHA1."
+       <> metavar "STRING"
+        )
+    <*> cloudProviderParser
+    <*> option (eitherReader readConfigFormat)
+        ( long    "format"
+       <> help    "Machine configuration format"
+       <> value   CloudInit
```
geigerzaehler: This shows `(default: CloudInit)` in the help text but the parser only accepts `cloud-init`. It would also be great if we could document the possible formats, i.e. only `cloud-init` at the moment.

scripts/deploy/gce/node
adinapoli-mndc: I think perhaps I am the one missing something, as I think I have (in my previous PR) passed to the gossip the CLI address rather than the `self` one 🙈 : 

```haskell
            let myHost     = P2P.numericHost optHost
            let myApiInfo  = P2P.mkAddr myHost optApiPort
            let myNodeInfo = P2P.mkNodeInfo myApiInfo nid
            let mySK       = snd keys
            gossip <- managed $
                withGossip telemetry
                           P2P.BootstrapInfo
                               { P2P.bootNodeId      = pure myNodeInfo
                               , P2P.bootHttpApiAddr = pure myApiInfo
                               , P2P.bootGossipAddr  = P2P.mkAddr myHost optGossipPort
                               }
                           (Set.map (Nothing,) <$> disco)
                           (storage node)
                           (Handshake.secureHandshake
                               mySK
                               myNodeInfo
                               (P2P.Disco.optNetwork optDiscovery'))
```

src/Oscoin/Deployment/Internal/CloudInit.hs
adinapoli-mndc: I see -- didn't think these were a concern for a deployment tool, good to know 😅

src/Oscoin/Node/Options.hs
kim: Yeah, our Go `Duration` is just so utterly useless. I'll change this to a `Word8`.

adinapoli-mndc: _Alfredo casually drops https://github.com/oscoin/oscoin/issues/518 and leaves. 😬_

src/Oscoin/Deployment/Internal/CloudInit.hs
kim: Conversely: what's the issue with vectors? They're almost always a better default than lists. One could argue that `users` and `groups` should be `Set`, but that would require unsound `Eq` instances.

cloudbuild.yaml
```diff
@@ -144,7 +144,40 @@ steps:
...
```
cloudhead:
```suggestion
    - OSCOIN_BENEFICIARY=0x0000000000000000000000000000000000000000
```
This should work now.

PR #583, opened 2019-05-23
Make withNode and nodeState polymorphic in the seal

adinapoli-mndc:
This PR makes some of the functions inside `Test.API.HTTP.Helpers` a bit more polymorphic (in the seal) so that `nodeState` and `withNode` can be used also elsewhere (namely, in the tests for node syncing, to name one) skipping the tedious step of crafting a full-fledged `Node.Handle`.

### DIFF
--- a/test/Oscoin/Test/HTTP/Helpers.hs
+++ b/test/Oscoin/Test/HTTP/Helpers.hs
@@ -1,7 +1,8 @@
 {-# OPTIONS_GHC -fno-warn-orphans #-}
 
 module Oscoin.Test.HTTP.Helpers
-    ( NodeState(..)
+    ( DummySeal
+    , NodeState(..)
     , nodeState
     , emptyNodeState
 
@@ -9,6 +10,7 @@ module Oscoin.Test.HTTP.Helpers
     , runSession
     , runEmptySession
     , runSessionWithState
+    , runSessionWithState'
     , liftWaiSession
     , liftNode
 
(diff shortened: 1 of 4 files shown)


test/Oscoin/Test/HTTP/Helpers.hs
```diff
@@ -191,38 +201,48 @@ emptyBlockchain :: IsCrypto c => Blockchain c (Tx c) DummySeal
...
+       , MonadIO m
+       , Serialise s
+       , Ord s
+       )
+    => s
+    -- ^ A seal for the genesis block.
+    -> [(Text, DummyPayload)]
+    -> Session c s ()
+    -> m ()
+runSessionWithState seal bindings (Session sess)= do
```
geigerzaehler: What do you think about splitting this up into a polymorphic `runSessionWithState'` and a monomorphic `runSessionWithState`. This would make the function more consistent with `runSession` and `runEmptySession`. It would also eliminate the `""` value in the tests that is totally non-obvious.

adinapoli-mndc: Thought about that, but was wondering if the fact this is used in only 2 places outside this module warranted the need for an extra function, but if that's your preference, I am happy to do that 😉

adinapoli-mndc: 6b459103

PR #594, opened 2019-05-23
Node syncing: Finalise IO data fetcher and add tests for the RealWord implementation

adinapoli-mndc:
This PR completes the `DataFetcher` for the IO (RealWorld) implementation, fixes the http streaming for `GetBlocks` and adds some tests (running in IO, against the "real deal(TM)").

What is still missing (will arrive as part of one final PR):

- [ ] Wire the syncing process to the `Protocol`, at startup;
- [ ] Support `GetBlockHeaders` efficiently (that would require extending the HTTP API, and we might decide to postpone till devnet);
- [ ] Address previous PR's review comments.

_This PR address some (but not most of) your previous PR's feedback. I will do that in one last final PR where I will also wire the syncing with the `Protocol`._


src/Oscoin/API/HTTP/Internal.hs
```diff
@@ -146,6 +154,14 @@ runApi :: Api c tx s i ()
 runApi app port hdl =
     runSpock port (mkMiddleware app hdl)
 
+-- | Like 'runApi', but it doesn't display the 'Spock is running' banner.
```
kim: Aww, please make this the default!

src/Oscoin/Protocol/Sync.hs
```diff
@@ -89,11 +89,18 @@ data SyncError =
       RequestTimeout ProtocolRequest Addr Timeout
       -- ^ The given peer exceeded the request timeout when serving this
       -- 'SyncProtocolRequest'.
+    | RequestNetworkError ProtocolRequest Addr String
+    -- ^ NOTE (adn) Ideally this would carry an exception, but neither
+    -- 'SomeException' nor 'HttpException' implements 'Eq'. We could theoretically
```
kim: And... why do we need `Eq`?

src/Oscoin/Protocol/Sync/RealWorld.hs
```diff
@@ -67,65 +62,104 @@ newDataFetcher
...
+                    cborBlob <- HTTP.brRead (HTTP.responseBody res)
+                    let (blks :: API.Result [Block c tx (Sealed c s)]) =
+                            CBOR.deserialise . toS $ cborBlob
+                    case blks of
+                      API.Ok b -> pure $ map blockHeader b
+                      API.Err err -> throwM $ RequestNetworkError GetTip peerAddr (toS err)
+            pure $ join r
+
+-- | Streams the input list one element at the time.
+streamList :: Monad m => ConduitT [a] a m ()
```
kim: Isn't this just `Data.Conduit.Combinators.concat`?

test/Test/Oscoin/Protocol/Sync.hs
```diff
@@ -173,59 +222,175 @@ prop_getBlockHeaders_sim = property $ do
...
+    testPeer1@(peer1, chain1) <- forAll (genNakamotoPeer d)
+    (peer2, _) <- forAll (genNakamotoPeer d)
+    (peer3, _) <- forAll (genNakamotoPeer d)
+
+    node1 <- liftIO $ spinUpNode d testPeer1
+    node2 <- liftIO $ spinUpNode d (peer2, unsafeToBlockchain [nakamotoGenesis d])
+    node3 <- liftIO $ spinUpNode d (peer3, chain1)
+
+    -- Give time to the nodes to boot up.
+    liftIO $ threadDelay 1000000
```
kim: I wonder if this could be made more robust. `runSpock` is just `spockAsApp >>= Wai.run`, and wai's `Settings` support a [thing to run after the listen socket is bound](https://hackage.haskell.org/package/warp-3.2.27/docs/Network-Wai-Handler-Warp.html#v:setBeforeMainLoop).

src/Oscoin/API/HTTP/Internal.hs
adinapoli-mndc: Hehe, yes, the default banner can be annoying -- definitely 👍 to make this the default, unless @geigerzaehler or @cloudhead are really attached to the banner 😁

src/Oscoin/Protocol/Sync.hs
adinapoli-mndc: Only for tests, alas!

src/Oscoin/Protocol/Sync/RealWorld.hs
adinapoli-mndc: Is it! You spotted the `conduit` noob 🙋‍♂

test/Test/Oscoin/Protocol/Sync.hs
adinapoli-mndc: oh, very sweet -- I'll investigate into that, thanks!

src/Oscoin/API/HTTP/Internal.hs
kim: If we could instead funnel it through the standard logger, that'd be great

adinapoli-mndc: I suppose we could extend `runApiWithBanner` to be a bit more powerful -- namely, to be passed a `Logger` and perhaps this `IO ()` callback to be used for testing. We could even rename it to simply `runApi'` which would now be a (more complex) version of `runApi`, which could then be implemented in terms of `runApi'`. Will give it a stab 😏

adinapoli-mndc: Oh, but we will still need a 3rd function for the banner/no-banner I suppose? 🤔 (which again could be implemented by `runApi'` with a boolean-of-sort to display or not the banner).

adinapoli-mndc: 98e8dcb6 Didn't touch `runApi` though, we can do that later in a separate PR.

src/Oscoin/Protocol/Sync/RealWorld.hs
adinapoli-mndc: 98e8dcb6

test/Test/Oscoin/Protocol/Sync.hs
adinapoli-mndc: 98e8dcb6

Issue #590, opened 2019-05-23
OscoinTx support for SQLite BlockStore


cloudhead:
The current SQLite backend doesn't support OscoinTx. We should implement `FromRow` and `ToRow` instances for `OscoinTx`, as well as a `Serialise` instance.

The integration plan could be as follows:
1. Switch `node/Main.hs` to use the **in-memory** blockstore with the **old** (current) `Tx` type.
2. Merge the new SQLite (OscoinTx) backend.
3. Once things are running fine, switch `Node` over completely from `Tx` to `OscoinTx`, and move back to the SQLite blockstore.

@kim : does this create any complications in terms of the deployment? Specifically if `blockstore.db` is missing, and/or when switching from memory to sqlite?

kim:
I am not sure about the benefits of using the in-memory blockstore, vs. just writing to two tables for a while. Since there are no transactions, I also don't see a problem with just migrating the schema.

That being said, SQLite creates the `blockstore.db` if it doesn't exist, all we do during deployment is mount a directory on an ephemeral SSD into the container.

cloudhead:
Closed by #607

PR #595, opened 2019-05-23
CBOR instances for TxPayload

cloudhead:
In preparation for #590 and #568

### DIFF
--- a/src/Oscoin/Data/Ledger.hs
+++ b/src/Oscoin/Data/Ledger.hs
@@ -24,12 +24,17 @@ import           Oscoin.Prelude
                  , Word8
                  , div
                  , fromIntegral
+                 , liftA2
                  , map
                  , maxBound
                  , mod
+                 , show
                  , sum
                  , ($)
+                 , (++)
                  , (.)
+                 , (<$>)
+                 , (<*>)
                  , (<>)
                  )
 
@@ -38,7 +43,11 @@ import           Oscoin.Crypto.Blockchain.Block (BlockHash)
 import qualified Oscoin.Crypto.Hash as Crypto
 import           Oscoin.Crypto.PubKey (PublicKey, Signature, Signed)
 
+import           Codec.Serialise (Serialise)
 import qualified Codec.Serialise as CBOR
+import qualified Codec.Serialise.Decoding as CBOR
+import qualified Codec.Serialise.Encoding as CBOR
+import           Control.Monad.Fail (fail)
 import           Crypto.Data.Auth.Tree (Tree)
 import qualified Crypto.Data.Auth.Tree as WorldState
 import           Data.ByteString (ByteString)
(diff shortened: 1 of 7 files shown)


src/Oscoin/Data/Ledger.hs
```diff
@@ -24,12 +24,17 @@ import           Oscoin.Prelude
...
                  , map
                  , maxBound
                  , mod
+                 , show
                  , sum
                  , ($)
+                 , (++)
                  , (.)
+                 , (<$>)
+                 , (<*>)
```
kim: :eyes:

src/Oscoin/Data/Ledger.hs
```diff
@@ -225,6 +259,28 @@ deriving instance (Eq (Crypto.Hash c), Eq (AccountId c))     => Eq   (Dependency
...
+    encode (Undepend proj) =
+           CBOR.encodeListLen 2
+        <> CBOR.encodeWord 1
+        <> CBOR.encode proj
+
+    decode = do
+        pre <- liftA2 (,) CBOR.decodeListLen CBOR.decodeWord
+        case pre of
+            (3, 0) -> Depend   <$> CBOR.decode <*> CBOR.decode
+            (2, 1) -> Undepend <$> CBOR.decode
```
adinapoli-mndc: Small nit -- why not pattern match first only on the tag (which uniquely identifies the constructor) and then check that the length is the requested one? I guess now it doesn't matter, but it would give you more precise errors in case you successfully decoded the tag but the length mismatched -- if anything, you will know which constructor failed to decode vs now where the failure is quite opaque. What do you think?

src/Oscoin/Data/Ledger.hs
```diff
@@ -225,6 +259,28 @@ deriving instance (Eq (Crypto.Hash c), Eq (AccountId c))     => Eq   (Dependency
...
+        <> CBOR.encodeWord 0
+        <> CBOR.encode proj
+        <> CBOR.encode hsh
+    encode (Undepend proj) =
+           CBOR.encodeListLen 2
+        <> CBOR.encodeWord 1
+        <> CBOR.encode proj
+
+    decode = do
+        pre <- liftA2 (,) CBOR.decodeListLen CBOR.decodeWord
```
adinapoli-mndc: Apparently also the list len has its own canonical version -- `decodeListLenCanonical`

src/Oscoin/Data/OscoinTx.hs
```diff
@@ -79,6 +115,61 @@ data TxMessage c =
...
+    decode = do
+        pre <- liftA2 (,) CBOR.decodeListLen CBOR.decodeWord
+        case pre of
+            (2, 0) -> TxRegisterProject   <$> CBOR.decode
+            (2, 1) -> TxUnregisterProject <$> CBOR.decode
+            (3, 2) -> TxAuthorize         <$> CBOR.decode <*> CBOR.decode
+            (3, 3) -> TxDeauthorize       <$> CBOR.decode <*> CBOR.decode
+            (5, 4) -> TxCheckpoint        <$> CBOR.decode <*> CBOR.decode <*> CBOR.decode <*> CBOR.decode
+            (2, 5) -> TxUpdateContract    <$> CBOR.decode
+            (3, 6) -> TxTransfer          <$> CBOR.decode <*> CBOR.decode
```
adinapoli-mndc: Same as above -- perhaps it would help with more precise error reporting if you tuck the list-len checking inside each branch, after decoding the tag.

src/Oscoin/Data/Ledger.hs
cloudhead: Yeah :thinking: the intent was to um keep track of what we used since the "contracts" are implemented in here, but it's a bit annoying because there's other stuff in this module too, and it's difficult to split the module apart.

src/Oscoin/Data/Ledger.hs
cloudhead: Hmm I don't think it matters too much as long as we only have this implementation, since the roundtrip is tested. Perhaps in the future it would be helpful to know exactly what was wrong?

adinapoli-mndc: Yeah, this is more for the future -- the good news is that we can change it later as it won't alter the structure of the payload, only the way we decode it.

PR #596, opened 2019-05-24
Node syncing - cleanup the API and address review comments

adinapoli-mndc:
This PR addresses previous comments from @kim and @cloudhead. On top of that, it also cleans up the API considerably, and make the syncing code genuinely polymorphic.

After we merge this one, only two things will be missing (coming in one last final PR):

- [ ] Wire the syncing code to the protocol, at startup;
- [ ] Think about ways we can test the `syncNode` function (as that is essentially an endless loop);
- [ ] Telemetry


src/Oscoin/Protocol/Sync.hs
```diff
@@ -176,35 +201,35 @@ data SyncContext c tx s m = SyncContext
     -- variation of it.
     , scLocalChainReader  :: BlockStoreReader c tx s m
     -- ^ A read-only handle to the local blockchain.
-    , scUpstreamConsumers :: [SyncEvent c tx s -> m ()]
+    , scEventHandlers :: [SyncEvent c tx s -> m ()]
     -- ^ A list of \"upstream consumers\", represented as effectful actions from
```
cloudhead: Comment needs updating :)

src/Oscoin/Protocol/Sync.hs
```diff
@@ -1,9 +1,14 @@
+-- | Syncing nodes in oscoin
+--
+-- Based on the spec available at https://hackmd.io/2cPkrWTjTIWo2EmHZpPIQw
```
cloudhead: This is not permanent, I'd open a separate PR so we can merge it into the repo perhaps.

adinapoli-mndc: Yeah. I have been thinking about it, and I'd like to experiment with post-processing in hackmd so perhaps we can render all those nice graphs & latex into a `.pdf` (or similar). In the meantime I have opened an issue to track that (cfr. #597 ). Do you think we can tackle this separately?

src/Oscoin/Protocol/Sync.hs
adinapoli-mndc: Can do here or in my (final) PR (which I will hopefully submit by the end of the day) -- whatever works best for you 😉

src/Oscoin/Protocol/Sync.hs
cloudhead: Sure :+1:

src/Oscoin/Protocol/Sync.hs
cloudhead: :man_shrugging:

PR #599, opened 2019-05-24
Change 'Evaluator' type to prepare for OscoinTx

### LINKED ISSUE #591: Reward beneficiary with block coinbase
### Problem

Block evaluation as implemented abstractly by `LedgerStore` is incompatible with
the evaluation required by the Oscoin ledger. This blocks the integration of OscoinTx (https://github.com/oscoin/oscoin/issues/568)

When evaluating a block (i.e. using the block data to produce a new state) the
Oscoin ledger will do the following.

1. Credit `blockDataBeneficiary` with the coinbase
2. Evaluate all transactions in turn using the `blockDataBeneficiary` as an
   additional input.

This means that the block evaluation function for `Block c tx s` 

~~~haskell
applyTx :: Beneficiary c -> tx -> TxState tx -> (TxState tx, TxOutput tx)
applyBeneficiary :: Beneficiary c -> TxState tx -> TxState tx
~~~

In our current system the `LedgerStore` handles block evaluation. The
`LedgerStore` only accepts an evaluation function of the type

~~~haskell
eval :: tx -> TxState tx -> Either EvalError (TxState tx, TxOutput tx)
~~~

This means that the `LedgerStore` is incompatible with the Oscoin ledger.

### Suggested Solution

Instead of the `LedgerStore` receiving and evaluator of the type given above it
will receive an evaluator of the type

~~~haskell
evalBlock :: Beneficiary c -> [tx] -> TxState tx -> (TxState tx, [TxOutput tx])
~~~

The Oscoin Ledger will provide a function of this type by combining `applyTx`
and `applyBeneficiary`.

In the code we would change the `Evaluator` type alias to the type of
`evalBlock`. In addition the legacy tx (defined in `Oscoin.Data.Tx`) and all the
test tx types would be changed to use the new eval block type.

geigerzaehler:
We change the `Evaluator` type to make it compatible with the evaluation functions of the Oscoin ledger. 

Fixes #591

This commit eliminates the tests for `buildBlock`. The function is now mainly glue code and tests would require substantial ammounts of code.

### DIFF
--- a/test/Test/Oscoin/DummyLedger.hs
+++ b/test/Test/Oscoin/DummyLedger.hs
@@ -11,7 +11,7 @@ module Test.Oscoin.DummyLedger
     ( DummyTx(..)
     , DummyState
     , DummyOutput
-    , dummyEval
+    , dummyEvalBlock
     ) where
 
 
@@ -63,5 +63,5 @@ type DummyState = [DummyTx]
 type DummyOutput = DummyTx
 
 -- | Prepends the transaction to the state and output the transaction.
-dummyEval :: Evaluator DummyState DummyTx DummyOutput
-dummyEval tx txs = Right (tx, tx:txs)
+dummyEvalBlock :: Evaluator c DummyState DummyTx DummyOutput
+dummyEvalBlock _beneficiary txs st = (map Right txs, reverse txs <> st )
--- a/exe/node/Main.hs
+++ b/exe/node/Main.hs
@@ -116,7 +116,7 @@ main = do
                     gen
             -- FIXME(adn) Replace with a proper evaluator & state once we switch to
             -- the OscoinTx type.
-            let dummyEval _ s = Right ([], s)
+            let dummyEval _ _ s = ([], s)
             ledger <- liftIO $
                 Ledger.newFromBlockStoreIO dummyEval (fst blkStore) mempty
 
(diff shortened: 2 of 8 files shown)


test/Test/Oscoin/DummyLedger.hs
```diff
@@ -62,6 +63,10 @@ type DummyState = [DummyTx]
 
 type DummyOutput = DummyTx
 
+-- -- | Prepends the transaction to the state and output the transaction.
+-- dummyEval :: Evaluator DummyState DummyTx DummyOutput
+-- dummyEval tx txs = Right (tx, tx:txs)
```
cloudhead: Can we kill this?

test/Test/Oscoin/DummyLedger.hs
```diff
@@ -11,7 +11,8 @@ module Test.Oscoin.DummyLedger
     ( DummyTx(..)
     , DummyState
     , DummyOutput
-    , dummyEval
+    -- , dummyEval
```
cloudhead: Can we kill this?

test/Test/Oscoin/DummyLedger.hs
geigerzaehler: Yes indeed! Friday afternoon artifact ;)

test/Test/Oscoin/DummyLedger.hs
geigerzaehler: See above

PR #601, opened 2019-05-24
Node syncing

### LINKED ISSUE #531: Node syncing

adinapoli-mndc:
Fixed #531 . The results are encouraging, but as "there will be bugs" I have marked this as a WIP so that I can allocate Monday (and maybe even Tuesday) for the hunt session. However, opening a final PR and Friday evening always give good vibes, so here we go :shipit: .

I have tested this very briefly with two nodes so started:

```
stack exec oscoin -- --genesis data/genesis.yaml --blockstore blockstore1.db \
--block-time-lower 5 --beneficiary 0x0000000000000000000000000000000000000000 \
--enable-mdns --api-port 7777 --gossip-port 6000  --network devnet
```

```
stack exec oscoin -- --genesis data/genesis.yaml --blockstore blockstore2.db \
--block-time-lower 5 --beneficiary 0x0000000000000000000000000000000000000000 \
--enable-mdns --api-port 8888 --gossip-port 7000 --network devnet
```

Where the first one starts immediately and the second after ~30 secs. Upon starting up, node2 notices that it's lagging up:

```
Node syncing started local_tip=block_hash=2DrjgbEiCx2ZWa2ob5HgZMaDFTgaY1r2aFDRGbFQKzhJRnfkZK local_height=0 remote_tip=block_hash=2DrjgbCzUrFfo6UsWkBjiccNutJ181MubqrvsyaXPXuTUGJ21H remote_height=8  ns="sync" loc="Oscoin.Protocol.Sync:197:7"
```

And node1 start sending some blocks:

```
debug | handshake complete peer_addr=127.0.0.1:56754 peer_nodeid=2DrjgbLwtFSV6TTLsnhwpPvN5UsaWJnXNis7PUR3NGy71P44cD ns="p2p" loc="Oscoin.P2P:201:13"
info  |  status=200 method=GET path=/blockchain/tip params="" service=144us ns="http-api" loc="Oscoin.Telemetry.Middleware:63:9"
info  |  status=200 method=GET path=/blocks/by-height params="start=1&end=7" service=891us ns="http-api" loc="Oscoin.Telemetry.Middleware:63:9"
info  |  status=200 method=GET path=/blocks/by-height params="start=8&end=8" service=287us ns="http-api" loc="Oscoin.Telemetry.Middleware:63:9"
info  |  status=200 method=GET path=/blockchain/tip params="" service=69us ns="http-api" loc="Oscoin.Telemetry.Middleware:63:9"
```

Now, in no particular order, here are some things which I will need to figure out on Monday:

- [x] Ideally we need a slightly more robust retry/isDone policy, because I have noticed a classic catch22 situation: `node2` thinks (rightly) that is lagging up as the remote tip is `8` and its own is `0`, but now `node1` will think a rollback happened as `node1` tip is < of its own tip. So I think I should explore @cloudhead suggestion to also take into account the block difficulty in the `isDone` calculation (will require amending the spec) so that we do not fire needless HTTP requests and flood the system with blocks we don't need;

- [x] I need to review the entire `retrying` stuff -- not sure I've got the sematic right (I mean, from a point of view of the use of the `retry` library 😅 );

- [x] I need to check this story of the HTTP handshake -- now it does work as everything happens on `localhost`, but @kim sort of planted the seed in my brain that I am not passing the `self` bootstrap address to the handshake, so this won't work on our Gorgonzola Clown infrastructure. 

- [x] The telemetry when a node start syncing can be polished, as it logs that spurious `block_hash=`


.hlint.yaml
```diff
@@ -51,3 +51,6 @@
 - ignore:
     name: Avoid restricted qualification
     within: [Oscoin.Protocol.Sync]
+- ignore:
+    name: Avoid restricted qualification
```
kim: Can we just get rid of this warning entirely?

adinapoli-mndc: 👍 , I find it not very helpful.

adinapoli-mndc:
@cloudhead I think this is ready for prime-time, albeit perhaps it's better if we wait for you to complete your work on the `OscoinTx` before merging this, as I have create and e2e integration test which relies on the SQLite blockstore with a concrete `Tx` type (due to the use of `NodeState`).

src/Oscoin/Protocol/Sync.hs
```diff
@@ -442,3 +451,65 @@ syncBlocks rng = do
...
+        recordEvent $
+            Telemetry.Events.NodeSyncFinished (blockHash localTip, height localTip)
+       else do
+        recordEvent $
+            Telemetry.Events.NodeSyncStarted (blockHash localTip, height localTip)
+                                             (blockHash remoteTip, height remoteTip)
+        for_ (range localTip remoteTip) syncBlocks
+        newLocalTip <- lift (getTip (scLocalChainReader ctx))
+        recordEvent $
+            Telemetry.Events.NodeSyncFinished (blockHash newLocalTip, height newLocalTip)
```
cloudhead: It's hard to make out the actual logic here because of the telemetry - it would be nice if this was factored out.

src/Oscoin/Protocol/Sync/RealWorld.hs
```diff
@@ -85,12 +85,18 @@ newDataFetcher httpManager = DataFetcher $ \peer args ->
...
-                    case CBOR.deserialise . toS $ cborBlob of
-                      API.Ok t    -> pure t
-                      API.Err err -> throwM $ RequestNetworkError GetTip peerAddr (toS err)
+                response <- C.http (mkRequest GetTip mempty peer) httpManager
+                mbBlock  <- runConduit $ HTTP.responseBody response
+                         .| mapOutput (\r ->
+                             case r of
+                                 API.Ok o  ->
+                                     o :: Block c tx (Sealed c s)
+                                 API.Err e -> panic (toS e) -- any better way?
```
cloudhead: Can you add a comment here as to why we panic?

src/Oscoin/Protocol/Sync/RealWorld.hs
```diff
@@ -227,33 +229,20 @@ runSync syncContext (SyncT s) = do
...
-          -> SyncT c tx s m Bool
-      checkSynced ctx _retryStatus () = do
-          -- We need to refetch the tips.
-          remoteTip <- getRemoteTip
-          localTip  <- lift (BlockStore.getTip (scLocalChainReader ctx))
-          pure $ not (isDone (scNu ctx) localTip remoteTip)
+syncNode = forever $ do
+    sync `catchError` \(e :: SyncError) ->
+       recordEvent $ Telemetry.Events.NodeSyncError (toException e)
+    liftIO $ threadDelay 30000000 -- Wait 30 seconds.
```
cloudhead: What's the idea here with waiting 30s?

src/Oscoin/Protocol/Sync.hs
adinapoli-mndc: It's a bit of a shame, yes, but if we factor out the telemetry what we lose is a precise callstack of where this happened -- will think about this a bit more.

src/Oscoin/Protocol/Sync/RealWorld.hs
adinapoli-mndc: Yes -- essentially the reason is that if we get an `Err` there , it means the request didn't succeed, and in theory the impure exception would be caught by the `try` inserted inside `catchNetworkError` and reported as a `SyncError`. Alas that entire expression is pure and I cannot call `throwM` 😏

src/Oscoin/Protocol/Sync/RealWorld.hs
adinapoli-mndc: Mainly just as throttling mechanism -- I have picked a number out of thin air, we can either make this configurable or to pick something different.

src/Oscoin/Protocol/Sync.hs
adinapoli-mndc: 91af9d63

src/Oscoin/Protocol/Sync/RealWorld.hs
adinapoli-mndc: 91af9d63

src/Oscoin/Protocol/Sync/RealWorld.hs
adinapoli-mndc: 91af9d63

PR #607, opened 2019-05-27
Move to OscoinTx

### LINKED ISSUE #492: Oscoin Ledger Tx/State
This issue is here to track initial implementation work on the Ledger, that is the high-level domain logic of oscoin, comprising transactions and state.

6. [x] Serialization of ledger types
7. [x] Integration with codebase and replacement of `Data.Tx`
1. [x] Initial ledger types and functions (#436)
2. [x] Rename "payload" to "message", and allow multiple messages per tx
3. [x] ~~Distinguish between project accounts (funds) and regular accounts~~
4. [x] Tx generators
5. [x] Tests for `applyTx`, `TxTransfer` and `TxRegister/TxUnregister`
8. [x] Use `PublicKey` everywhere in the ledger instead of `Address`
9. [x] Add `Author` field to transaction
10. [x] Block fees

cloudhead:
This switches over the SQLite BlockStore to use `OscoinTx`, and consequently switches `exe/node/Main.hs` to use `OscoinTx`. In the process, I made the SQLite backend monomorphic in the `tx` type, so it now only works with `OscoinTx`.

The legacy `Data.Tx` still remains in the codebase, and it will be removed in a subsequent changeset.

Things missing:
- [x] Roundtrip tests for `StateVal`

Closes #492 #568 #590


src/Oscoin/Storage/Block/SQLite/Tx.hs
```diff
@@ -0,0 +1,162 @@
...
+import           Database.SQLite.Simple as Sql (Connection, Only(..))
+import qualified Database.SQLite.Simple as Sql
+import           Database.SQLite.Simple.FromField (FromField)
+import           Database.SQLite.Simple.FromRow (FromRow, field)
+import           Database.SQLite.Simple.Orphans ()
+import           Database.SQLite.Simple.QQ
+import           Database.SQLite.Simple.ToField (ToField, toField)
+
+-- | Constraints required for a @Tx c@ to be storable in SQLite.
+type StorableTx c =
```
adinapoli-mndc: I know that I am "that guy" and probably a terrible person, but I would be 👎 for creating this type synonym and for exporting it at the top-level. I have resisted very, very hard the temptation to do so for the crypto stuff, and the reason it's simple (_puts on his Blade Runner goggles and interprets the replicant at the end of the movie_): I have seen with my own eyes what happens if you start to leak all these piecemeal type synonyms in type signatures. Soon enough we will create `type Behemoth c = (StorableTx c, Chocobo c, ...)`  and the poor new programmers joining the codebase will have to keep in their head all these layers of indirections to actually get to the constraints. I have been there, it's terrible 😅 

May I propose a middle ground? If you still want to have the conveniency of `StorableTx` _locally for this module_, I am OK with that. The bit I would suggest we refrain from, it's to export it.

Middle ground (of the middle ground): _if_ you need this _only_ in this module _and_ the SQLite module stuff, may I propose you rename it like `InternalStorableTx`? This way, it would be extremely clear this is an internal detail of the SQLite stuff, we other programmers won't be tempted to make it leak into top-level signatures.

What do you think?

src/Oscoin/Storage/Block/SQLite/Tx.hs
```diff
@@ -0,0 +1,162 @@
...
+    , txRowFee       = txFee txPayload
+    , txRowBurn      = txBurn txPayload
+    , txRowAuthor    = txAuthor txPayload
+    }
+
+fromTxRow :: StorableTx c => TxRow c -> Either TxRowDecodeError (Tx c)
+fromTxRow TxRow{..} = first toDecodeError $ do
+    txNetwork   <- pure $ toEnum (fromIntegral txRowNetwork)
+    txSignature <- pure txRowSignature
+    txMessages  <- deserialiseOrFail $ LBS.fromStrict txRowMessages
```
adinapoli-mndc: Wish list: we should create a `Codec.Serialise.Extended` where we export `deserialise'` and `deserialiseOrFail'` which both takes strict `ByteString`s. We have been doing this extra conversion enough in the codebase to be annoying (and repetitive) 😁 Mind opening a ticket? (this is orthogonal to this code review, obviously!)

src/Oscoin/Storage/Block/SQLite/Tx.hs
```diff
@@ -0,0 +1,162 @@
...
+            }
+        , txNetwork
+        , txSignature
+        }
+  where
+    toDecodeError :: DeserialiseFailure -> TxRowDecodeError
+    toDecodeError = TxRowDecodeError . displayException
+
+getTx :: forall c. (StorableTx c, Typeable c) => Sql.Connection -> Hash c -> IO (Maybe (Tx c))
+getTx conn hsh = do
```
adinapoli-mndc: Bikeshed: `lookupTx` maybe? 🤔

src/Oscoin/Storage/Block/SQLite/Tx.hs
```diff
@@ -0,0 +1,162 @@
...
+            , toField txRowNonce
+            , toField txRowFee
+            , toField txRowBurn
+            , toField txRowAuthor
+            , toField blockHash
+            ]
+
+-- | Get the transactions belonging to a block.
+getBlockTxs :: forall c. (StorableTx c, Typeable c) => Connection -> BlockHash c -> IO [Tx c]
+getBlockTxs conn h = do
```
adinapoli-mndc: I find `getTx/getBlockTxs` a bit weird from the API perspective -- the latter doesn't return a `Maybe`, whereas the former does. Now, I do appreciate that probably in the latter case is hard (if not impossible) to distinguish the failure from the fact we return an empty list because there are no transactions for a block, but still my OCD wonder whether we could improve this somehow. It doesn't matter too much, but it would still be nice to think about it 😉

test/Oscoin/Test/Storage/Block/Equivalence.hs
```diff
@@ -102,8 +103,8 @@ type StoresUnderTest c tx s m =
 -- SQLite store to pass through chain selection.
 withStores
     :: forall c a.
-       IsCrypto c
-    => (StoresUnderTest c DummyTx DummySeal IO -> IO a)
+       (IsCrypto c, SQLite.StorableTx c)
```
adinapoli-mndc: Right, if we use the `InternalStorableTx` suggestion, you can still write this, or simply do a copy-paste for the sake of this module 😉

test/Test/Oscoin/Data/OscoinTx.hs
```diff
@@ -1,30 +1,25 @@
-module Test.Oscoin.Data.OscoinTx where
+module Test.Oscoin.Data.OscoinTx (tests, genTx) where
```
adinapoli-mndc: I am grateful for the `genTx` generator -- it will make my life of integrating the last part of the Sync tests with this new `SQLite` store much easier 😉

src/Oscoin/Storage/Block/SQLite/Tx.hs
cloudhead: :thinking: I kind of agree, but basically just copied what was already there: https://github.com/oscoin/oscoin/pull/607/files#diff-b0b113fc3c1a6c626c272326eda78d7cL67 (and also exported) - Let me see how bad it would be not to export it.

src/Oscoin/Data/OscoinTx.hs
```diff
@@ -21,10 +22,34 @@ import           Codec.Serialise (Serialise)
...
+
+instance CBOR.Serialise (StateVal c) => Query (TxState c) where
+    query key (TxState st) = LBS.toStrict . CBOR.serialise <$> WorldState.lookup key st
+
+instance Monoid (TxState c) where
+    mempty = TxState WorldState.empty
+
+instance Semigroup (TxState c) where
+    (<>) (TxState a) (TxState b) =
+        TxState $ WorldState.fromList $ WorldState.toList a <> WorldState.toList b
```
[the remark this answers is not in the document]
cloudhead: It's probably expensive, but only used in tests. I'll add a comment.

src/Oscoin/Data/Ledger.hs
```diff
@@ -312,12 +365,26 @@ data Project c = Project
     , pUpdateContract  :: UpdateContract' c
     }
 
+instance
+    (Serialise (Crypto.Hash c), Serialise (AccountId c), Serialise (Signature c), Ord (AccountId c))
```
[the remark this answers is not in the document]
cloudhead: I wrote so many of these over the last few days that at some point I just stopped bothering :rofl:

src/Oscoin/Storage/Block/SQLite/Tx.hs
cloudhead: Indeed!

src/Oscoin/Storage/Block/SQLite/Tx.hs
cloudhead: I think renaming the former to `lookupTx` is a first step to distinguishing them. I also just copied these over from the old SQLite.Transaction module, so didn't really think about it.

test/Test/Oscoin/Data/OscoinTx.hs
cloudhead: I am grateful for having been able to write `arbitrary = hedgehog genTx` :dancer:

src/Oscoin/Storage/Block/SQLite/Tx.hs
geigerzaehler: I think this synonym is mostly used in the tests where we also require `IsCrpyto c`. We can get rid of it if we extend `IsCrypto c` with the constraints listed here. Also see my comment about the `ToField` and `FromField` instances above. We can get rid of them if we use explicit `fieldFromSerial` instead of just `field` in `FromRow (TxRow c)` below.

src/Oscoin/Data/OscoinTx.hs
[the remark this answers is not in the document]
cloudhead: Yup! aa0231d

src/Oscoin/Storage/Block/SQLite/Tx.hs
cloudhead: Since this synonym is used a lot from `SQLite.Internal`, which imports `Tx`, I have to export it. But what I'll do is rename the `Tx` module to `Internal.Tx`.

cloudhead: b7ddefe

src/Oscoin/Storage/Block/SQLite/Tx.hs
cloudhead: 24a88b1

src/Oscoin/Storage/Block/SQLite/Tx.hs
cloudhead: #638

src/Database/SQLite/Simple/Orphans.hs
```diff
@@ -64,6 +64,15 @@ instance ( CBOR.Serialise msg
...
     toField = toFieldSerial
 
+instance ( CBOR.Serialise (Crypto.Signature c)
+         , Typeable c
+         ) => FromField (Crypto.Signature c) where
+    fromField = fromFieldSerial
+
+instance ( CBOR.Serialise (Crypto.Signature c)
+         ) => ToField (Crypto.Signature c) where
+    toField = toFieldSerial
```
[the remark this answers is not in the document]
cloudhead: I removed one unused instance, but the others are part of `IsCrypto`'s constraints now..

PR #610, opened 2019-05-28
Monitoring

kim:
Makes available a simple monitoring server at https://monitoring.os.computer:3000

The provisioning is rather procedural atm due to various quirks. After running it for another 300x, I might feel confident to collapse more things into the cloud-init config.

**Caveats**

* The prometheus Web UI is not exposed on the internet, as it doesn't support authentication.

   Use SSH port-forwarding to access it: `gcloud compute ssh oscoin-monitoring --zone=europe-west3-a --ssh-flag="-L 9090:localhost:9090"`, or the included script `scripts/gce/port-forward`

* Using port 3000 instead of 443 is a bit unfortunate, we'll have to build a custom `grafana` image with the `net_bind_service` capability set on the executable

* Annoyingly, there is still no sane method to manage a set of "default" dashboards, apart from making a database backup, or saving the dashboard as JSON. 

  **If you adjust the provided dashboards (which are mostly wrong), please grab the JSON and check it into version control**

### DIFF
--- a/scripts/deploy/gce/etc/grafana/provisioning/dashboards/p2p.json
+++ b/scripts/deploy/gce/etc/grafana/provisioning/dashboards/p2p.json
@@ -1,0 +1,317 @@
+{
+  "annotations": {
+    "list": [
+      {
+        "builtIn": 1,
+        "datasource": "-- Grafana --",
+        "enable": true,
+        "hide": true,
+        "iconColor": "rgba(0, 211, 255, 1)",
+        "name": "Annotations & Alerts",
+        "type": "dashboard"
+      }
+    ]
+  },
+  "editable": true,
+  "gnetId": null,
+  "graphTooltip": 0,
+  "id": 2,
+  "links": [],
+  "panels": [
+    {
+      "aliasColors": {},
+      "bars": false,
+      "dashLength": 10,
+      "dashes": false,
+      "datasource": "Prometheus",
+      "fill": 1,
+      "gridPos": {
+        "h": 8,
+        "w": 12,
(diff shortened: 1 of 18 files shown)


scripts/deploy/gce/etc/letsencrypt/renewal-hooks/deploy/fix-permissions
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+
```
geigerzaehler: I think it wouldn’t hurt to add `set -euo pipefail`

scripts/deploy/gce/etc/grafana/provisioning/dashboards/p2p.json
```diff
@@ -0,0 +1,317 @@
+
+tations": {
```
cloudhead: :eyes:

scripts/deploy/gce/etc/letsencrypt/renewal-hooks/deploy/fix-permissions
kim: Only `-e` would make any sense here, but it doesn't matter anyway: this is invoked by `certbot` during a timer run. No-one will be looking anyway. If it fails, Grafana will fail to start back up, with an informative error as to why.

kim:
> answer the question how to look at the dashboard

https://drive.google.com/open?id=1Oa3vYs7K8vl0Fd37oj7lx6YLfoPiyX7X59pAAtYIDIE

PR #611, opened 2019-05-28
Fix silly bug in protocol async worker

adinapoli-mndc:
This PR fixes a silly bug in the `Protocol`, namely that the async worker responsible for fetching the async `dispatchBlock` events from the `TBQueue` wasn't implemented as a proper endless loop, therefore it was fetching the first block and shutting down.

~~I have also simplified a bit the logic around the startup of the `Protocol` by tucking the `Async` internally to the `Handle`~~ (This second bit doesn't work well with testing, will revert).


src/Oscoin/Protocol.hs
```diff
@@ -126,6 +126,14 @@ runProtocol validateFull scoreBlock telemetry bs config use =
         let o = O.emptyOrphanage scoreBlock
         proto               <- newMVar (Protocol bs o validateFull scoreBlock config)
         incomingBlocksQueue <- atomically $ newTBQueue 64
+
+        worker <- Async.async $ forever $ do
+                      block  <- atomically $ readTBQueue incomingBlocksQueue
+                      events <- modifyMVar proto $ \p -> stepProtocol p block
+                      forM_ events (Telemetry.emit telemetry)
+
+        Async.link worker
```
geigerzaehler: Could we document this behavior? “All exceptions raised by the protocol worker are re-thrown to the calling thread“

adinapoli-mndc: I guess we could, although it's kind-of implicit from the documentation of `link` 😉

adinapoli-mndc: aac062c9

geigerzaehler: I actually meant documenting this in the function documentation so that you know about this behavior without looking at the code.

adinapoli-mndc: I see, I see. Ok. 👍

adinapoli-mndc: e1881017 

As I was at it, I have also partially rewrote the top-level documentation altogether, as it was quite stale.

PR #613, opened 2019-05-28
Generate genesis block from parameters

### LINKED ISSUE #589: Oscoin CLI generate genesis compatible with OscoinTx
We have to generate genesis blocks that are compatible with the OscoinTx ledger. Since we don't need any transactions in the genesis, it should always be empty.

This should only require us to adjust the existing CLI command `genesis`.

### LINKED ISSUE #602: Remove remaining JSON instances
### Problem

The code requires a plethora of ToJSON and FromJSON instances to accomplish just one thing: Generate and read the gensis block from a YAML file. To alleviate the burden of maintaing these instances and implementing them for OscoinTx I propose to eliminate them.

### Proposed solutions

We have two solutions to this problem.

#### Storing the genesis block as CBOR

Instead of using YAML to store the genesis block we use the CBOR and the existing Serialise instances.

#### Store only data required to construct Genesis block

We store only the data that is required to construct the genesis block. I.e. the state hash, the timestamp, the beneficiary, the initial difficulty and the pow seal. All other data, like parent hash, data hash, height, etc can be derived from that.

We could then implement some dedicated `genesisFromJson` and `genesisParamsToJson` functions that would not require the ToJSON and FromJSON instances and would be rather minimal.

### Recommendation

I suggest going with the second solution. Decoupling the definition of the genesis block from the serialization of the genesis block will give us more flexibility down the road. For example we could extend the parameters to not include the state hash but an actual definition of the seed state.

geigerzaehler:
Before this commit the node loaded the genesis block from a YAML file. This required us to be able to serialise a whole block as YAML. Now the node only loads the `GenesisParameters` from a YAML file and uses the parameters to build the genesis block.

Conversely, we change the `genesis` command of the CLI to create these `GenesisParameters` from command line arguments.

As a consequence we can remove a multitude of `ToJSON` and `FromJSON` instances and close #602. Note that there still remain some instances for example for `Crypto.ShortHash` that are not trivial to get rid of.

The change also makes it possible to build the genesis block for every type of transaction (since the transaction set is empty). This closes #589.


src/Oscoin/Crypto/Blockchain/Genesis.hs
```diff
@@ -0,0 +1,134 @@
...
+-- of work so that the block generated
+createGenesisParameters
+    :: ( Serialise (Crypto.Hash c)
+       , Crypto.HasHashing c
+       , ByteArrayAccess (BlockHash c)
+       , AuthTree.MerkleHash (Crypto.Hash c)
+       , Serialise (Beneficiary c)
+       ) => Beneficiary c -> Timestamp -> Difficulty -> Maybe (GenesisParameters c)
+createGenesisParameters gpBeneficiary gpTimestamp gpStartDifficulty = do
+    gpPoW <- findPoW blockHeader
```
cloudhead: Kind of odd that creating the parameters involves PoW - I think it would make more sense if this was passed in.

data/genesis.yaml
```diff
@@ -1,13 +1,5 @@
...
-  dataHash: 2DrjgbGEpqLFjKYwNp8hhAYt3KpK2GPgk8aLAvG3Joo454M9tb
-  targetDifficulty: ffff0021
-  stateHash: 2DrjgbJVAd2Q9WQVZT1Vtam2hCNGZSefi5DwPJNbSaxWmuUypo
-  seal: 0
-  timestamp: 1558543390487199604
-  parentHash: 2Drjgb4FtifQz3pNjpYitXQVQXvVG3WBeub54JJnq438GFasNF
+pow: 0
+startDifficulty: ffff0021
+beneficiary: '0x0000000000000000000000000000000000000000'
+timestamp: 1559041196429336548
```
cloudhead: If this is no longer representing a block or block header, then I'd almost say these should be passed as CLI flags :thinking:

geigerzaehler:
@kim: Could you take a quick look? As far as I know, the renaming of a node option should not have any adverse effect.

PR #614, opened 2019-05-28
Use `unagi-chan` in the Protocol

adinapoli-mndc:
Builds on #611.

This PR is not strictly necessary, but it can be considered a "nice to have".

In GHC, when working with concurrent data structures & primitives, a useful property is [fairness](http://hackage.haskell.org/package/base-4.12.0.0/docs/Control-Concurrent-MVar.html). In a nutshell, it ensures that when threads wakes up (after a `takeMVar` for example) they are woken up in a FIFO order. Practically speaking, this is also useful in queue-like data structures which can block and/or when we have multiple threads reading/writing from/into them.

Unfortunately, `STM.TBQueue` doesn't have the strong fairness guarantees of MVars. There is a `Chan` data structure in `base` which offers fairness, but it doesn't provide a bounded version to use.

[unagi-chan](https://hackage.haskell.org/package/unagi-chan) seems to be pretty established these days. I haven't looked too much into the implementation, but it seems like it's using mostly atomic primops and mvars, and therefore fairness guarantees should transfer.

This PR switches to use `unagi-chan` for the `Protocol`. At the moment I do not envision this change to make any difference, but it should ensure that, if by any chance the queue fill up and the threads starts blocking each other, they are woken up in a round-robin fashion, giving some time for the protocol to actually process the incoming blocks.


adinapoli-mndc:
> The library may be of limited usefulness outside of x86 architectures where the fetch-and-add instruction is not available.

@kim is that a concern for us? Do you think we should back off? 🤔

cloudhead:
As long as it doesn't perform *worse* than `TBQueue` when we switch to Risc-V :smile:

kim:
We can turn of HT.

PR #618, opened 2019-05-29
Make Grafana reachable on port 443

kim:
Note that we need to tear down the iptables rule on ExecStopPost, as it
would interfere with certbot.


kim:
This already works. Enterprise wiki updated accordingly.

PR #621, opened 2019-05-29
Update oscoin + p2p dashboards

kim:


kim:
tl;dl (too long, didn't look): networking is not working at all

PR #630, opened 2019-05-30
Improve Orphanage robustness (and fix a couple of bugs)

adinapoli-mndc:
Description will follow tomorrow.


src/Oscoin/Storage/Block/Orphanage.hs
```diff
@@ -270,17 +297,99 @@ storeOrphan
...
+      -- Prunes the input 'ChainCandidate'. If this was the only candidate
+      -- for the root hash, also prune the orphan blocks.
+      -- N.B (adn) Is this sufficient or we need full-flow reference counting?
+      pruneCandidate ::  Map (ChainRoot c) (Candidates c s)
+                     -> (Map (ChainRoot c) (Candidates c s), Bool)
+      pruneCandidate old =
+          case map removeCandidate (Map.lookup rootHash old) of
+            Nothing -> (old, False)
+            Just [] -> (Map.delete rootHash old, True)
+            Just xs -> (Map.insert rootHash xs old, False)
```
geigerzaehler: Shouldn’t we return `True` here according to your diagram? If we remove a chain candidate we should also remove all blocks associated with it.

src/Oscoin/Storage/Block/Orphanage.hs
```diff
@@ -270,17 +297,99 @@ storeOrphan
...
--- the chain from the candidates
-pruneOrphanage
-    :: Ord (BlockHash c)
+-- the chain from the candidates.
+pruneOrphanageShallow
+    :: forall c tx s.
+       ( Ord (BlockHash c)
+       , Eq s
+       )
     => BlockHash c
```
geigerzaehler: I’m struggling to understand the significance of `BlockHash`. In my understanding a precondition is that `rootHash` is the parent hash of the `ChainCandidate`. If so we should definitely document that.

src/Oscoin/Storage/Block/Orphanage.hs
```diff
@@ -270,17 +297,99 @@ storeOrphan
...
+pruneOrphanageDeep
+    :: (Ord (BlockHash c)
+       , Eq s
+       )
+    => BlockHash c
+    -> ChainCandidate c s
+    -> Orphanage c tx s
+    -> Orphanage c tx s
+pruneOrphanageDeep rootHash bestFork@ChainCandidate{candidateChain} o =
+    foldl' (flip bulkDelete)
```
geigerzaehler: `bulkDelete` does not remove `orphans`. This probably introduces a memory leak. Also, we don’t remove chains forking off forks. I have the feeling that maybe `pruneOrphanageDeep` should call `pruneOrphanageShallow` recursively.

test/Test/Oscoin/Storage/Block/Orphanage.hs
```diff
@@ -122,6 +129,139 @@ propSelectBestCandidateComplex Dict = property $ do
...
+                       . reverse
+                       . blocks <$> genBlockchainFrom singleBlock
+          pure (chain1, singleBlock, chain2)
+    forAll gen $ \(chain1, singleBlock, chain2) -> do
+        let fusedChain = reverse (blocks chain1)   `mappend`
+                         OldestFirst [singleBlock] `mappend`
+                         reverse (blocks chain2)
+
+        let o'  = foldr insertOrphan (emptyOrphanage blockScore) (blocks chain1)
+        let o'' = foldr insertOrphan o' (blocks chain2)
```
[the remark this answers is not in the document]
adinapoli-mndc: No -- I simply did forget that function existed, believe it or not 😅

src/Oscoin/Storage/Block/Orphanage.hs
```diff
@@ -270,17 +297,99 @@ storeOrphan
...
+            |v|
+            |-|                       /-
+          /-|||                     /-
+        /-  |||                   /-
+      <-    |||                 <-
+            |v|                         \
+            | |\                         \
+            | | v                         v
+            | |
+            +-+
```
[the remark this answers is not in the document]
adinapoli-mndc: Thanks, I am very proud of it 😁

test/Test/Oscoin/Storage/Block/Orphanage.hs
```diff
@@ -122,6 +129,139 @@ propSelectBestCandidateComplex Dict = property $ do
...
+          chain1      <- genBlockchainFrom root
+          singleBlock <- genBlockFrom (tip chain1)
+          -- Strip @singleBlock@ from @chain2@.
+          chain2      <- unsafeToBlockchain
+                       . List.reverse
+                       . drop 1
+                       . toOldestFirst
+                       . reverse
+                       . blocks <$> genBlockchainFrom singleBlock
+          pure (chain1, singleBlock, chain2)
```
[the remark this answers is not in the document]
adinapoli-mndc: Good call -- I will experiment to see if indeed it clean up things or it makes them a bit clunky 😉

src/Oscoin/Storage/Block/Orphanage.hs
adinapoli-mndc: The problem is that if you _always_ remove all the orphans (this was the case before) if you had any other chain candidate referencing those orphans, it would be wrong to remove them, and this why I am returning `False` here in case we do still have candidates.

However, while talking with you, I think this is a problem only if you remote the "root hash" (which candidates might share) but it's actually fine if delete only the entries pointed by the candidate. I will have a try and see how it turns out.

As I have written in my comment somewhere, in theory the only way we would be sure we are not deleting an orphan which is shared by multiple chains is by using a reference-counting mechanism for the actual orphans, but that complicates things a little bit so I have refrained from doing it just yet as part of this PR.

src/Oscoin/Storage/Block/Orphanage.hs
adinapoli-mndc: yes, your understanding is correct -- we could in theory derive it directly from the `ChainCandidate`, at the cost of calling `lookupChainRoot`. The advantage is that it would simplify the API, making things easier to use for end-users. Will give it a shot.

src/Oscoin/Storage/Block/Orphanage.hs
adinapoli-mndc: Yeah, I do feel the same, but the implementation it's quite head-wrecking so I went with the easiest I could think of, heh.

src/Oscoin/Storage/Block/Orphanage.hs
geigerzaehler: I see. But doesn’t this result in memory leaks because we don’t garbage collect the blocks that were only referenced by the input candidate?

(This brings me to a related question: Why we use a manual pointer scheme with block hashes. Why doesn’t `candidateChain` hold the blocks and we get rid of `orphans`?)

src/Oscoin/Storage/Block/Orphanage.hs
geigerzaehler: Or you could add it as a field to `ChainCandidate`

src/Oscoin/Storage/Block/Orphanage.hs
adinapoli-mndc: > I see. But doesn’t this result in memory leaks because we don’t garbage collect the blocks that were only referenced by the input candidate?

In the current scheme, that would be caught by the `Just []` case, where we would yield `True` and thus delete all the orphans.

> (This brings me to a related question: Why we use a manual pointer scheme with block hashes. Why doesn’t candidateChain hold the blocks and we get rid of orphans?)

I am not opposed to have the blocks directly inside each `candidateChain`. The reason why I didn't want to do that is that I was concerned you could have many chains referencing the same candidate, and a block can potentially be big (say, max 1MB in case we are very, very unlucky). Eventually I have settled on this design as it's more contrived to work with, but should also be more memory efficient (I hope!)

geigerzaehler: Thanks for clarifying. I think it’s nothing to be addressed right away.

src/Oscoin/Storage/Block/Orphanage.hs
geigerzaehler: Ok, fine with me for now.

test/Test/Oscoin/Storage/Block/Orphanage.hs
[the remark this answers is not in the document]
adinapoli-mndc: 34492b94

test/Test/Oscoin/Storage/Block/Orphanage.hs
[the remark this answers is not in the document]
adinapoli-mndc: 34492b94

src/Oscoin/Storage/Block/Orphanage.hs
adinapoli-mndc: 34492b94

src/Oscoin/Storage/Block/Orphanage.hs
```diff
@@ -105,9 +115,17 @@ fromChainSuffix :: (Block c tx (Sealed c s) -> Score)
                 -> OldestFirst NonEmpty (Block c tx (Sealed c s))
                 -> ChainCandidate c s
 fromChainSuffix scoreBlock (toOldestFirst -> blks) = ChainCandidate
-    { candidateChain = Seq.fromList . NonEmpty.toList . map blockHash $ blks
+    { candidateChain = OldestFirst
+                     . Seq.fromList
+                     . NonEmpty.toList
+                     . map blockHash
+                     $ blks
     , candidateTipHeader = blockHeader . NonEmpty.head $ blks
```
geigerzaehler: I just noticed that this is incorrect. Since we get the oldest block first the tip needs to be the last block. Feel free to fix this here or create an issue.

adinapoli-mndc: This was very quick to change, so I did amend it as part of this PR. Probably in a separate ticket though we should add a bunch of tests to check these invariants (i.e. that the tip is what advertised and ditto for the rootHash) as it's always a bit disheartening when tests don't catch such bugs, it means we got too few of them 😉

src/Oscoin/Storage/Block/Orphanage.hs
```diff
@@ -294,26 +447,40 @@ pruneOrphanage rootHash ChainCandidate{candidateChain} o =
...
+    -- references some, it really means that the candidates associated to this
+    -- block needs to be \"prometed\" to the candidates of the parent.
+    mergeCandidates :: Maybe (Orphanage c tx s)
+    mergeCandidates =
+        case liftA2 (,) (Map.lookup parentHash currentCandidates)
+                        (Map.lookup bHash currentCandidates) of
+            Nothing -> Nothing
+            Just (_parentCandidates, myCandidates) -> do
+                let chains' = map (consBlock (scoreBlock o) b) myCandidates
+                pure $ bulkAdjust parentHash chains' (bulkDelete bHash o)
```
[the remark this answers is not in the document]
adinapoli-mndc: Totally agreed on the duplication -- I didn't dare to try simplifying things together as I didn't want the size of this PR to inflate even more. I will quickly try to do the proposed refactoring and, if it proves to me more complicated than needed, tackle this on another PR.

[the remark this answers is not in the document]
adinapoli-mndc: @geigerzaehler Unfortunately those two formulations aren't equivalent (i.e. trying to simplify `extendInFront` the way you suggested makes all the tests fail). I think those two formulations are similar but ever so slightly different: in particular, `mergeCandidates` requires not only that we find a match during the `Map.lookup bHash candidates`, but also that the parent hash matches a list of candidates. I think we should be able to generalise this, but I would be keen on merging this PR first (which contains a lot of useful improvements) and fire another follow-up PR.

What do you think? That's an acceptable compromise for you?

PR #633, opened 2019-05-31
Actually disco forever

kim:
Also:

* emit `NodeIsolated` event if/when we don't have any peers during continuous disco
* tame the giant NotableEvent pattern match a bit.

### DIFF
--- a/src/Oscoin/P2P/Disco/Trace.hs
+++ b/src/Oscoin/P2P/Disco/Trace.hs
@@ -1,0 +1,22 @@
+module Oscoin.P2P.Disco.Trace
+    ( DiscoEvent (..)
+    )
+where
+
+import           Oscoin.Prelude
+
+import qualified Oscoin.P2P.Disco.MDns as MDns
+
+import           Data.IP (IP)
+import qualified Network.DNS as DNS
+import           Network.Socket (PortNumber)
+
+
+data DiscoEvent =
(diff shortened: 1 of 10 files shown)


src/Oscoin/P2P/Disco/Trace.hs
```diff
@@ -0,0 +1,22 @@
+module Oscoin.P2P.Disco.Trace
+    ( DiscoEvent (..)
```
adinapoli-mndc: Nice -- Something else you could do (but I am not sure if circular deps will bite you) is to have a module like `Oscoin.Telemetry.Events.Disco` and put the data type there.

Having said that, I think I am going to steal your pattern here, as I wanted to do the same for syncing events but it didn't occur to me to put the data type in the `Oscoin.Protocol.Sync` namespace, but rather I did create a `Oscoin.Telemetry.Events.Sync`, with the result I had to use a totally opaque stringly-typed exception type as I did hit circular deps 😬

kim: It is indeed necessary to split into many modules due to circular dependencies -- but there's also a modularity argument: the _application_ code defines what events it exposes, and `Oscoin.Telemetry` merely interprets them. If you would think of this namespace as a proper package, with your approach _everything_ would depend on `telemetry`.

PR #634, opened 2019-05-31
Configure logging explicitly, instead of relying on `Environment`

kim:
Deploy using debug logging and no layout + colour.

I left the defaults at colourful logging for convenience, at the expense of CLI options and logger defaults no differing.

### DIFF
--- a/src/Oscoin/Telemetry/Options.hs
+++ b/src/Oscoin/Telemetry/Options.hs
@@ -1,0 +1,63 @@
+module Oscoin.Telemetry.Options
+    ( Options(..)
+
+    , telemetryOptionsParser
+    , telemetryOptionsOpts
+    , renderTelemetryOptionsOpts
+    )
+where
+
+import           Oscoin.Prelude hiding (option)
+
+import qualified Oscoin.Telemetry.Logging as Log
+
+import qualified Formatting as F
+import           Options.Applicative
+import           System.Console.Option
+
+data Options = Options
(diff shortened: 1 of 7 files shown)


src/Oscoin/Telemetry/Options.hs
```diff
@@ -0,0 +1,63 @@
...
+           <> metavar (enumBounded (Proxy @Log.LayoutFormat))
+           <> value   Log.HumanReadable
+           <> showDefault
+            )
+
+    enumBounded
+        :: forall proxy a. (Show a, Enum a, Bounded a)
+        => proxy a
+        -> String
+    enumBounded _ = intercalate "|" $ map (show @a) [minBound..maxBound]
```
adinapoli-mndc: This looks like a generally-useful function, I would be in favour of having it at the top level 😉

kim: Sure... but ... where?

adinapoli-mndc: I do wonder if something _like Oscoin.Prelude_ (but perhaps not the prelude directly) might be a good candidate 🤔

kim: I don't think this is in any way a prelude-ish function, it applies specifically to `optparse-applicative` `metavar`s, where you'd want to display the available options from an enum. If you can think of more options-related things, that may justify some options utils fns module.

PR #635, opened 2019-06-03
Better suited genesis block constructors for tests

geigerzaehler:
We introduce `someSealedGenesisBlock` and `someSealedGenesisBlock'` in `Oscoin.Test.Crypto.Blockchain.Block.Generators` which allow us to simplify test code that relies on some fixed genesis block

As a result we can eliminate the block constructors `emptyGenesisBlock` and `emptyGenesisFromState` from `Oscoin.Crypto.Blockchain.Block`.

As part of the change we also merge `defaultBeneficiary` into the `Block.Generators` module and delete `Oscoin.Test.Crypto.Blockchain.Block.Helpers`

### DIFF
--- a/test/Oscoin/Test/Consensus/Nakamoto.hs
+++ b/test/Oscoin/Test/Consensus/Nakamoto.hs
@@ -19,12 +19,10 @@ import qualified Oscoin.Storage.Block.Pure as BlockStore.Pure
 import           Oscoin.Storage.HashStore
 import qualified Oscoin.Storage.Ledger as Ledger
 import           Oscoin.Storage.Receipt
-import           Oscoin.Time
 
 import           Oscoin.Test.Consensus.Network
 import           Oscoin.Test.Crypto
-import           Oscoin.Test.Crypto.Blockchain.Block.Helpers
-                 (defaultBeneficiary)
+import           Oscoin.Test.Crypto.Blockchain.Block.Generators
 import           Test.Oscoin.DummyLedger
 
 import           Codec.Serialise (Serialise)
@@ -88,7 +86,7 @@ instance (IsCrypto c) => TestableNode c PoW (NakamotoNode c) (NakamotoNodeState 
         let stateStore = mkStateHashStore nakStateStoreL
         let receiptStore = mkStateReceiptStore nakReceiptStoreL
         let ledger = Ledger.mkLedger (fst blockStore) stateStore dummyEvalBlock receiptStore
-        maybeBlock <- mineBlock ledger nakConsensus tn defaultBeneficiary
+        maybeBlock <- mineBlock ledger nakConsensus tn someBeneficiary
         -- NOTE (adn): We are bypassing the protocol at the moment, but we
         -- probably shouldn't.
         forM maybeBlock $ \blk -> do
@@ -125,7 +123,7 @@ initNakamoto nid = NakamotoNodeState
     , nakReceiptStore = Map.empty
     }
   where
-    genBlk   = sealBlock emptyPoW $ emptyGenesisFromState epoch defaultBeneficiary genesisState
+    genBlk = someGenesisBlock' emptyPoW $ Crypto.fromHashed $ Crypto.hash genesisState
     genesisState = mempty :: DummyState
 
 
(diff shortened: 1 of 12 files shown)


test/Oscoin/Test/Consensus/Nakamoto.hs
```diff
@@ -88,7 +86,7 @@ instance (IsCrypto c) => TestableNode c PoW (NakamotoNode c) (NakamotoNodeState
         let stateStore = mkStateHashStore nakStateStoreL
         let receiptStore = mkStateReceiptStore nakReceiptStoreL
         let ledger = Ledger.mkLedger (fst blockStore) stateStore dummyEvalBlock receiptStore
-        maybeBlock <- mineBlock ledger nakConsensus tn defaultBeneficiary
+        maybeBlock <- mineBlock ledger nakConsensus tn someBeneficiary
```
adinapoli-mndc: I find the notion of `some` a bit .. vague perhaps? What about `testBeneficiary` or `mockBeneficiary`? At least now the purpose and intention in clear. I suspect the same could apply for `someSealedGenesisBlock` & co.

test/Oscoin/Test/Crypto/Blockchain/Block/Generators.hs
```diff
@@ -61,6 +67,36 @@ genStandaloneBlock = do
                }
     pure $ mkBlock header blockDataBeneficiary blockDataTxs
 
+someBeneficiary :: Crypto.HasHashing c => Beneficiary c
+someBeneficiary = Crypto.zeroShortHash
```
adinapoli-mndc: Bikeshedding on my bikeshed .. here the purpose is clearly to distinguish between arbitrary beneficiaries vs a "stock", fixed one .. perhaps `fixedBeneficiary` or `stockBeneficiary` is better fitting?

test/Oscoin/Test/Crypto/Blockchain/Block/Generators.hs
```diff
@@ -61,6 +67,36 @@ genStandaloneBlock = do
...
+       , Serialise (Crypto.Hash c)
+       , Crypto.HasHashing c
+       )
+    => s -> Block c tx (Sealed c s)
+someSealedGenesisBlock seal = someSealedGenesisBlock' seal Crypto.zeroHash
+
+-- | Returns an empty genesis block with the given seal and state hash.
+-- The block is not intrinsically valid. It uses 'someBeneficiary'.
+someSealedGenesisBlock' :: (Serialise s, Serialise (Crypto.Hash c), Crypto.HasHashing c) => s -> Crypto.Hash c -> Block c tx (Sealed c s)
+someSealedGenesisBlock' seal stateHash = mkBlock header someBeneficiary []
```
adinapoli-mndc: For some reason, I find the `Sealed` in the name completely unnecessary, as by default each "valid" block must be sealed -- what about `stockGenesisBlock` & co ? That would go hand in hand with the `stockBeneficiary` idea.

test/Oscoin/Test/Crypto/Blockchain/Block/Helpers.hs
```diff
@@ -1,8 +0,0 @@
-module Oscoin.Test.Crypto.Blockchain.Block.Helpers where
-
-import           Oscoin.Crypto.Blockchain.Block (Beneficiary)
-import           Oscoin.Crypto.Hash (HasHashing, zeroShortHash)
-
--- | The default beneficiary.
-defaultBeneficiary :: HasHashing c => Beneficiary c
-defaultBeneficiary = zeroShortHash
```
adinapoli-mndc: I've met once a colleague which was strongly against helpers module as in his eyes they were some sort of anti pattern -- perhaps he was right, who knows 😉

test/Oscoin/Test/Crypto/Blockchain/Block/Generators.hs
geigerzaehler: How about `beneficiaryFixture`?

test/Oscoin/Test/Crypto/Blockchain/Block/Generators.hs
geigerzaehler: 👌 on removing “Sealed”.

test/Oscoin/Test/Crypto/Blockchain/Block/Generators.hs
adinapoli-mndc: I guess it boils down to personal preferences -- for some reason I prefer an `adjective + noun` schema here, but I won't antagonise if you pick `beneficiaryFixture` 😉

geigerzaehler:
@adinapoli-mndc: Could you have another look? Thank you

Issue #578, opened 2019-05-22
Account key management


geigerzaehler:
This issue describes the current requirements for account key management, provides a specification of the CLI and outlines the implementation plan.

This all is required to build the minimal devnet CLI (#548).

Requirements
------------

* As a user I want the CLI to use my locally stored key pair if necessary (e.g. to sign transactions).
* As a user I want to create an account (i.e. a keypair) from the CLI.
* As a user I want to display the account address for my locally stored key pair so that I can share it (and use it as the miner beneficiary).

Commandline Interface
---------------------

### `oscoin-cli account create`

Generates an Ed25519 keypair and stores it in `~/.config/oscoin/account.key` and `~/.config/oscoin/account.pub`. It notifies the user of the location of the files.

If one of those files already exists we refuse to overwrite them and notify the user.

### `oscoin-cli account show-address`

Shows the formatted address corresponding the stored keypair for the devnet network. The formatted address is the base32 encoding defined in `Oscoin.Crypto.Address`

### Reading keys

Commands that require the account keypair (like `transfer-money`) will try read the key pair from. If this fails they will show an error to the user.

Relation to node keys
---------------------

There already exists a `oscoin-cli keypair generate` command. This command generates an Ed25519 keypair and stores it in `~/.config/oscoin/id.{key,pub}`. These files are read by the oscoin node and the keypair is used by gossip. To avoid confusion we should rename these files to `~/.config/oscoin/node.{key,pub}`.

Tasks
-----

- [ ] Adjust `MonadKeyStore` so that we can control the name of the key files. Currently it is fixed to `id.{key,pub}`
- [ ] Implement `account create`
- [ ] Implement `account show-address`
- [ ] Use `node.{key,pub}` as the node key location

cloudhead:
It would also be useful to have `account show-id`, that prints the account id, for use in `--beneficiary`, now that it takes an `AccountId`.

PR #636, opened 2019-06-03
Reusable keypair file storage

geigerzaehler:
We extract the functionality to read and write key paris to disk from `Oscoin.CLI.KeyStore` into `Oscoin.Crypto.PubKey.FileStore`. This makes this functionality easier to use since we are providing just functions instead of an MTL-style type class.

This is a necessary step towards having separate key management for the node keys and the account keys (#578).

### DIFF
--- a/exe/node/Main.hs
+++ b/exe/node/Main.hs
@@ -3,7 +3,6 @@ module Main (main) where
 import           Oscoin.Prelude hiding (option)
 
 import qualified Oscoin.API.HTTP as HTTP
-import           Oscoin.CLI.KeyStore (KeyStoreError(..), readKeyPair)
 import           Oscoin.Configuration
 import qualified Oscoin.Consensus as Consensus
 import qualified Oscoin.Consensus.Config as Consensus
@@ -11,6 +10,7 @@ import qualified Oscoin.Consensus.Nakamoto as Nakamoto
 import           Oscoin.Crypto (Crypto)
 import           Oscoin.Crypto.Blockchain.Genesis (buildGenesisBlock)
 import qualified Oscoin.Crypto.PubKey as Crypto
+import qualified Oscoin.Crypto.PubKey.FileStore as Crypto
 import           Oscoin.Data.Tx
 import           Oscoin.Node (runNodeT, withNode)
 import qualified Oscoin.Node as Node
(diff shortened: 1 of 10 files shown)


exe/node/Main.hs
```diff
@@ -63,10 +64,9 @@ main = do
                  (progDesc "Oscoin Node")
 
     keys <-
-        flip runReaderT (Just $ keysDir optPaths) $
-            catches readKeyPair
+            catches (Crypto.readKeyPair $ keysDir optPaths </> "id")
                 [ Handler $ \case
-                    KeyNotFound{}
+                    Crypto.KeyNotFound{}
```
kim: At this point I'm wondering if we should just create a `Oscoin.Crypto` module which re-exports everything below 🤔

src/Oscoin/Crypto/PubKey/FileStore.hs
```diff
@@ -0,0 +1,88 @@
...
+-- | Read a keypair from disk. See 'writeKeyPair' about how the path
+-- parameter determines the key location. May raise 'ReadKeyError'.
+readKeyPair :: (MonadIO m) => FilePath -> m (Crypto.KeyPair Crypto)
+readKeyPair keysPathPrefix = liftIO $ do
+    let (pkPath, skPath) = keyPaths keysPathPrefix
+    pk <- readKeyFile deserialiseOrFail pkPath
+    sk <- readKeyFile Crypto.RealWorld.deserialisePrivateKey skPath
+    pure (pk ,sk)
+  where
+    readKeyFile customDeserialiseOrFail path = do
```
kim: Should this live at the top-level? Or what's the idea behind passing in the deserialiser?

geigerzaehler: The custom deserializer is passed out of necessetiy since we don’t have a `Serialise` instance for the private key. I don’t think it needs to be a top-level function. Ideally the signature of this would be
~~~haskell
readKeyFile :: (Serialise a) => FilePath -> IO a
~~~
but this fails because there is no `Serialise PrivateKey` instance. The `readKeyFile` hacks around this by accepting the serialization as an argument.

exe/node/Main.hs
geigerzaehler: I’d be all up for that.

PR #640, opened 2019-06-03
Implement 'evaluate' on OscoinTx

cloudhead:


src/Oscoin/Data/OscoinTx.hs
```diff
@@ -169,6 +170,64 @@ mkTxPayload ms pk = TxPayload
...
+    :: ( Serialise (AccountId c)
+       , Ord (AccountId c)
+       , Show (AccountId c)
+       , Show (Crypto.Hash c)
+       , Crypto.Hashable c (PublicKey c)
+       ) => Evaluator c (TxState c) (Tx c) TxOutput
+evaluate beneficiary txs (TxState ws) = do
+    case mintCoinbase beneficiary ws of
+        Left err  -> ([Left (toEvalErr err)], TxState ws)
+        Right ws' -> second TxState (applyTxs beneficiary txs ws')
```
cloudhead: Need to think more about what it means for `mintCoinbase` to fail.

src/Oscoin/Data/OscoinTx.hs
```diff
@@ -1,5 +1,6 @@
 {-# LANGUAGE DataKinds            #-}
 {-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE NumericUnderscores   #-}
```
geigerzaehler: ❤️ Maybe something that we can add to `package.yaml`.

src/Oscoin/Data/OscoinTx.hs
```diff
@@ -169,6 +170,64 @@ mkTxPayload ms pk = TxPayload
...
+    -> [Tx c]
+    -> WorldState c
+    -> ([Either EvalError TxOutput], WorldState c)
+applyTxs beneficiary transactions world =
+    apply transactions ([], world)
+  where
+    apply (Tx'{txPayload} : txs) (results, ws) =
+        case applyTxPayload txPayload beneficiary ws of
+            Left err         -> apply txs (Left (toEvalErr err)  : results, ws)
+            Right (ws', out) -> apply txs (Right out : results, ws')
```
geigerzaehler: This we will need to adjust after #641. It will be simpler.

PR #643, opened 2019-06-03
Various

kim:
See commit messages

### DIFF
--- a/src/Oscoin/Deployment/Internal/Systemd.hs
+++ b/src/Oscoin/Deployment/Internal/Systemd.hs
@@ -10,6 +10,8 @@ module Oscoin.Deployment.Internal.Systemd
     , ServiceSection
     , InstallSection
 
+    , RestartPolicy (..)
+
     , Service
     , renderService
     , dockerDaemonService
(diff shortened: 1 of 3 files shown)


src/Oscoin/Deployment/Internal/Systemd.hs
```diff
@@ -10,6 +10,8 @@ module Oscoin.Deployment.Internal.Systemd
     , ServiceSection
     , InstallSection
 
+    , RestartPolicy (..)
+
```
adinapoli-mndc: In case is useful, I did want to put this on your radar, in case you didn't know it already: http://hackage.haskell.org/package/systemd-2.0.0/docs/System-Systemd-Daemon.html

Used in the past, clunky but it gets the job done.

kim: Thanks, but that's a different thing, no? I'm only generating unit files, not talking to DBUS.

adinapoli-mndc: It is -- that's more of a braindump I could have attached to any deployment-related PR, but the word `watchdog` for some reason triggered some unconscious recollection 😁

PR #645, opened 2019-06-04
Correctly set color logging from cli options

geigerzaehler:


src/Oscoin/Telemetry/Options.hs
```diff
@@ -32,10 +32,10 @@ telemetryOptionsParser = Options
     <*> styleParser
   where
     styleParser = Log.StyleFormatter
-        <$> switch
+        <$> (not <$> switch
```
kim: `flag True False` might be clearer

PR #646, opened 2019-06-04
GCE: fix log source

kim:
Apparently, the fluentd version which comes with COS doesn't understand `match` (yet), and just silently ignores the source.

Note that there is at least one builtin rule which may cause log records to be shipped multiple times: any journal message with severity >= warning will end up `logName=projects/[PROJECT]/logs/cos_journal_warning`.


kim:
Here's the magic incantation of `gcloud` to view the 10 most recent log records for `oscoin`:

```shell
gcloud logging read "logName=projects/opensourcecoin/logs/containers AND jsonPayload.CONTAINER_NAME=oscoin" --limit=10
```

I am rather disappointed that there is no streaming :(

PR #647, opened 2019-06-04
Start the miner only after first sync finishes

### LINKED ISSUE #628: Disable miner when syncing the first time
While testing node syncing and consensus, I did wonder the following: should the miner produce blocks even though syncing is in progress? It seems to me that this shouldn't be the case: while technically eventually nodes should reach consensus, it simply looks wasteful to me: we are mining blocks on a potentially very stale tip, only to throw them away (forever) once we start chasing the correct chain.

In order start the miner _a posteriori_, we could potentially extend the sync interface with an `onReady` callback that could be called once we finished syncing the first time. This would require a small adjustment in the syncing interface, but it would also avoid keep around mutable state or worse do something with MVars (like using one to block the miner up until we finished syncing).

If we implement the `onReady` idea, one would wire everything inside `node/Main.hs` and only when we are ready, we would spawn the miner (which runs in an async thread anyway).

Incidentally, it occurred to me we could use `onReady` also to fix #588: In this case `onReady` would also (likely) update a `Tvar` which the HTTP API can then later read to answer the question: is the node ready or not? It would also mitigate all @kim's concerns about rollabacks and the remote tip being a moving target.

@cloudhead @kim @geigerzaehler thoughts?

adinapoli-mndc:
Fix #628. 

This PR extends the `sync` interface for the `RealWorld` implementation with an `onReady` callback of type `Either SyncError (LocalTip c tx s, RemoteTip c tx s) -> m ()` which will be called once the node completed the first (mandatory) sync round.

The use of `Either` in the callback ensure that in case of errors (for example if the node is isolated) the miner will be started anyway, and the error propagated to the callback which then can deal with it (or ignore it).

At the moment in `exe/Main.hs` we are not doing anything with the result of the sync process, but future PR might capitalise on this by using the local & remote tips in interesting ways.

### DIFF
--- a/exe/node/Main.hs
+++ b/exe/node/Main.hs
@@ -193,14 +193,19 @@ main = do
                                     ]
                                     (Telemetry.telemetryProbe telemetry)
 
+            -- When the node finished syncing for the first time, start the
+            -- miner and link the async to the main thread.
+            let onNodeReady _result = liftIO $
+                    link =<< async (miner node gossip)
+
             liftIO . runConcurrently $
                    Concurrently (HTTP.run (fromIntegral optApiPort) node)
-                <> Concurrently (miner node gossip)
                 <> (Concurrently $
                         maybeRunMetrics telemetry optMetricsHost optMetricsPort)
                 <> (Concurrently $
                         maybeRunEkg metricsStore optEkgHost optEkgPort)
-                <> (Concurrently $ Sync.runSync syncContext Sync.syncNode)
+                <> (Concurrently $
+                        Sync.runSync syncContext (Sync.syncNode onNodeReady))
 
     either (const exitFailure) (const exitSuccess) res
   where
--- a/src/Oscoin/Protocol/Sync/RealWorld.hs
+++ b/src/Oscoin/Protocol/Sync/RealWorld.hs
@@ -158,7 +158,7 @@ catchNetworkError protoReq addr action = do
 newEventTracer :: Probe IO -> Tracer IO
 newEventTracer = probed
 
--- ^ Given a monadic IO operation and a maximum Duration for this action to
+-- | Given a monadic IO operation and a maximum Duration for this action to
 -- take, it either returns the value @a@ or a 'Timeout'.
 timed :: MonadUnliftIO m => Duration -> m a -> m (Either Timeout a)
 timed maxTimeout action = do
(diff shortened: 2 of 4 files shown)


exe/node/Main.hs
```diff
@@ -193,14 +193,19 @@ main = do
                                     ]
                                     (Telemetry.telemetryProbe telemetry)
 
+            -- When the node finished syncing for the first time, start the
+            -- miner and link the async to the main thread.
+            let onNodeReady _result = liftIO $
+                    link =<< async (miner node gossip)
```
geigerzaehler: This will crash the syncing if the miner crashes. Are we ok with that? We also don’t get a handle on the miner thread. This might be an issue if we want to decouple it more.

src/Oscoin/Protocol/Sync/RealWorld.hs
```diff
@@ -239,9 +242,19 @@ syncNode
...
-    sync `catchError` \(e :: SyncError) ->
-       recordEvent $ Telemetry.Events.NodeSyncError (toException e)
-    -- Throttle each iteration of the algorithm by 30 seconds.
-    liftIO $ threadDelay 30000000
+    => (Either SyncError (LocalTip c tx s, RemoteTip c tx s) -> m ())
+    -- ^ A callback which will fire once the sync has been performed the
+    -- first time.
+    -> SyncT c tx s m ()
+syncNode onReady = do
+    lift . onReady =<< catchingSyncErrors (syncUntil (\_ _ _ -> False))
```
geigerzaehler: Why don’t we use `isDone` here?

adinapoli-mndc: The intention was to make sure `syncUntil` was actually fetching some blocks _no matter what_, but I think you are right, I think calling `isDone` should be perfectly fine here.

exe/node/Main.hs
adinapoli-mndc: > This will crash the syncing if the miner crashes. Are we ok with that?

I suppose we don't have too much of an alternative, due to the way our spaghetti `Concurrently` is orchestrated at the moment. I think (but I might be wrong) but if any of the `Concurrently` dies, the other ones will follow, so this won't change (fundamentally) the disaster recovery strategy of our node.

> We also don’t get a handle on the miner thread. This might be an issue if we want to decouple it more.

Agreed, but I wouldn't try to get this perfect the first time, we can easily extend this further as we see the need for.

src/Oscoin/Protocol/Sync/RealWorld.hs
adinapoli-mndc: 8c9272c5

adinapoli-mndc:
@kim fancy taking a look and tell me what do you think?

PR #652, opened 2019-06-05
Add CLI option to disable the miner

adinapoli-mndc:
Requires #647.

This PR adds a `--no-mining` switch to our parser and selectively starts the miner in the `onReady` callback in `exe/Main.hs`.


kim:
The iffy thing about this is that it is not a valid configuration unless "lenient consensus" is used. I would prefer an options parse tree which doesn't allow such things.

This would mean a bit of reorganisation, though: I think we'd get rid of the `Environment` paradigm entirely, and perhaps introduce something like

```haskell
data ConsensusOptions
    = NakamotoStrict NakamotoStrictOptions
    | NakamotoLenient NakamotoLenientOptions
```

adinapoli-mndc:
> The iffy thing about this is that it is not a valid configuration unless "lenient consensus" is used. I would prefer an options parse tree which doesn't allow such things.

Yes, that's a valid concern 😞 

> This would mean a bit of reorganisation, though: I think we'd get rid of the Environment paradigm entirely, and perhaps introduce something like

I think that, at this stage, we are not harnessing the use of the `Environment` anywhere in the codebase (I think). The last piece standing was logging, which you recently changed, so perhaps indeed we should just get rid of the `Environment` and go with your proposal.

adinapoli-mndc:
Thinking out loud here: are we sure this hasn't to do more with the particular network we are running (i.e. devnet vs mainnet) and less with the particular consensus algorithm we are using?

kim:
> less with the particular consensus algorithm

It's about the validation rules: if a non-mining node doesn't receive blocks for `N` amount of time, it will reject any subsequent rules. Unless running with lenient validation rules.

adinapoli-mndc:
@kim PTAL. I am not sure if there is a better way to do this more elegantly with `optparse-applicative`, but here I have used `flags'` to selectively switch from one on the other subparser. It's a bit gross as the "consensus-specific" options leaks into the general help, but it gets the job done and it exhibits the right behaviour. I can use the lenient consensus just fine:

```
monadic/oscoin (adinapoli/issue-605-part-1) > stack exec oscoin -- --beneficiary 0x0000000000000000000000000000000000000000 --blockstore blockstore.db --nakamoto-consensus-lenient --block-time-lower 1 --no-mining                         
info  | node starting logical_network=devnet physical_network=z11r6q981qmxgme4qf6o7c6wi2em0bz41y7lnhr7rn6g8btwn4r605k-n4q1 genesis=2Drjgb4dT8cu6efEWUK9HZqjzNiwrKXpCEmJBZDndMThk9Gwq5 loc="Oscoin.Node:82:13"
Spock is running on port 8477
debug | performing discovery...  ns="disco" loc="Oscoin.P2P.Disco:100:14"
```

But if I try to use those options for the `--strict`:

```

monadic/oscoin (adinapoli/issue-605-part-1) > stack exec oscoin -- --beneficiary 0x0000000000000000000000000000000000000000 --blockstore blockstore.db --nakamoto-consensus-strict --block-time-lower 1 --no-mining                          
Invalid option `--block-time-lower'

Usage: oscoin [-h|--host ARG] [--gossip-port ARG] [--api-port ARG]
              [--network STRING|mainnet|testnet|devnet] [--seed HOST:PORT]
              [--sd-domain DOMAIN NAME] [--enable-mdns] [--keys FILEPATH]
              [--blockstore FILEPATH] [--genesis-parameters FILEPATH]
              ([--nakamoto-consensus-strict] | [--nakamoto-consensus-lenient]
              [--block-time-lower SECONDS] [--no-mining]) [--metrics-host ARG]
              [--metrics-port ARG] [--ekg-host ARG] [--ekg-port ARG]
              [--allow-ephemeral-keys] --beneficiary ARG
              [--log-level Debug|Info|Err] [--no-log-colours]
              [--log-layout NoLayout|HumanReadable]
  Oscoin Node
```

adinapoli-mndc:
Ooops, gotta update the deployment scripts methinks.

PR #654, opened 2019-06-06
Decouple legacy tx output from oscoin tx output

geigerzaehler:
To make it easier to change `TxOutput` from `OscoinTx` we decouple it from `TxOutput` from `Tx` (the legacy tx) by changing the `TxOutput` type family instance. This change has some repercussions and requires us adapt `withNode` from `Oscoin.Test.HTTP.Helpers`.

This in turn would require us to run the sync tests with the OscoinTx evaluator. Since this evaluator does not yet exist we opt for running the sync tests with the legacy tx. This requires a switch to the memory block store.

### DIFF
--- a/src/Oscoin/Data/Tx.hs
+++ b/src/Oscoin/Data/Tx.hs
@@ -4,6 +4,7 @@ module Oscoin.Data.Tx where
 import           Oscoin.Prelude
 
 import           Oscoin.Crypto.Blockchain.Block (BlockHash)
+import           Oscoin.Crypto.Blockchain.Eval (Evaluator)
 import           Oscoin.Crypto.Hash
 import qualified Oscoin.Crypto.Hash as Crypto
 import           Oscoin.Crypto.PubKey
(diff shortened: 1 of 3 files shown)


adinapoli-mndc:
@geigerzaehler I promise to get back to you and review this one as soon as I am out of the woods with my current tasks -- in theory this shouldn't create too many conflicts in my strand of work (but will create _some_ as I am touching the `Oscoin.Protocol.Sync.Test` module), so it might be easier to address these on your side (or perhaps not, we will see 😉 )

geigerzaehler:
Could you have another look at this @adinapoli-mndc? It is blocking my work on #640

src/Oscoin/Data/Tx.hs
```diff
@@ -120,3 +123,9 @@ validateTx Tx{..} =
     if verify txPubKey txMessage
     then Right ()
     else Left TxInvalidSignature
+
+
+evaluateBlock :: Evaluator c LegacyTxState (Tx c) DummyPayload
+evaluateBlock _beneficiary txs st = foldl' go ([], st) txs
+  where
+    go (output, st') tx = (output <> [txMessageContent tx], st')
```
adinapoli-mndc: Dumb question -- any reason why this cannot be prepended rather than appended? I assume because that's linear ordering, i.e. the tx being evaluated depends from the context of what we have evaluated so far, and thus it needs to go at the end.

I do wonder if it would be more efficient to append at front and do a final `reverse` at the end -- that's usually the case for these kind of folds.

geigerzaehler: The reason for this is indeed that we want the order of the outputs to match the order of the transactions. But you’re right that it would make sense to prepend and reverse.

PR #655, opened 2019-06-07
Improved syncing algorithm to deal with dishonest behaviour and forks

### LINKED ISSUE #651: Improve node sync for non-optimal consensus and potential adversarial behaviour
## Summary

_Our current node sync works well only if the majority (if not all) the active peers are in agreement, which might not be the case in production. On top of that, it doesn't account for adversarial behaviour when splitting work between peers in `syncBlock`. I propose we modify the syncing algorithm (and the spec) as explained below to account for that._

## Problem

While conducting some local experiments, I did unravel a flaw of the current implementation of `syncBlocks`, for which I am going to post only a stripped down version:

```haskell
syncBlocks rng = do
    ctx <- ask
    let fetcher = scDataFetcher ctx
    withActivePeers $ \ activePeers ->

        lift $ runConduit $
               C.yieldMany (workDistribution activePeers)
            .| C.mapM (fetchBlocks fetcher >=> notifyUpstreamConsumers ctx)
            .| C.concat
            .| C.iterM (fetchMissing fetcher activePeers ctx)
            .| C.sinkNull
```

My laudable goal here was to split work across peers, so that we wouldn't overload a particular peer with tons of requests and that could in future be made parallel with relatively little effort. However, doing this height-based work-split is quite dangerous: in case _not all the peers reached consensus_, we could get back from a peer a particular fragment of a chain which is either stale (i.e. a rollback occurred), totally bogus (dishonest player) or simply a divergent chain (for any reason whatsoever). In such scenario, we are not doing the consensus layer any favour; we are sending it a bunch of uncorrelated chain fragments, with the likely outcome that the node being synced won't reach consensus (ever).

To illustrate this better, let's consider the following scenario (taken from a real experiment): we have 3 nodes: `node1` has the longest chain of 100 blocks, `node2` is on a divergent fork and `node3` is the node about to start syncing. Below let's find the `local_height` for all the three:

| Node name | Local Height |
|-----------|--------------|
| Node 1    | 100          |
| Node 2    | 80           |
| Node 3    | 60           |

When `node3` starts syncing, it requests the best remote tip to `node1` and `node2`. If you recall the algorithm, we pick the block with most frequency or, in case of draws, the highest one. Here `node1` and `node2` will return different blocks, but `node1` has the highest, so we start the sync by calling `syncBlocks` on the range `(61,100)` inclusive.

**Now, here is the problem**: we divide the work between `node1` and `node2` equally, but some (or all) the slices associated to `node2` will be relevant to a completely different chain respect to `node1`, which means that, as a result, `node3` won't receive a meaningful, consistent chain, and might be able to apply some blocks here and there. The rest, will be put in the orphanage.

What is worst is that perhaps `node3` 's local height might be able to advance of a bunch of blocks (either due to `node3` 's miner or by serendipitous block application) but the process will repeat similarly during the next "tick" of the sync algorithm.

While in principle the node should be able to _eventually_ recover from this by piggybacking on gossip's block retransmission, it looks to me a lame way of syncing blocks.

First of all, let's get the elephant out of the room: despite my original intention was to maximise download speed and parallelism, I have now realised this is a fool's errand because (as "the internet" warned us already) even if we maximise the network throughput the syncing bottleneck is always going to be I/O bound, because we will need to apply and validate each block, and each call to `dispatchBlockSync` acquires a mutex on the orphanage, so everything happens sequentially. While we could in the future mitigate this with checkpoints (ala Parity) for now "correct" triumphs "fast".

With this in mind, here is my proposal, which is articulated in two steps. It doesn't require a complete overhaul of the syncing (for which the API won't change at all) but this would be merely internal.

During the first step, the node will compute what we can call the `commonChainHeight`: in order to actually fetch the block it needs, the nodes has to `A)` understand it's on a fork and `B)` how many blocks it has to fetch. If you recall the example I have posted above, let's assume that `node3` was also on a fork (to spice things up). When `node3` starts, we would run this `commonChainHeight` function which would eventually return the _height_ *all* the peers share in common, i.e. the height which correspond to a block which hash matches for *all* peers. Once this height has been acquired, the node can proceed to actually request the blocks in the range `(commonHeight + 1, bestRemoteTip)`, essentially as before.

Step 2 would involve changing the implementation of `syncBlocks`. Rather than trying to fetch the blocks in chunks based on a height range, we would focus on a classic iterative algorithm where we would start from `bestRemoteTip` and, for each block, request to _all peers, in a round robin fashion_, the block at hash `blockPrevHash`. We can obviously stop as soon as the block is found, skipping unnecessary HTTP calls. This is easy to do as our HTTP API already exposes such endpoint. If the block cannot be retrieved immediately, we have no choice but to wait and retry (either a fixed number of times or with a timeout). If we cannot fetch a block, we simply stop and restart the sync (i.e wait for the next "tick" of the endless loop) to give time to either the peer topology to adjust or for the main chain to reconcile.

## commonChainHeight implementation details

Implementing this function piggybacks on ideas from other projects. Essentially `node3` would ask all its peers for the `blockHash` corresponding to the block at height `60` (remember the table, it's the height we start after a reboot). If we get perfect agreement (i.e. all the hash matches), we are done. If not, we backtrack the height using a fibonacci sequence, so we would request heights 59,59,58,57,55,52,... up until a fixed limit (it could be 2016 or simply configurable on the `SyncContext`). If we cannot ever reach agreement.. see the section below.

## commonChainHeight and adversarial behaviours
 
There is a complication when implementing `commonChainHeight`: if one of our peers is dishonest (or simply not in agreement) we might end up backtracking past the limit without finding a suitable common height. In this scenario, I propose to do what (for example) `grin` does: we *start back from genesis*. The rationale would be that if we didn't reach an agreement it means something is off, either with the consensus or with the network in general (perhaps we are under attack), so the most conservative choice is to resume syncing from scratch.

## commonChainHeight and its variations

One might ask: why not implement `commonChainHeight` such that we don't require perfect agreement but we take the height of the block which root hash matches for _most peers_? We could do that, however it doesn't work that well for small clusters; if we have a group of 3 nodes like the one I have just presented, now we might have that (for example) `node1` and `node2` are in disagreement (i.e. `node2` is dishonest or on a very stale fork). In this situation, we don't have enough active peers to be able to resolve the ambiguity, and we cannot meaningfully pick `node1` or `node2` without having an oracle that knows `node1` is the "good" node (alas we don't have such oracle).

_Having said that_, the probability of *not* finding a `commonChainHeight` after the full backtrack window is very rare and it's likely to happen only in case of dishonest peers (which are the exceptions, not the rule). Even a node on a fork lagging severely behind should conceivably have some stale-but-far-away-from-genesis height, so agreement should be reached most of the times. If it doesn't it means our consensus is buggy, but that's a separate issue 😁 

cc @cloudhead @kim @geigerzaehler

adinapoli-mndc:
Fixes #651.

This commit overhauls the internals of `Protocol.Sync` to provide a syncing algorithm which works in the presence of forks and adversarial behaviour.

I have tested this briefly locally and it works **really, really well**: Now a cluster of 3 nodes achieve perfect consensus smoothly and nodes are capable of backtracking in case they pick the wrong fork:

Screenshot 2019-06-07 at 12 35 58

A bunch of things to notice:

* We **do need** a new HTTP endpoint to fetch the block hashes cheaply. For now I have implemented `GetBlockHashes` by re-using `/blocks/by-height`, but that's obviously very expensive. There is no need to step this PR from being merged though, it will take a while before this becomes a performance problem and surely that's not too important for devnet. Sure enough I will open a PR to provide this endpoint today or next week;

* The `isDone` predicate could be improved, in particular by computing `commonChainHeight` rather than relying on `localTip` which might be on a fork and thus on the wrong height. In my tests though this doesn't seem to make too much of a difference, so that's an improvement that can go in later (not to mention `commonChainHeight` is more expensive than fetching the local tip, so we might simply don't bother).

* The spec will need to be updated to reflect the new behaviour.


src/Oscoin/Protocol/Sync.hs
```diff
@@ -354,109 +386,317 @@ getRemoteHeight
...
+    :: forall c tx s m.
+       ( Monad m
+       , Ord (Hash c)
+       )
+    => BacktrackPolicy
+    -> Block c tx (Sealed c s)
+    -> SyncT c tx s m Height
+commonChainHeightWith policy localTip = do
+    ctx <- ask
+    go ctx (height localTip) 0 5
```
kim: IIUC, you're not using the `retry` package here due to `threadDelay` being mockable. I still wonder if this could be made a bit clearer - and allow for instrumentation - if the recursion could be factored out of the actual work. As in `go :: SyncContext c tx s m -> Height -> SyncT c tx s m (Either RetryCondition Height)` + a `loop` function which scrutinises `RetryCondition` and either emits some telemetry and retries, or returns `Height`.

adinapoli-mndc: > IIUC, you're not using the retry package here due to threadDelay being mockable.

Yes, that's exactly it. Unfortunately all the combinators from `retry` lives in `MonadIO`, which kills any hope to use the package in a pure setting. I did also try briefly to use _only_ the `RetryStatus` and `RetryPolicy` parts (which are pure) and integrate them into our code, but the result was horrible, as those have a very specific use case, tailored to that library.

> I still wonder if this could be made a bit clearer - and allow for instrumentation - if the recursion could be factored out of the actual work. As in go :: SyncContext c tx s m -> Height -> SyncT c tx s m (Either RetryCondition Height) + a loop function which scrutinises RetryCondition and either emits some telemetry and retries, or returns Height.

Maybe -- I will try rehashing things and see how things turns out.

PR #659, opened 2019-06-10
Fix bug in bulkAdjust

### LINKED ISSUE #656: Investigate "!! index too large" error
While running a local experiment, I did stumble across the following error:

```
error | Exception was thrown exception="ExceptionInLinkedThread ThreadId 6 Prelude.!!: index too large" callstack="CallStack (from HasCallStack):\n  logException, called at src/Oscoin/Telemetry/Logging.hs:288:63 in oscoin-0.1.0.0-9HJ5UpoPyUD4ZtWgF5IPX7:Oscoin.Telemetry.Logging\n  withExceptionLogged, called at exe/node/Main.hs:102:11 in main:Main" loc="Main:102:11"
```

I am fairly sure this comes from the orphanage, more specifically either from [here](https://github.com/oscoin/oscoin/blob/master/src/Oscoin/Storage/Block/Orphanage.hs#L501) or [here](https://github.com/oscoin/oscoin/blob/master/src/Oscoin/Storage/Block/Orphanage.hs#L547).

The use of `!!` is probably a no-no anyway, so I could seize the chance to rethink a little bit the internals.

adinapoli-mndc:
Fixes #656 .

This commit fixes a bug in 'bulkAdjust' which was causing
the orphanage to have less tips of the correct number, likely
causing an invalid lookup with List.!! . On top of this, I have also switched from using a list for the candidates to a vector, so that indexing works in `O(1)`.

While recreating the offending error is very tricky (and our tests passes with more than 30k iterations, but nevertheless we might not be exercising the orphanage in a way to trigger the problem) I have also added a variation of `Vector.!` that fails with a precise stacktrace, so that we can start isolating the crash to a specific function, when and if it will happen again. Example of `unsafeIndex` in action:

Screenshot 2019-06-10 at 17 16 24

As you can see, we now get the offending line in the Orphanage code.


src/Oscoin/Storage/Block/Orphanage.hs
```diff
@@ -412,10 +409,10 @@ pruneOrphanageShallow inputCandidate@ChainCandidate{candidateChain, candidateRoo
       pruneCandidate old =
           case map removeCandidate (Map.lookup candidateRootHash old) of
             Nothing -> (old, False)
-            Just [] -> (Map.delete candidateRootHash old, True)
+            Just xs | xs == Vector.empty -> (Map.delete candidateRootHash old, True)
```
kim: nit: `Vector.null xs`?

src/Oscoin/Storage/Block/Orphanage.hs
```diff
@@ -515,11 +512,22 @@ insertOrphan b (storeOrphan b -> o) =
                    $ tips orph
             }
 
+-- | A variation over 'Vector.!' that fails with a stacktrace.
+unsafeIndex :: HasCallStack => Vector a -> Int -> a
+v `unsafeIndex` ix = withFrozenCallStack $
+    let msg =  "Orphanage.unsafeIndex: index out of range ("
```
kim: nit: `unwords`?

src/Oscoin/Storage/Block/Orphanage.hs
```diff
@@ -515,11 +512,22 @@ insertOrphan b (storeOrphan b -> o) =
...
+            ++ ")"
+    in fromMaybe (Prelude.error msg) (v Vector.!? ix)
+
 findCandidate :: (ChainCandidate c s -> Bool)
               -> Candidates c s
               -> Maybe (ChainCandidate c s, Index)
 findCandidate predFn xs =
-    case List.filter (\(c,_) -> predFn c) (zip xs [0..]) of
+    let zipped = Vector.zip xs (Vector.fromList [0..])
+    in case Vector.toList $ Vector.filter (\(c,_) -> predFn c) zipped of
```
kim: nit: the `protolude` `head` should work here without converting to a list

src/Oscoin/Storage/Block/Orphanage.hs
adinapoli-mndc: Jawohl!

src/Oscoin/Storage/Block/Orphanage.hs
adinapoli-mndc: I do wonder why in the previous incarnation of this function I did request (at the pattern matching level) that a singleton list has to be returned. I don't think that's necessary, so indeed I think we can use `Protolude.head` here.

adinapoli-mndc: be0b7391

src/Oscoin/Storage/Block/Orphanage.hs
adinapoli-mndc: be0b7391

src/Oscoin/Storage/Block/Orphanage.hs
adinapoli-mndc: be0b7391

PR #665, opened 2019-06-13
Short hash parsing and serialising

geigerzaehler:
The account CLI needs to display and parse human readable short hashes (account IDs). To make this easy I propose a couple of changes.

Currently, the `HasHashing` class provides a method to parse a short hash from `Text`. A partial inverse of this function only exists as `formatShortHash` which returns a formatter and `show`. It is not clear that these functions are inverse to each other and this is nowhere enforced.

I propose to decouple the byte representation of a short hash from the human readable representation (encoding) of those bytes. In particular `parseShortHash` is removed from the `HasHashing` type class and replaced with a `Binary (ShortHash c)` constraint.

This has the additional advantage that the encoding is independent of the type class. We also make it easier to implement different polymorphic encoding/decoding functions for short hashes in the future

The same approach could be adapted for `Hash`, too.

Todo

- [ ] Test tripping
- [ ] Implement `Binary` instances

### DIFF
--- a/src/Oscoin/Crypto/Hash.hs
+++ b/src/Oscoin/Crypto/Hash.hs
@@ -13,6 +13,7 @@ module Oscoin.Crypto.Hash
     , formatHash
     , formatHashed
     , formatShortHash
+    , parseShortHash
 
     , hashBinary
     , hashSerial
@@ -39,6 +40,7 @@ import qualified Data.ByteString.BaseN as BaseN
 import qualified Data.ByteString.Lazy as LBS
 import           Data.Multihash (Multihash, Multihashable)
 import qualified Data.Multihash as Multihash
+import qualified Data.Text as T
 import qualified Database.SQLite.Simple.ToField as Sql
 import           Formatting (Format)
 import qualified Formatting as Fmt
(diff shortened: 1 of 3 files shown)


kim:
Why `Binary` vs. `ByteArrayAccess`? The former implies a certain encoding (e.g. length prefix), while the latter is just "raw bytes".

geigerzaehler:
> Why `Binary` vs. `ByteArrayAccess`? The former implies a certain encoding (e.g. length prefix), while the latter is just "raw bytes".

I was not aware that `Binary` requires a certain encoding. I assumed that it just gives you raw a raw binary encoding. I chose it because it provides methods for both encoding and decoding and is not as low-level.

kim:
> Binary requires a certain encoding

Look e.g. at the [instance for Integer](https://hackage.haskell.org/package/binary-0.8.7.0/docs/src/Data.Binary.Class.html#line-316) or for [Either a b](https://hackage.haskell.org/package/binary-0.8.7.0/docs/src/Data.Binary.Class.html#line-605). This is essentially a tagged binary representation. By using `ByteArrayAccess`, there's no doubt about what's going on, and you can't (easily) compose those into an accidentially-tagged larger structure.

